PR #69, opened 2025-06-05
fix/createstestrundrawer

nicolai-rhesis:
- **fixed the testset selection field name as key instead of id as key issue**
- **fixed application and endpoint selection not showing values**


nicolai-rhesis:
fixes #53

PR #155, opened 2025-08-04
Add Document Support Generate Tests Endpoint 151

EmanueleDeRossi1:
This PR introduces changes from the `feature/add-document-support-generate-tests-endpoint-151` branch.

## 🔄 Changes

- refactor: simplify document handler, remove mime type detection (9b291d6)
- refactor: restructure test generation endpoints (472abee)
- feat: implement modular document processing for test generation endpoints (16c0498)

## 📁 Files Changed (       5 files)

```
apps/backend/src/rhesis/backend/app/dependencies.py
apps/backend/src/rhesis/backend/app/routers/services.py
apps/backend/src/rhesis/backend/app/schemas/services.py
apps/backend/src/rhesis/backend/app/services/document_handler.py
apps/backend/src/rhesis/backend/app/services/generation.py
```

## 📋 Commit Details

```
9b291d6 - refactor: simplify document handler, remove mime type detection (Emanuele De Rossi, 36 minutes ago)
472abee - refactor: restructure test generation endpoints (Emanuele De Rossi, 70 minutes ago)
16c0498 - feat: implement modular document processing for test generation endpoints (Emanuele De Rossi, 3 days ago)
```

## ✅ Checklist

- [ ] All tests pass locally

### DIFF
--- a/apps/backend/src/rhesis/backend/app/dependencies.py
+++ b/apps/backend/src/rhesis/backend/app/dependencies.py
@@ -25,46 +25,26 @@ def get_endpoint_service() -> EndpointService:
     return EndpointService() 
 
 async def process_form_documents(
-    documents: Optional[str] = Form(None),
     files: Optional[List[UploadFile]] = File(None),
 ) -> Tuple[List[Dict], DocumentHandler]:
     """
-    FastAPI dependency that processes documents from form data.
+    FastAPI dependency that processes uploaded files.
     
     Args:
-        documents: JSON string containing document specifications
         files: List of uploaded files
         
     Returns:
         Tuple of (processed_documents, document_handler)
         
     Raises:
-        HTTPException: If JSON parsing fails or document processing fails
+        HTTPException: If file processing fails
     """
     handler = DocumentHandler()
     
     try:
-        # Parse documents JSON string if provided
-        parsed_documents = None
-        if documents:
-            try:
-                doc_data = json.loads(documents)
-                parsed_documents = [DocumentSpecification(**doc) for doc in doc_data]
-            except (json.JSONDecodeError, ValueError) as e:
-                handler.cleanup()
-                raise HTTPException(
-                    status_code=400,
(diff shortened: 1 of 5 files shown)


harry-rhesis:
## Review Feedback

Kudos: Very nice PR introducing an important feature for the product! The document support for test generation is a valuable addition.

### General Feedback

**Commit Size**: The commits should be smaller - rather than having one commit with a lot of code at once, try to get smaller commits. This makes reviews easier and helps with debugging when issues arise.

### Specific Technical Concerns

#### 1. DocumentHandler Architecture
The  should be conceived as a more general entity, something like , because there will be multiple instances in the platform where file upload will be needed. This is not only relevant for PromptSynthesizer but for any entity of the codebase that requires file handling.

#### 2. Single Responsibility Principle
The  method takes both  and  parameters. It's not very clear what the difference between them is, and the method seems to be doing two things at the same time, which violates the principle of single responsibility.

#### 3. Breaking Changes ⚠️
Be super careful with breaking changes. Whenever possible, the code should be made backwards compatible, unless there is a very strong incentive to go otherwise. Kudos though for pointing out the potential for breaking changes.

For example, changing this important route would break the UI and SDK at once:
-  schema fields changed from // to Please ask your administrator.//
-  now accepts form data instead of JSON (for file upload support)
- Added  for pure JSON requests

#### 4. Misleading Method Naming
The naming of the method  is somewhat misleading. It reads as if we are handling JSON documents ( files), but rather this is processing documents which are specified in a JSON list format.

### Recommendation

Create a completely separate file upload endpoint that can be used across different use cases. Any clients that want to have files will need to invoke this endpoint first. Generate tests is only one of those clients. This would:

1. Make the API more modular and reusable
2. Avoid breaking changes to existing endpoints
3. Provide a cleaner separation of concerns
4. Allow other features to leverage file uploads without duplicating logic

This approach would maintain backward compatibility while adding the new functionality in a more extensible way.

Overall, great work on this feature - just some architectural considerations to make it more maintainable and future-proof! 🚀

Issue #156, opened 2025-08-04
rhesis.app.models.guid.GUID(), AttributeError:module 'rhesis' has no attribute 'app'


asadaaron:
- when running alembic upgrade head, it throws rhesis.app.models.guid.GUID(), AttributeError:module 'rhesis' has no attribute 'app'

PR #158, opened 2025-08-05
fix: correct GUID import path in alembic migrations

asadaaron:
- Changed rhesis.app.models.guid.GUID() to rhesis.backend.app.models.guid.GUID()
- Fixed import path in all alembic migration files
- test the migration: create new database in gap and run the `alembic upgrade head` . It successfully  created all the schema in the new database.


harry-rhesis:
@asadaaron, thanks for this PR! It fixes an important bug, and lays the foundation for our self-hosted Rhesis 🎉 In the future, consider using our `.github/pr` tool.

asadaaron:
Fixes: https://github.com/rhesis-ai/rhesis/issues/156

PR #161, opened 2025-08-05
Fix: Prevent Env Loading In Vscode

### LINKED ISSUE #107: Make environment variable parsing tolerant of quotes and whitespace
### Description
When setting environment variables like `RHESIS_BASE_URL`, users sometimes include extra quotes, spaces, or inline comments.

Example:
```
RHESIS_BASE_URL="https://api.rhesis.ai/"
RHESIS_BASE_URL=https://api.rhesis.ai/   
RHESIS_BASE_URL=https://api.rhesis.ai/  # optional
```

Right now, our code does not handle this consistently.
It can lead to URLs like `https://api.rhesis.ai/%20%20` being used if there are trailing spaces or comments.

### Proposed solution 1: Robust parsing in code
- Strip optional leading/trailing quotes and whitespace when reading environment variables.
- Do _not_ support inline comments in .env values (this would add unnecessary complexity).

#### Suggested helper function:
```
def clean_env_var(var, default=None):
    value = os.getenv(var, default)
    if value:
        value = value.strip().strip('"').strip("'")
    return value
```

#### Implementation options: 

1. Use clean_env_var() wherever environment variables are read (e.g., in each module that loads RHESIS_BASE_URL).
2. (recommended/cleaner): create a `config.py` module that loads all environment variables once using `clean_env_var()`, and then import config variables from there throughout the project. Example:

```
import os

def clean_env_var(var, default=None):
    value = os.getenv(var, default)
    if value:
        value = value.strip().strip('"').strip("'")
    return value

RHESIS_API_KEY = clean_env_var("RHESIS_API_KEY")
RHESIS_BASE_URL = clean_env_var("RHESIS_BASE_URL", "https://api.rhesis.ai/")
```

and then on all other files you can just call the variables: `from config import RHESIS_API_KEY, RHESIS_BASE_URL`

### Proposed solution 2: Just update the documentation
Instead of changing the code, we could update the README guide to explain the correct way to set environment variables on a .env file.

akwasigroch:
This PR introduces changes from the `fix/prevent-env-loading-in-vscode` branch.

## 📝 Summary

Solves the problem of inconsistent parsing of variables stored in `.env` file. The inconsistency is caused by the difference in the way `dotenv` and VS Code parse the strings. To avoid this and other unexpected problems, it was decided to disable loading of environment variables from `.env` file in VS Code.

## 🔄 Changes

- chore: remove VS Code settings from .gitignore, prevent loading environment variables by VS Code (c222ddd)

```
.gitignore
.vscode/settings.json
```

## 📋 Commit Details

```
c222ddd - chore: remove VS Code settings from .gitignore, prevent loading environment variables by VS Code (Arkadiusz Kwasigroch, 6 minutes ago)
```

## 🔗 Related Issues

Fixes #107

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -160,9 +160,6 @@ cython_debug/
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
 #.idea/
 
-# VS Code
-#  This ignores the .vscode/ folder which is used by Visual Studio Code to store its settings.
-.vscode/
 
 # Report JSON
 #  This ignores the json report file being created
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "python.envFile": "",
+}


harry-rhesis:
Thanks for fixing this issue!
A deceptively simple error — but with lots of subtle implications and edge cases.

I agree that your solution using VSCode settings is the most elegant one. Since the problem originates from VSCode, it’s fitting (and cleanest) to solve it there as well. 😊

Issue #234, opened 2025-08-22
Improve Deployment Process by Automating Environment Variable Handling


asadaaron:
## Description

The project currently uses hardcoded `NODE_ENV` and `ENVIRONMENT` (or `ENV`) variables, which require manual updates during each deployment. This is error-prone and inefficient, especially across different environments like development, staging, and production.

This ticket aims to automate environment configuration for both the frontend and backend by dynamically loading variables from `.env.local` (frontend) and `.env` (backend).

## Expected Outcomes

✅ `NODE_ENV` and `ENVIRONMENT` / `ENV` are loaded dynamically from `.env.local` and `.env` respectively  
✅ Docker configuration uses environment variables without hardcoding  
✅ Secrets are securely managed via GitHub Secrets  
✅ Application can be deployed without manually changing env values  

## Tasks

- [x] Update frontend to load `NODE_ENV` from `.env.local`  
- [x] Update backend to load `ENVIRONMENT` or `ENV` from `.env`  
- [x] Refactor Dockerfiles to support dynamic env loading  
- [x] Update `docker-compose.yml` if necessary to pass through environment variables  
- [x] Configure GitHub Secrets with all required environment values  
- [ ] Test deployment process across different environments  
- [x] Verify application behavior remains consistent post-refactor  

## Acceptance Criteria

- [x] Env variables load correctly from `.env.local` (FE) and `.env` (BE)  
- [x] Docker files no longer require hardcoded environment settings  
- [x] GitHub Secrets are configured and used properly  
- [ ] Application deploys successfully using updated env handling  
- [x] All existing functionality remains intact post-change  

## Why this matters

Automating environment configuration ensures a more scalable, secure, and consistent deployment process across environments. It reduces manual errors and simplifies onboarding and deployment for all contributors.

PR #237, opened 2025-08-25
Automate Env Variable #234

asadaaron:
This PR introduces changes from the `automate-env-variable-#234` branch.

```
.github/workflows/frontend.yml
apps/frontend/Dockerfile
```

## 📋 Commit Details

```
33070cb - Add FRONTEND_ENV build argument to Docker container build process (Md Asaduzzaman Miah, 2025-09-03 11:44)
dc3fb5c - Update Dockerfile to define FRONTEND_ENV argument with a default value and add debugging output for its value (Md Asaduzzaman Miah, 2025-09-03 11:19)
bc2f7f5 - Enhance Dockerfile for frontend: create .next directory for non-production environments and streamline copying of .next directory during build process. (Md Asaduzzaman Miah, 2025-09-03 10:12)
```

## ✅ Checklist

- [ ] All tests pass locally


asadaaron:
> Hi @asadaaron, before merging let's implement the changes discussed in Discord:
> 
> ## Key points
> * Problem: Currently, there is an ENVIRONMENT variable in the backend, but it’s unclear if it’s used properly. There’s no consistent setup for backend, frontend, and worker environments.
> * Decision: Each service (backend, frontend, worker) should have its own environment variable (e.g., BACKEND_ENV, FRONTEND_ENV, WORKER_ENV) to allow independent configuration.
> 
> ## Implementation:
> ```
> 1.	Review how other projects handle per-service environment configuration.
> 2.	Propose a consistent naming and usage pattern for all three services.
> 3.	Ensure each service configures itself (logging, modes, etc.) based on its variable.
> 4.	Default should be production unless overridden by the user.
> ```
> 
> Next step: Research solutions and confirm approach before moving forward.
> 
> If you want you can create multiple issues for this.

I have gone through some open source projects e.g: langfuse, eigenfocus. **Langfuse** only used one Dockerfile for each service and used NODE_ENV (by default production) in their docker file ([worker/Dockerfile](https://github.com/langfuse/langfuse/blob/main/worker/Dockerfile)) as environment variable. I did not find any clear separation of FRONTEND_ENV, BACKEND_ENV or WORKER_ENV in their project. **Eigenfocus** has used different Dockerfiles for each environment e.g: Dockerfile, [Dockerfile.dev](https://github.com/Eigenfocus/eigenfocus/blob/main/Dockerfile). Then it used the Dockerfile.dev in the docker compose and dockerfile in the production. 
Proposal: For our case, we can use three different variable considering we can use it in future and it is more flexible option in any case. Please let me know your opinion @harry-rhesis .

harry-rhesis:
Hi @asadaaron, let's then go with the different variables for each service. Langfuse's case is a bit different because they one giant node app that does everything, thanks for checking!

asadaaron:
> Hi @asadaaron, let's then go with the different variables for each service. Langfuse's case is a bit different because they one giant node app that does everything, thanks for checking!

Hi @harry-rhesis 
I have introduced three new variables to separate the env. Also removed the NODE_ENV.

harry-rhesis:
Thanks for the refactor here, @asadaaron! The environment variable handling in Docker and the migration scripts is now much clearer and more consistent. This makes configuration easier to reason about and reduces the risk of hidden issues.

One suggestion for next time: try to commit more frequently, especially when touching both Docker and migration logic. Smaller, incremental commits make it easier to review changes in isolation and trace back if something breaks.

PR #272, opened 2025-09-03
Extend LLMService with Multiple Providers & Offline Support

### LINKED ISSUE #202: Feature: Extend LLMService with Multiple Providers & Offline Support
**Summary**
We want to extend the SDK’s LLMService so it can work with different LLM providers (through Mirascope) and also support offline/local usage. The goal is to provide a flexible, transparent API for end users while keeping the current service functional.

📌 **Motivation**

- Developers need the ability to switch between different LLM providers without changing how they interact with the SDK.

- Offline/local support is required for users who want to run models without depending on external services.

- We should leverage Mirascope for provider integrations to reduce boilerplate and standardize behavior.

- To minimize disruption, the existing cloud provider implementation will be moved into its own class. All current usages of LLMService will be updated to point to this class so nothing breaks during migration.

✅ **Requirements**

- [ ]  Maintain the existing LLMService interface for now.
- [ ] Introduce support for multiple providers through Mirascope.
- [ ]  Add offline/local capability.
- [ ]  Move the current internal cloud provider logic into a separate class and rewire existing references to that class.
- [ ]  Centralize error handling to keep messages consistent.

📅 **Next Steps**

1. Create a new class for the current cloud provider and update references.
2. Add Mirascope integration to support external providers.
3. Add offline/local integration.
4. Ensure backwards compatibility and update documentation.

alexeyhramov123:
This Pull requests closes #202.

- Added LiteLLM and HuggingFaceLLM providers for model inference.

- Implemented error messages for model loading and initialization.

- Developed unit tests for HuggingFaceLLM and LiteLLM to ensure functionality and error handling.

- reused code from liteLLM for OpenAI and Gemini


sdk/src/rhesis/sdk/models/providers/huggingface.py
```diff
@@ -0,0 +1,174 @@
...
+        """
+        if self.model is not None:
+            print(WARNING_MODEL_ALREADY_LOADED_RELOAD.format(self.model_name))
+        if self.tokenizer is not None:
+            print(WARNING_TOKENIZER_ALREADY_LOADED_RELOAD.format(self.model_name))
+
+        model = AutoModelForCausalLM.from_pretrained(
+            self.model_name,
+        )
+        # check for memory size
```
akwasigroch: Where is the memory size?

sdk/src/rhesis/sdk/models/providers/huggingface.py
```diff
@@ -0,0 +1,174 @@
...
+)
+from rhesis.sdk.models.base import BaseLLM
+
+
+class HuggingFaceLLM(BaseLLM):
+    """
+    A standard implementation of a model available on Hugging Face's model hub.
+    This class provides a basic structure for loading and using models from Hugging Face.
+    It can be extended to include specific models or configurations as needed.
+    A complete implementation may be needed for unusual models or configurations.
```
akwasigroch, line 21: I would just add here how to use the models from huggingface, just two lines - initialization of the model, and generating the content with the model

sdk/src/rhesis/sdk/errors.py
```diff
@@ -0,0 +1,15 @@
+# Error messages used in the SDK. This file makes it easier to find and edit the used Error messages
+
+# For this file too long lines are fine for better readability
+# flake8: noqa: E501
+
+
+# LLM Errors
+NO_MODEL_NAME_PROVIDED = "The model name is not valid. Please provide a non-empty string."
+HUGGINGFACE_MODEL_NOT_LOADED = "Hugging Face model is not loaded. Set auto_loading=True to load it manually using `load_model()`."
+WARNING_MODEL_ALREADY_LOADED_RELOAD = (
```
akwasigroch: "LOADED_RELOAD" is quite confusing, maybe just leaving WARNING_MODEL_ALREADY_LOADED should be OK?

sdk/src/rhesis/sdk/models/providers/litellm.py
```diff
@@ -0,0 +1,78 @@
...
+
+        Usage:
+            >>> llm = LiteLLM(model_name="provider/model", api_key="your_api_key")
+            >>> result = llm.generate(prompt="Tell me a joke.", system_prompt="You are funny")
+            >>> print(result)
+
+        If a Pydantic schema is provided to `generate`, the response will be validated and returned
+        as a dict.
+        """
+        self.api_key = api_key  # cannot handle environment retrieval here. Too many options
```
akwasigroch: I think LiteLLM handles retrieval of API keys from env variables itself, and when there is no variable set, it raises the error, which is quite informative.

akwasigroch (changes requested):
Hi, 
I really like your implementation. With litellm module our code is more modular, and we will have less code repetition, when introducing additional providers.


There are just minor issues that I commented, and there is one bigger issue that has to be fixed:
When trying the hugging face model:
```
from rhesis.sdk.models.providers.huggingface import HuggingFaceLLM

model = HuggingFaceLLM(model_name="google-bert/bert-base-uncased")
print(model.generate(prompt="What is the capital of France?"))
```

I get a following error:
```
Exception ignored in: <function HuggingFaceLLM.__del__ at 0x11f9bc0d0>
Traceback (most recent call last):
  File "/Users/arek/Desktop/rhesis/sdk/src/rhesis/sdk/models/providers/huggingface.py", line 52, in __del__
  File "/Users/arek/Desktop/rhesis/sdk/src/rhesis/sdk/models/providers/huggingface.py", line 81, in unload_model
ModuleNotFoundError: import of gc halted; None in sys.modules
```
Puting `import gc` on the top of the file fixes the problem. But maybe there is another, nicer solution

sdk/src/rhesis/sdk/errors.py
alexeyhramov123: That is to signal, that it will be reloaded. In the furure, there might be a use for an error, where the model will not be overwritten, which is also a valid usecase, so I wanted it to be as close to the actual message as possible

akwasigroch: What about calling it MODEL_RELOAD_WARNING then?

akwasigroch (approved):

PR #321, opened 2025-09-10
Task: LLM Service In Synthesizers

akwasigroch:
This PR introduces changes from the task/llm-service-in-synthesizers branch.

## Summary

This PR integrates LLM service functionality into the synthesizers, allowing for more flexible model handling and improved test generation capabilities.

## Key Changes

- Added LLM service integration to TestSet entity
- Updated PromptSynthesizer to accept model parameters  
- Introduced Pydantic schema for TestSet properties
- Refactored backend to use LLM service for test generation
- Updated dependencies and cleaned up unused code

## Files Changed

11 files modified including:
- Backend main.py and generation service
- SDK entities and synthesizers
- Updated dependencies in uv.lock
- Cleaned up issue templates

## Commits

- f3101e7: introduce pydantic schema
- 08282fb: add missing import
- 32f1207: fix testset generation
- f724ff1: add LLM Service to TestSet
- f99ddad: fix
- 11c897b: changed backend to use llm service - promptsynth accepting models


apps/backend/src/rhesis/backend/app/services/generation.py
harry-rhesis, line 1: I would frame this as an SDK-wide default that can be set by clients, and maybe call it DEFAULT_LANGUAGE_MODEL or so for clarity. 

The background: I find it a little bit cumbersome to have to provide the model on all calls, if I know that the model I need is gemini. By having a SDK-wide, user-defined default, we can spare the users this pain. Same goes for the metrics.

harry-rhesis: I am referring to `DEFAULT_MODEL`.

sdk/src/rhesis/sdk/synthesizers/prompt_synthesizer.py
harry-rhesis, line 1: The `PromptSynthesizer` should also leverage the nice schema capabilities. 🙌 Currently it has code that is required only because of the lack of a schema.

sdk/src/rhesis/sdk/synthesizers/utils.py
harry-rhesis, line 1: `create_test_set` seems a bit off as part of utils. Should it be part of the TestSet entity?

apps/backend/src/rhesis/backend/app/services/generation.py
akwasigroch: I agree, right now, when the user doesn't specify the model, the default model is used (the model that we are serving in the backend). I agree that we should have a functionality that the user can set the model for every place where it needs to be used, but this is the topic for another PR as it touches many areas (Synthesizers, metrics) and it also needs additional code in the LLM Service.

sdk/src/rhesis/sdk/synthesizers/prompt_synthesizer.py
akwasigroch: We have decided that LLM Service implementation in syntehsizers would be done in 2 steps:
1. Replacement of old LLM Service with the new one
2. Introduction of schemas, to limit the size of the prompt.
The introduction of schemas would go in the next PR

sdk/src/rhesis/sdk/synthesizers/utils.py
akwasigroch: I agree that the function should not be part of the utils, but it is also the topic for another PR and refactoring (entities in SDK). I didn't touch it as it is not in the scope of this PR.

PR #453, opened 2025-10-02
fix: resolve duplicate key error in Run Test Drawer with same project names

### LINKED ISSUE #442: Run Test Drawer throws an error if there is a project with the same name
The Run Test Drawer throws an error if there is a project with the same name.

## Steps to Reproduce
	1.	Open the Test feature / Run Test Drawer.
	2.	Create (or already have) multiple projects with the same name.
	3.	Try to run a test against such a project.
	4.	Observe: An error is thrown.

## Expected Behavior

Tests should run successfully, even if multiple projects share the same name.

## Actual Behavior

An error is thrown because the project list uses name as the key.

## Screenshots / Recordings

Image

## Environment
	•	Browser: (e.g., Chrome 128.0.6613.119)
	•	OS: (e.g., macOS 14.5)
	•	Network: Wi-Fi
	•	Rhesis AI version: (e.g., 1.4.2)

nicolai-rhesis:
## Summary

Fixes the issue where the Run Test Drawer throws an error when there are multiple projects with the same name.

## Root Cause

The issue was caused by Material-UI's Autocomplete components missing the `renderOption` prop. When this prop is not provided, Material-UI defaults to using the option's string representation (the project name) as the React key, which causes duplicate key errors when multiple projects have the same name.

## Changes Made

Added missing `renderOption` prop to Autocomplete components in:
- `TrialDrawer.tsx` - The main Run Test Drawer component
- `GenerateTestsStepper.tsx` - Test generation component  
- `TestRunDrawer.tsx` - Test run management component
- `ExecuteTestSetDrawer.tsx` - Test set execution component

Each `renderOption` implementation now uses `option.id` as the React key instead of the project name:

```tsx
renderOption={(props, option) => {
  const { key, ...otherProps } = props;
  return (
    <Box component="li" key={option.id} {...otherProps}>
      {option.name}
    </Box>
  );
}}
```

## Expected Behavior After Fix

- ✅ Tests can run successfully even when multiple projects share the same name
- ✅ No duplicate key errors will be thrown
- ✅ Project selection works correctly in all affected components
- ✅ The UI continues to display project names to users while using unique IDs internally

## Testing

- [x] Linting passes with no new errors
- [x] TypeScript compilation successful
- [x] No breaking changes to existing functionality

Resolves #442

### DIFF
--- a/apps/frontend/next-env.d.ts
+++ b/apps/frontend/next-env.d.ts
@@ -1,6 +1,5 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
-/// <reference path="./.next/types/routes.d.ts" />
 
 // NOTE: This file should not be edited
 // see https://nextjs.org/docs/app/api-reference/config/typescript for more information.
--- a/apps/frontend/src/app/(protected)/test-runs/components/TestRunDrawer.tsx
+++ b/apps/frontend/src/app/(protected)/test-runs/components/TestRunDrawer.tsx
@@ -377,6 +377,15 @@ export default function TestRunDrawer({
               value={project}
               onChange={(_, newValue) => setProject(newValue)}
               getOptionLabel={option => option.name}
+              isOptionEqualToValue={(option, value) => option.id === value.id}
+              renderOption={(props, option) => {
+                const { key, ...otherProps } = props;
+                return (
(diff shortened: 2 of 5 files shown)


nicolai-rhesis:
@harry-rhesis this is not a must have for this release. Decide if you like to merge it or not.

Issue #531, opened 2025-10-09
Organization name overflow in application menu interferes with collapsible control


harry-rhesis:
## Summary
Long organization names in the application menu override the collapsible control, making it difficult or impossible to interact with the menu collapse button.

<img width="244" height="209" alt="Image" src="https://github.com/user-attachments/assets/fec4a4ad-b0ba-4cc4-84ba-b1602b6700de" />

## Steps to Reproduce
1. Create or select an organization with a long name
2. Navigate to any page in the application
3. Observe the application menu (sidebar)
4. Note that the organization name overlaps/overrides the collapsible control

## Expected Behavior
Organization names should be truncated with ellipsis (e.g., "Very Long Organization Na...") to ensure the collapsible control remains accessible and visible.

## Actual Behavior
Long organization names extend beyond their container and override the collapsible control button.

## Impact
- Frequency: Occurs with any organization name exceeding the available space
- Impact: Reduces usability, makes menu collapse/expand functionality inaccessible

## Acceptance Criteria (testable)
- [ ] Organization names are truncated with ellipsis when they exceed available space
- [ ] Collapsible control remains fully visible and clickable
- [ ] Hovering over truncated names shows full organization name in a tooltip
- [ ] Works consistently across different screen sizes

PR #626, opened 2025-10-20
Fix: #531 Organisation Name Overlaps

conradkirschner:
This PR should fix the overlapping organisation name in the navigation mentioned in the issue #531 

It implements the tooltip on top of the organisation name when exeeding the width limit (15rem/240px) 

Side Note:
Having the server action for the protected navigation is not clean, since it is not related to #531 I left it as it was.
Also because of this I did not extracted the width to a flexible variable. (That would be imported here)


nicolai-rhesis:
@conradkirschner thanks so much for contributing to our project! Keep the good work coming 👍

PR #635, opened 2025-10-21
282 fixes width of project select box

### LINKED ISSUE #282: Project selector full-screen width in Endpoint creation dialog
## Summary
The project selection field in the Endpoint creation screen should not open full width.

## Steps to Reproduce

Example:
1. Go to https://app.rhesis.ai
2. Navigate to Endpoints
3. Create new endpoint
4. Select a project
5. Observe: The available project selection dialog opens full-width

## Expected Behavior
It opens in the same width as the field. Also, there should not be any descriptive text be shown.

## Screenshots / Recordings

Image

## Environment
DV, Prod

Example:
- [ ] The selection popup opens in the same width as the field.

conradkirschner:
This PR will fix #282 

I decided against unifying the form usage of the project selection since it should keep in the issue scope. 

Preferable I would like to add variant=Selection|Autoselect and use it in the other forms.

### DIFF
--- a/apps/frontend/src/app/(protected)/endpoints/components/EndpointDetail.tsx
+++ b/apps/frontend/src/app/(protected)/endpoints/components/EndpointDetail.tsx
@@ -482,7 +482,7 @@ export default function EndpointDetail({
                 projectId={
                   isEditing ? editedValues.project_id : endpoint.project_id
                 }
-                onChangeAction={nextProjectId => {
+                onChange={nextProjectId => {
                   handleChange('project_id', nextProjectId);
                 }}
               />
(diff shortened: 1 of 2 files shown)


nicolai-rhesis (changes requested):
Hi @conradkirschner! 👋
Thank you for taking the time to work on issue #282 - we really appreciate external contributions to the project!

I've reviewed the changes and while the new component architecture is well-structured, I think we might be able to solve this width issue with a much simpler approach. The root cause seems to be that the project descriptions are making the dropdown extremely wide when they contain long text.

Instead of creating a new component system, could we try one of these lighter approaches first.

Option 1: Remove descriptions from dropdown

<ListItemText
  primary={project.name}
  // Remove secondary={project.description}
/>

Option 2: Truncate descriptions

<ListItemText
  primary={project.name}
  secondary={project.description?.length > 50 
    ? `${project.description.substring(0, 50)}...` 
    : project.description
  }
/>

This would keep the existing codebase simpler and achieve the same UX goal with just a few lines of change. What do you think? Would you be willing to try one of these approaches instead?

Thanks again for your contribution! 🙏

PR #654, opened 2025-10-23
fix: clear task reference when comment is deleted

### LINKED ISSUE #469: Task reference to associated comment remains even if comment is deleted
## Summary
When you are in a task, and it shows the "go to associated comment" chip at the top, this should always work. However, when the comment was removed in the meantime, the task's reference to the comment is not removed.

## Steps to Reproduce
Go here, click on associated comment link, try to find the comment: https://stg-app.rhesis.ai/tasks/f0dc0fb6-aa77-4e86-9502-e53bc5e0319f

## Expected Behavior
If the referenced comment was deleted, the link to go to the associated comment should no longer be shown on a task

Image

## Actual Behavior
Tasks still link to comments that do not even exist anymore.

nicolai-rhesis:
## Summary

Fixes #469

When a comment is deleted, tasks that reference that comment in their `task_metadata.comment_id` field now have that reference automatically cleared. This prevents orphaned references that would show an incorrect 'Go to associated comment' link.

### Backend
- Modified `delete_comment()` in `crud.py` to clear `comment_id` from `task_metadata` before deletion
- Added SQL query to find and update all tasks referencing the deleted comment
- Used PostgreSQL's JSON operator `-` to remove the specific key while preserving other metadata fields

### Tests
- Added `test_delete_comment_clears_task_references` to verify the fix works correctly
- Test creates a task with a comment reference, deletes the comment, and verifies the reference is cleared
- Test also verifies other metadata fields remain intact

## Testing

The new test can be run with:
```bash
pytest tests/backend/routes/test_comment.py::TestCommentEntityRelationships::test_delete_comment_clears_task_references -v
```

## Technical Details

The fix uses a PostgreSQL UPDATE query with JSON operators to efficiently remove the `comment_id` key from all affected tasks:
```sql
UPDATE task
SET task_metadata = task_metadata - 'comment_id',
    updated_at = NOW()
WHERE task_metadata->>'comment_id' = :comment_id
AND organization_id = :organization_id
```

This ensures:
- Only tasks referencing the deleted comment are updated
- Other fields in `task_metadata` remain untouched
- Organization isolation is maintained
- The `updated_at` timestamp reflects the change

### DIFF
--- a/apps/backend/src/rhesis/backend/app/crud.py
+++ b/apps/backend/src/rhesis/backend/app/crud.py
@@ -2644,14 +2644,16 @@ def delete_comment(
         # Clear comment_id from task_metadata using SQLAlchemy JSONB operators
         # Cast JSON to JSONB to use the '-' operator for key removal
         db.query(models.Task).filter(
-            models.Task.task_metadata['comment_id'].astext == str(comment_id),
-            models.Task.organization_id == organization_id
+            models.Task.task_metadata["comment_id"].astext == str(comment_id),
+            models.Task.organization_id == organization_id,
         ).update(
             {
-                models.Task.task_metadata: cast(models.Task.task_metadata, JSONB).op('-')('comment_id'),
-                models.Task.updated_at: func.now()
+                models.Task.task_metadata: cast(models.Task.task_metadata, JSONB).op("-")(
+                    "comment_id"
+                ),
+                models.Task.updated_at: func.now(),
             },
-            synchronize_session=False
+            synchronize_session=False,
         )
 
         # Commit the task metadata updates before deleting the comment


harry-rhesis:
Thanks for tackling this @nicolai-rhesis! Good catch on fixing the orphaned comment references.

The implementation looks solid already. One suggestion for the current implementation:

## Using More Idiomatic SQLAlchemy

This could be achieved using SQLAlchemy's JSONB operations instead of raw SQL. The session would automatically handle the commit:

```python
from sqlalchemy import func

# Clear comment_id from task_metadata using SQLAlchemy JSONB operators
db.query(Task).filter(
    Task.task_metadata['comment_id'].astext == comment_id,
    Task.organization_id == organization_id
).update(
    {
        Task.task_metadata: Task.task_metadata.op('-')('comment_id'),
        Task.updated_at: func.now()
    },
    synchronize_session=False
)
# Session automatically handles commit
```

This approach:
- Keeps everything in SQLAlchemy's ORM layer
- Leverages existing session management (no manual commit needed)
- Maintains better portability and type safety
- Makes the intent clearer in the codebase

## Broader Pattern

This fix highlights a general pattern: when entities are soft-deleted, any JSON metadata references to them should be cleaned up. Looking at the `Task` model, `entity_type` can be `Test`, `TestSet`, `TestRun`, or `Comment` - all of which support soft deletion.

The codebase already has a robust cascade system (`services/cascade.py` + `config/cascade_config.py`) for parent-child relationships. This pattern could be extended to handle metadata cleanup generically - I've created **issue #673** to track implementing that architectural improvement.

For now, merging this as-is (or with the SQLAlchemy refactor) is perfectly fine! The generic system can be added later. Thank you!

PR #778, opened 2025-11-04
feat: Multi-Turn Test Configuration UI

harry-rhesis:
## Summary
This PR implements the complete Multi-Turn test configuration UI, allowing users to create and manage multi-turn tests with configurable parameters.

### Multi-Turn Configuration UI
- Implemented conditional rendering to show multi-turn fields when test type is "Multi-Turn"
- Hide Test Prompt and Expected Response fields for multi-turn tests
- Created EditableField component with edit button UI pattern matching single-turn fields
- Added optional fields (Instructions, Restrictions, Scenario, Max. Turns) with add/remove functionality
- Included helper text for all fields explaining their purpose
- Added Remove buttons for optional fields positioned next to Edit button
- Renamed 'Test Executable' to 'Test Prompt' with explanatory helper text

image

### Tests Data Grid Updates
- Replaced 'Category' column with 'Test Type' column displaying test type value
- Modified 'Content' column to show Goal for multi-turn tests instead of prompt content
- Added color-coded chips for test types (primary for Single-Turn, secondary for Multi-Turn)
- Updated column header to 'Content' for clarity

image

### Test Set Data Grid Updates
- Replaced text from 'Prompt' to 'Content' 
- Introduced test type as part of the grid

image

### Code Quality
- Used theme-based styles instead of hard-coded values for consistency
- Validated Goal field is not empty before saving multi-turn configuration
- Ran Prettier formatting on all modified files
- Removed unused components

## Testing
- Multi-turn test configuration can be created and edited
- Optional fields can be added and removed
- Test grid properly displays test types and goal/content based on type
- All styling uses theme values for consistency

### DIFF
--- a/apps/frontend/src/app/(protected)/test-sets/[identifier]/components/TestSetTestsGrid.tsx
+++ b/apps/frontend/src/app/(protected)/test-sets/[identifier]/components/TestSetTestsGrid.tsx
@@ -15,7 +15,10 @@ import { TestSetsClient } from '@/utils/api-client/test-sets-client';
 import { useNotifications } from '@/components/common/NotificationContext';
 import DeleteIcon from '@mui/icons-material/Delete';
 import { isMultiTurnTest } from '@/constants/test-types';
-import { isMultiTurnConfig } from '@/utils/api-client/interfaces/multi-turn-test-config';
+import {
+  getTestContentValue,
+  renderTestContentCell,
+} from '@/app/(protected)/tests/components/test-grid-helpers';
 
 interface TestSetTestsGridProps {
   sessionToken: string;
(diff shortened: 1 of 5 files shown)


nicolai-rhesis (changes requested):
This is pretty nice! Well done, Harry! Thanks for moving this forward so quickly.

A few thing I noticed:

**TEST GRID**
- In the tests grid, the chips for test type should not be blue and orange, keep em neutral as the others eg for category and topic. 
- The first column is called content but what is content in the case of a single-turn (prompt?) test and in case of a multi-turn test (goal?)?

**TEST DETAIL PAGE**
- the max turn field has a lower limit (1) but no upper limit; I don't think we should go deeper than (99) to avoid unwanted costs.
- the goal, scenario and restriction fields do not have input limitations. Are there DB based limits? Otherwise I would restrict it to 10000 chars.
- When I switch a single-turn test to a multi-turn test, and then save the goal, the title of the test does not update. This should update to the goal text though.
image
- When making it this flexible to switch between single-turn and multi-turn, users might execute a test as single turn and then later as multi-turn; this will lead to complications in the test execution (how do you execute it?) and in the results presentation; I believe we should always only have one initial type and keep it (ie. make it none-switchable).
- Allowing switching also causes problems in the grid, where it pulls the test executable text for multi-turn tests if no goal is provided: 
image

- When I remove an optional field, and later add it again, the old value remains, is this intended? It feels unnatural to me. If I removed the field and add it later, I would expect it to start in a cleared state.

**Test Set Data Grid**
 - The test-type column does not show up for me when running the branch locally, I see it worked in your screenshot but it's not showing for me. 
 

 **Test Set Detail Page**
 - we do allow mixed test sets now. I think today we argued against this: 
image
- Test sets need a field for test type now in the details section



**CLAUDE FEEDBACK:**
⚠️ Issues to Address
🔴 HIGH - Code Duplication (must fix)
The content column logic (showing goal vs. prompt) is duplicated across TestsGrid.tsx (lines 138-178) and TestSetTestsGrid.tsx (lines 98-141)
Action: Extract into reusable grid column helper functions (getTestContentColumn(), getTestTypeColumn())
🟡 MEDIUM - Consider Fixing
isMultiTurnConfig() type guard only checks for 'goal' in config - doesn't validate field types
Error handling in EditableField.handleConfirmEdit() - component stays in edit mode on error with no user feedback

PR #790, opened 2025-11-06
feat(auth): remove Microsoft and Apple login options

nicolai-rhesis:
Removed Microsoft and Apple as login options from the authentication page. The login page now only offers:
- Email sign-in
- Google OAuth
- GitHub OAuth


harry-rhesis:
Thanks @nicolai-rhesis, that's the type of attention to detail that we need. @asadaaron, thanks for flagging it!

PR #796, opened 2025-11-06
fix(metrics): restrict detail page access to rhesis metrics only

nicolai-rhesis:
## Changes
- Hide detail page arrow icon for non-rhesis metrics in both tabs (SelectedMetricsTab and MetricsDirectoryTab)
- Add redirect for non-rhesis metrics when accessing detail page directly via URL
- Show warning notification when attempting to access non-rhesis metric details

## Behavior
- Arrow icon (OpenInNewIcon) now only appears on Rhesis metrics
- Users trying to navigate to non-Rhesis metric detail pages are redirected back to /metrics with a warning
- All other metrics (custom, deepeval, framework) are prevented from accessing the detail page

### DIFF
--- a/apps/frontend/src/app/(protected)/metrics/[identifier]/page.tsx
+++ b/apps/frontend/src/app/(protected)/metrics/[identifier]/page.tsx
@@ -124,8 +124,9 @@ export default function MetricDetailPage() {
           setModels([metricData.model]);
         }
 
-        // Check if this is NOT a rhesis metric - redirect non-rhesis metrics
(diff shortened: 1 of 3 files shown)


harry-rhesis:
Thanks @nicolai-rhesis, excellent catch; those metrics should not be manipulated by users indeed! This also needs to be blocked at the backend level. Let's merge this one for now; I created another issue, so we do not forget to do it at some point: https://github.com/rhesis-ai/rhesis/issues/798

PR #811, opened 2025-11-11
Polyphemus for google cloud

alexeyhramov123:
First full benchmarking for polyphemus. Ready for running on Google Cloud.


sdk/src/rhesis/sdk/models/providers/huggingface.py
```diff
@@ -131,14 +137,42 @@ def unload_model(self):
...
+            total_bytes = param_size + buffer_size
+            return round(total_bytes / (1024**3), 3)  # Convert to GB
+        except Exception:
+            return 0.0
+
     def generate(
         self,
         prompt: str,
         system_prompt: Optional[str] = None,
+        schema: Optional[BaseModel] = None,
```
akwasigroch: Here, should be Type[BaseModel], because we are passing the class, not the object as a schema

PR #890, opened 2025-11-20
Fix incorrect section name in source selector message

akwasigroch:
## Purpose
Fix incorrect section name reference in the source selector component. The alert message was referring to "Sources section" when it should reference "Knowledge section" to match the actual UI.

## What Changed
- Updated alert message in SourceSelector component to reference "Knowledge section" instead of "Sources section"

## Additional Context
- This is a simple text correction to improve user experience and accuracy


harry-rhesis:
Agreed, as per our discussion, @akwasigroch.

PR #963, opened 2025-11-28
Fix Dockerfile build optimization and add UV link mode

akwasigroch:
## Purpose
Fix Docker build issues by adding the required UV_LINK_MODE environment variable and optimize layer caching by moving Bun binary copies earlier in the build stage.

## What Changed
- Added `UV_LINK_MODE=copy` environment variable in Dockerfile (required by the cache)
- Moved Bun binary COPY instructions earlier in the runtime stage for better layer caching
- Applied same Bun copy optimization to Dockerfile.dev

## Additional Context
- These changes improve Docker build reliability and caching efficiency
- The UV_LINK_MODE=copy setting is required when using uv with mounted cache volumes

### DIFF
--- a/apps/backend/Dockerfile
+++ b/apps/backend/Dockerfile
@@ -39,6 +39,8 @@ ENV UV_COMPILE_BYTECODE=1
 # the size of the image.
 ENV UV_NO_MANAGED_PYTHON=1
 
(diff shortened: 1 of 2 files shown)


harry-rhesis:
@akwasigroch thanks for that!

PR #1001, opened 2025-12-04
Fix onboarding checklist collapsing during active tours

nicolai-rhesis:
## Purpose
Fix the issue where the onboarding checklist remains expanded when users are actively participating in an onboarding tour. The checklist should stay collapsed during tours to avoid UI clutter and keep users focused on the guided experience.

## What Changed
- Added `useSearchParams` hook to detect when a tour is active via the `tour` query parameter
- Updated auto-collapse logic to keep the checklist collapsed when `tour` parameter is present in the URL
- The checklist now collapses in two scenarios:
  1. When on the dashboard page (existing behavior)
  2. When in an active onboarding tour (new behavior)
- Fixed duplicate import in TestRunPerformance.tsx (linter error)

## Additional Context
- When users click on an onboarding step, they navigate to pages like `/projects/{id}?tour=endpoint-setup`
- This fix ensures the onboarding checklist automatically collapses and stays collapsed during the tour
- Improves UX by reducing visual distraction during guided onboarding flows

## Testing
1. Start the onboarding flow
2. Click on any onboarding step (e.g., "Create your first endpoint")
3. Verify the onboarding checklist collapses when navigating to the tour page
4. Verify the checklist remains collapsed while the tour is active
5. Verify users can still manually expand it if needed

### DIFF
--- a/apps/frontend/src/components/onboarding/OnboardingChecklist.tsx
+++ b/apps/frontend/src/components/onboarding/OnboardingChecklist.tsx
@@ -59,7 +59,8 @@ export default function OnboardingChecklist() {
 
   // Auto-collapse on dashboard page or when in an onboarding tour (avoid hydration issues)
   useEffect(() => {
-    const isInTour = searchParams.get('tour') !== null;
+    const tourParam = searchParams.get('tour');
+    const isInTour = tourParam !== null && tourParam.length > 0;
     const shouldCollapse = pathname === ONBOARDING_COLLAPSE_PATH || isInTour;
     setExpanded(!shouldCollapse);
   }, [pathname, searchParams]);


apps/frontend/src/components/onboarding/OnboardingChecklist.tsx
```diff
@@ -56,10 +57,12 @@ export default function OnboardingChecklist() {
     setMounted(true);
   }, []);
 
-  // Auto-collapse on dashboard page (avoid hydration issues)
+  // Auto-collapse on dashboard page or when in an onboarding tour (avoid hydration issues)
   useEffect(() => {
-    setExpanded(pathname !== ONBOARDING_COLLAPSE_PATH);
-  }, [pathname]);
+    const isInTour = searchParams.get('tour') !== null;
```
asad-miah: The current check:
const isInTour = searchParams.get('tour') !== null;
This treats ?tour= as a tour. If that’s intended, it’s fine. If you want to require a non-empty value:
const isInTour = searchParams.get('tour') !== null && searchParams.get('tour') !== '';

Or more concisely:
const tourParam = searchParams.get('tour');const isInTour = tourParam !== null && tourParam.length > 0;

nicolai-rhesis: Good catch! I've updated the code to check for a non-empty tour parameter value using your suggested approach:

const tourParam = searchParams.get('tour');
const isInTour = tourParam !== null && tourParam.length > 0;This prevents edge cases where `?tour=` (empty value) would incorrectly trigger the collapse. Changes pushed in commit edd8397.

PR #1033, opened 2025-12-10
Fix: Source Detail Cursor Focus

EmanueleDeRossi1:
This PR introduces changes from the `fix/source-detail-cursor-focus` branch.

## 📝 Summary

When editing the title or description on the source detail page, users could only type one character at a time before the cursor would disappear, requiring them to click the field again to continue typing.

### Solution
Converted the title and description input fields from controlled components (using `value` and `onChange` with state) to uncontrolled components (using `defaultValue` and `inputRef`). This matches the pattern already used successfully in the metrics detail page.

## 📁 Files Changed (       1 files)

```
apps/frontend/src/app/(protected)/knowledge/[identifier]/components/SourcePreviewClientWrapper.tsx
```

## 📋 Commit Details

```
d29926d4 - fix(source): resolve cursor focus loss in title and description fields (Emanuele De Rossi, 2025-12-10 15:08)
```

### DIFF
--- a/apps/frontend/src/app/(protected)/knowledge/[identifier]/components/SourcePreviewClientWrapper.tsx
+++ b/apps/frontend/src/app/(protected)/knowledge/[identifier]/components/SourcePreviewClientWrapper.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import React, { useState, useCallback } from 'react';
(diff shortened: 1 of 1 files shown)


harry-rhesis:
Excellent catch, @EmanueleDeRossi1, thanks for fixing this!

PR #1211, opened 2026-01-29
Fix: Allow nested objects in response_mapping

harry-rhesis:
## Purpose
Fix validation error when editing endpoint's response_mapping with nested metadata objects.

## What Changed
- Changed `response_mapping` type from `Dict[str, str]` to `Dict[str, Any]` in endpoint schema
- Updated `ResponseMapper` to recursively process nested dictionaries and lists
- Updated TypeScript interfaces for frontend consistency
- Added `.cursor/plans/` and `docs/src/public/_pagefind/` to gitignore

## Additional Context
The error occurred when trying to save a nested metadata mapping like:
```json
{
  "metadata": {
    "model": "$.model",
    "object": "$.object"
  }
}
```

Downstream code (response_extractor, penelope_target) already handles both string and dict metadata types, so this change is safe.


harry-rhesis:
Recreating PR with correct base - branch was out of sync

PR #1229, opened 2026-01-30
fix(frontend): duplicate breadcrumb routes

thisisharsh7:
Before - 
Screenshot 2026-01-30 at 7 08 20 PM

After - 
Screenshot 2026-01-30 at 7 17 16 PM

Screenshot 2026-01-30 at 7 17 06 PM


harry-rhesis:
@thisisharsh7, thanks for the patience, we checked, and looks much nicer! We are now in the process of revamping our design system, so this will be handled with a more principled approach.

PR #1235, opened 2026-02-01
feat(frontend): add change detection to editable components

thisisharsh7:
This PR introduces a reusable `useFormChangeDetection` and several changes across various text fields and form to  allow save changes only when there is a change until then keep the button disabled. This saves from unnecessary API calls.

Integrated change detection into:

- `ContactInformationForm`
Screenshot 2026-02-02 at 8 21 17 PM

- `OrganizationDetailsForm`
Screenshot 2026-02-02 at 8 21 11 PM

- `BehaviorDrawer`
Screenshot 2026-02-02 at 8 21 59 PM

- `ProjectDrawer`
Screenshot 2026-02-02 at 8 21 47 PM

- `TestExecutableField`
Screenshot 2026-02-02 at 8 35 10 PM

- `MetricEditSection`
Screenshot 2026-02-02 at 8 42 20 PM

### DIFF
--- a/apps/frontend/src/hooks/useFormChangeDetection.ts
+++ b/apps/frontend/src/hooks/useFormChangeDetection.ts
@@ -1,11 +1,11 @@
-import { useState, useEffect } from 'react';
+import { useState, useEffect, useMemo } from 'react';
 
 function normalizeFormValue(value: string | undefined | null): string {
   return (value || '').trim();
 }
 
 function compareFormData<
-  T extends Record<string, string | number | boolean | undefined | null>
+  T extends Record<string, string | number | boolean | undefined | null>,
 >(current: T, initial: T): boolean {
   const keys = Object.keys(current) as Array<keyof T>;
 
@@ -13,7 +13,10 @@ function compareFormData<
     const currentValue = current[key];
     const initialValue = initial[key];
 
-    if (typeof currentValue === 'string' || typeof initialValue === 'string') {
+    const isStringLike = (val: unknown): val is string | null | undefined =>
+      typeof val === 'string' || val === null || val === undefined;
+
+    if (isStringLike(currentValue) && isStringLike(initialValue)) {
       return (
         normalizeFormValue(currentValue) !== normalizeFormValue(initialValue)
       );
@@ -34,16 +37,21 @@ interface UseFormChangeDetectionReturn {
 }
 
 export function useFormChangeDetection<
-  T extends Record<string, string | number | boolean | undefined | null>
+  T extends Record<string, string | number | boolean | undefined | null>,
 >({
   initialData,
   currentData,
 }: UseFormChangeDetectionOptions<T>): UseFormChangeDetectionReturn {
   const [trackedInitialData, setTrackedInitialData] = useState<T>(initialData);
 
+  const initialDataString = useMemo(
+    () => JSON.stringify(initialData),
+    [initialData]
+  );
+
   useEffect(() => {
     setTrackedInitialData(initialData);
-  }, [JSON.stringify(initialData)]);
+  }, [initialDataString, initialData]);
 
   const hasChanges = compareFormData(currentData, trackedInitialData);
 


apps/frontend/src/hooks/useFormChangeDetection.ts
```diff
@@ -0,0 +1,54 @@
...
+interface UseFormChangeDetectionReturn {
+  hasChanges: boolean;
+  resetChanges: () => void;
+}
+
+export function useFormChangeDetection<T extends Record<string, any>>({
+  initialData,
+  currentData,
+}: UseFormChangeDetectionOptions<T>): UseFormChangeDetectionReturn {
+  const initialDataRef = useRef<T>(initialData);
```
harry-rhesis: Nice work on the hook, @thisisharsh7! Small cleanup suggestion: `initialDataRef` is declared here but never read anywhere in the hook, and the `useRef` import (line 1) is only needed for it. Would you mind removing both to keep things tidy?

apps/frontend/src/hooks/useFormChangeDetection.ts
```diff
@@ -0,0 +1,54 @@
...
+  const [trackedInitialData, setTrackedInitialData] = useState<T>(initialData);
```
harry-rhesis, line 45: Good catch using `useState` here to track the initial snapshot, @thisisharsh7. One thing to consider: if the parent re-fetches data from the API (e.g. after another user updates the record), `initialData` will change but `trackedInitialData` will still hold the stale value from mount time — which could cause false-positive "has changes" reports.

A `useEffect` that syncs `trackedInitialData` when `initialData` actually changes would make the hook more robust:

```ts
useEffect(() => {
  setTrackedInitialData(initialData);
}, [JSON.stringify(initialData)]);
```

(Using `JSON.stringify` for a stable comparison since `initialData` is a new object on each render. Alternatively a deep-equal utility would work.)

apps/frontend/src/hooks/useFormChangeDetection.ts
```diff
@@ -0,0 +1,54 @@
...
+    const initialValue = initial[key];
+
+    if (typeof currentValue === 'string' || typeof initialValue === 'string') {
+      return (
+        normalizeFormValue(currentValue) !== normalizeFormValue(initialValue)
+      );
+    }
+
+    return currentValue !== initialValue;
+  });
```
harry-rhesis, line 26: Thanks for making this generic, @thisisharsh7 — the type signature `Record<string, any>` suggests this hook could be used with any field type, but the comparison on line 23 (`currentValue !== initialValue`) uses strict reference equality. This works well for strings, booleans, and numbers, but would silently report false positives for arrays or objects (since they're new references on every render, even if the content is identical).

This isn't a bug today since all current callers pass simple primitives, but it could become a subtle trap for future usage. A couple of options:

1. Add a `JSON.stringify` fallback for non-primitive values:
```ts
if (typeof currentValue === 'object' && typeof initialValue === 'object') {
  return JSON.stringify(currentValue) !== JSON.stringify(initialValue);
}
```

2. Or narrow the generic constraint to signal that only primitives are supported:
```ts
T extends Record<string, string | boolean | number | undefined | null>
```

Either approach would make the intent clearer.

apps/frontend/src/hooks/useFormChangeDetection.ts
thisisharsh7: thanks will do

apps/frontend/src/hooks/useFormChangeDetection.ts
thisisharsh7: got it

apps/frontend/src/hooks/useFormChangeDetection.ts
thisisharsh7: will update

harry-rhesis:
@thisisharsh7 thanks for the changes, appreciate it! Merging this PR!

PR #1237, opened 2026-02-02
DB Vector Storage

EmanueleDeRossi1:
## Purpose

Add dedicated vector embedding storage with multi-dimension support for semantic search and retrieval.

## What Changed

**New Components:**
- **Embedding Model** (`embedding.py`): New table with separate columns for 384, 768, 1024, and 1536 dimension vectors (3072 excluded due to pgvector HNSW index limitations)
- **EmbeddableMixin** (`mixins.py`): Mixin providing `to_searchable_text()` interface for entities supporting vector embeddings
- **Migration** (`a1b2c3d4e5f8_create_embedding_table.py`): Creates embedding table with HNSW indexes and full-text search support

**Modified Components:**
- **Test & Source Models**: Added `EmbeddableMixin` and polymorphic `embeddings` relationship
- **Models Init**: Exported new `Embedding` model

## Key Features

- **Type Safety**: PostgreSQL constraint ensures exactly one vector column is populated per record
- **High Performance**: HNSW indexes (m=8-12, ef_construction=16-24) for fast similarity search
- **Full-Text Search**: PostgreSQL TSVECTOR with GIN index on searchable text
- **Traceability**: Tracks model_id, config_hash, text_hash, and status for each embedding
- **Polymorphic Design**: Single table supports Test, Source, and future entity types

## Additional Context

- Adds `pgvector>=0.3.0` dependency
- Vector dimensions 384, 768, 1024, 1536 supported
- Conservative HNSW parameters for 2-4x faster builds while maintaining query performance

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -71,7 +71,6 @@ services:
   # PostgreSQL Database
   postgres:
     image: mirror.gcr.io/pgvector/pgvector:pg16
-    container_name: rhesis-postgres
     environment:
       POSTGRES_DB: ${SQLALCHEMY_DB_NAME:-rhesis-db}
       POSTGRES_USER: ${SQLALCHEMY_DB_USER:-rhesis-user}


harry-rhesis:
@EmanueleDeRossi1 thanks for putting together this PR! The backs and forths have been many, but I guess we have a nice architecture in place. Now merging.

PR #1313, opened 2026-02-11
Use flat schema for synthesizer batch generation

akwasigroch:
## Purpose
Refactor synthesizer batch generation in the SDK to use a flat request/response schema and repack to nested structures internally.

## What Changed
- **SDK**: Refactored batch generation to use flat schema and repack to nested format

## Testing
- Run SDK tests from `sdk`: `make test`


harry-rhesis:
@akwasigroch thanks for swiftly tackling this issue!

PR #1345, opened 2026-02-14
fix(hooks): resolve infinite loop in useFormChangeDetection

thisisharsh7:
Fix infinite re-renders in useFormChangeDetection

This PR fixes a regression introduced in #1235.

The useEffect inside useFormChangeDetection had redundant dependencies (initialData and initialDataString), causing unnecessary executions and in some cases infinite re-renders when initialData was recreated by reference.


harry-rhesis:
@thisisharsh7, thanks for the fix; this behavior was also flagged during development of another feature. We fixed it in another branch. Thank you for checking, though!

PR #1489, opened 2026-03-12
Fix broken filter layout on metrics overview page

nicolai-rhesis:
## Purpose
Fix the broken filter UI on the metrics overview page where the filter chips (All, Custom, Garak, Rhesis, Deepeval, Ragas) and action buttons (New Metric, Filters) overlapped and rendered in a floating card-like element instead of a clean single row.

Bildschirmfoto 2026-03-12 um 13 08 09

## Root Cause
`SearchAndFilterBar` placed the filter children (ButtonGroup) inline with the search field inside a shared flex row. When many backend filter options are present, the ButtonGroup exceeded the available space, causing the Filters badge button and the New Metric action button to wrap into a floating card that overlapped the page content below.

## What Changed
- Restructured `SearchAndFilterBar` layout into two explicit rows:
  - **Row 1**: Search field + action buttons (New Metric / Reset) with `justify-content: space-between`
  - **Row 2**: Filter children (chips, Filters button) with full-width flex wrap
- Removed the intermediate left/right split that caused the overflow

## Additional Context
- Reported against `release/v0.6.9`
- Affects any page using `SearchAndFilterBar` with children (Metrics, Behaviors, Projects)
- No behaviour changes — only layout restructuring

## Testing
Navigate to `/metrics` — filter chips and action buttons should render cleanly on separate rows with no overlap at any viewport width.

### DIFF
--- a/apps/frontend/src/components/common/SearchAndFilterBar.tsx
+++ b/apps/frontend/src/components/common/SearchAndFilterBar.tsx
@@ -35,22 +35,24 @@ export default function SearchAndFilterBar({
 }: SearchAndFilterBarProps) {
   return (
     <Box
+      data-testid="search-action-row"
       sx={{
         display: 'flex',
-        flexDirection: 'column',
-        gap: 1.5,
+        flexDirection: { xs: 'column', md: 'row' },
+        gap: 2,
+        alignItems: { xs: 'stretch', md: 'center' },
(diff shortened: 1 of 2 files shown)


nicolai-rhesis:
@harry-rhesis not sure why you introduced the change (on Mar 10) that lead to this bug, but if it was intentional, please review carefully.

PR #1493, opened 2026-03-12
Fix tests grid random reordering with stable secondary sort

nicolai-rhesis:
## Purpose
Tests in the grid reorder randomly on each page load because \`created_at DESC\` is the only sort key. When multiple tests share the same timestamp (e.g. bulk-generated tests), the database returns them in an arbitrary order.

## What Changed
- Added a \`content\` hybrid property to the \`Test\` model that resolves to \`test_configuration['goal']\` for multi-turn tests and \`prompt.content\` for single-turn tests — matching exactly what the Content column in the grid displays. Both Python and SQL paths treat NULL and empty-string goal the same (fall back to \`prompt.content\`).
- Extended \`QueryBuilder.with_sorting\` to accept optional \`secondary_sort_by\` / \`secondary_sort_order\` parameters (documented, default \`asc\`), making multi-level sorting reusable across entities
- Extended \`get_items_detail\` to pass those parameters through
- Updated \`get_tests\` to pass \`secondary_sort_by="content"\`
- Always appends \`id ASC\` as a final unique tiebreaker so results are strictly deterministic even when all other sort keys tie

Final sort order for tests: \`created_at DESC, content ASC, id ASC\`

## Scope note
This PR only modifies 4 backend files (models/test.py, utils/query_utils.py, utils/crud_utils.py, crud.py). The \`GET /test_runs/{test_run_id}/metrics\` endpoint visible in the diff against \`main\` was already present in \`release/v0.6.9\` and is **not** part of this PR.

## Additional Context
- No API contract changes — purely a backend query ordering fix
- The \`content\` hybrid property is reusable for future sorting/filtering needs

## Testing
1. Deploy the backend
2. Open /tests in the app
3. Reload the page multiple times — test order should remain consistent

### DIFF
--- a/apps/backend/src/rhesis/backend/app/utils/query_utils.py
+++ b/apps/backend/src/rhesis/backend/app/utils/query_utils.py
@@ -239,6 +239,10 @@ class QueryBuilder:
                 self.query = self.query.order_by(desc(secondary_column))
             else:
                 self.query = self.query.order_by(secondary_column)
+        # Always append id ASC as a final unique tiebreaker so results are
+        # strictly deterministic even when all other sort keys are equal.
+        if self._sort_by and hasattr(self.model, "id"):
+            self.query = self.query.order_by(self.model.id)
 
     def _apply_pagination(self):
         """Apply pagination if configured"""


harry-rhesis:
Thanks for the PR! I like the idea of `content` hybrid property 💡, should be similar for has_attachments, or has_files on the test level. Instead of having "secondary_sort_by", I would make it an array, something like:

sort_by = [
  { "field": "created_at", "direction": "desc" },
  { "field": "name", "direction": "asc" }
]

Then it would be very nice for other entities as well.

PR #1605, opened 2026-04-04
feat(telemetry): add retry with exponential backoff to exporter

AlexMaggioni:
## Summary

The `RhesisOTLPExporter` overrides `export()` from OpenTelemetry's `OTLPSpanExporter` to support the custom JSON/Pydantic wire format (the backend accepts `OTELTraceBatch` JSON, not standard OTLP protobuf). By overriding `export()`, we lost the built-in retry loop that the base exporter provides (up to 6 retries with exponential backoff on 5xx/408/ConnectionError).

This PR re-introduces retry logic using `tenacity`.

## What Changed

- Added `tenacity` as a dependency to `packages/rhesis`
- Extracted the HTTP POST into a `_post()` method, called via a `Retrying` instance with:
  - Exponential backoff with jitter (1s initial, 2x base, 2s jitter, 10s max)
  - Up to 3 attempts (configurable via `max_retries`)
  - Retries only on truly transient errors: `ConnectionError`, `Timeout`, HTTP 429 (Too Many Requests), HTTP 503 (Service Unavailable)
  - Does **not** retry client errors (401, 403, 422) — those are permanent failures

## Retry vs consecutive failure tracking

The exporter now has two layers of failure handling — this is open for discussion:

- **Tenacity retry** (new): retries a single `export()` call up to 3 times. Absorbs short-lived transient errors (a 503 that lasts 1-2 seconds) so the batch is not lost.
- **Consecutive failure tracking** (existing): counts how many `export()` calls fail *after* retries are exhausted. Alerts the user when the backend is persistently down (5+ consecutive failures). This counter now only increments when all retry attempts have been exhausted, making it harder to trigger — which arguably better reflects a real infrastructure issue rather than a momentary blip.

Whether both layers are necessary or if the alerting thresholds need adjusting is worth discussing.

## Dependencies

This PR must be merged **after** #1603 (`feat: add lightweight rhesis-telemetry package`), which creates the `packages/rhesis/src/rhesis/telemetry/` module that this PR modifies. Will open the PR in `Ready for review` once #1603 has been merged or closed.

## Testing

- Ruff lint and format pass

---

> [!NOTE]
> **Medium Risk**
> Changes telemetry export behavior by introducing automatic retries and backoff, which can affect delivery timing and failure/alert semantics under load or partial outages. Scope is limited to client-side telemetry exporting and dependencies, with no auth/data model changes.
> 
> **Overview**
> Adds `tenacity` as a dependency for `packages/rhesis` (and updates lockfiles) to reintroduce **transient-failure retrying** for telemetry trace exports.
> 
> `RhesisOTLPExporter` now wraps its HTTP POST in a configurable retry loop (`max_retries`, exponential backoff with jitter) that retries only on connection/timeouts and HTTP `429`/`503`, logging each retry and extracting the POST into `_post()`.
> 
> Test coverage is expanded to validate retry/no-retry behavior across transient vs. permanent HTTP errors and to confirm retry exhaustion behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1425c3099555dcd270969d3bb46630ebdd994541. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


harry-rhesis:
@AlexMaggioni: thanks for taking up this! The explanation on why the scope has grown was very helpful, thank you! 🙌

~~Just a short question: have the issues pointed out by the AI reviewers been addressed?~~ YES, I just checked they've all been addressed. Thank you for the excellent work! 💪

PR #1607, opened 2026-04-07
Fix: External Dns Controller

asadaaron:
This PR introduces changes from the `fix/external-dns-controller` branch.

## 📁 Files Changed (       9 files)

```
kubernetes/base/external-dns/external-dns-application.yaml
kubernetes/base/internal-dns/internal-dns-application.yaml
kubernetes/clusters/dev/argocd/argocd-ingress.yaml
kubernetes/clusters/dev/external-dns/kustomization.yaml
kubernetes/clusters/prd/argocd/argocd-ingress.yaml
kubernetes/clusters/prd/external-dns/kustomization.yaml
kubernetes/clusters/stg/argocd/argocd-ingress.yaml
kubernetes/clusters/stg/external-dns/kustomization.yaml
terraform/infrastructure/scripts/verify-dev-cluster.sh
```

## 📋 Commit Details

```
40cd91a70 - fix: - - The external-dns controller is configured to domain-filter=internal-example.com, which should be rhesis.ai instead - The argocd ingress is configured to argocd.dev.rhesis.ai, which is not a valid domain for the internal dns controller (domain-filter=rhesis.internal) - Both dns controllers have policy=upsert-only. That is indeed the default, but that means they don’t clean up after themselves. Both have owner-ids set, so it should be safe to set the policy to „sync“ - The argocd ingress is configured to use TLS. That should be disabled and the cluster-issuer annotation removed (Md Asaduzzaman Miah, 2026-04-02 17:11)
```

## ✅ Checklist

- [ ] All tests pass locally

### DIFF
--- a/kubernetes/base/external-dns/external-dns-application.yaml
+++ b/kubernetes/base/external-dns/external-dns-application.yaml
@@ -22,8 +22,8 @@ spec:
                 name: cloudflare-api-token
                 key: cf-api-token
         domainFilters:
-          - example.com # need to change to the actual domain
-        policy: upsert-only
+          - rhesis.ai
+        policy: sync # sync or upsert
         txtOwnerId: PLACEHOLDER
         sources:
           - service
--- a/kubernetes/base/internal-dns/internal-dns-application.yaml
+++ b/kubernetes/base/internal-dns/internal-dns-application.yaml
@@ -34,7 +34,7 @@ spec:
                 key: tsig-secret
         domainFilters:
           - PLACEHOLDER
-        policy: upsert-only
+        policy: sync
         txtOwnerId: PLACEHOLDER
         sources:
           - service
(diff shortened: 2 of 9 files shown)


asadaaron:
> is it intentional that **all** envs use `domain-filter=rhesis.ai` (vs e.g. `dev.rhesis.ai`/`stg.rhesis.ai`)? If so, worth double-checking the dev/stg Cloudflare API tokens are scoped so a mis-annotated Service/Ingress can’t mutate unrelated prod records.

Yes, it’s intentional. We’re using a single public zone rhesis.ai, so we set domain-filter=rhesis.ai in every environment (one filter for the whole zone). That keeps config simple and matches hostnames like dev-app.rhesis.ai / stg-app.rhesis.ai, which are still under rhesis.ai.

harry-rhesis:
Thanks @asadaaron for addressing the concerns by Pit!

PR #1628, opened 2026-04-09
Fix: log batch item failures in polyphemus service

alexeyhramov123:
## Purpose
Batch item failures in the polyphemus service were returned as `{"error": ...}` to the caller but never appearing in logs, making partial batch failures invisible server-side.


asadaaron (approved):

harry-rhesis:
@alexeyhramov123 thanks for making these changes!

PR #1638, opened 2026-04-14
Add batch chunking to OTEL exporter

harry-rhesis:
## Purpose

Customers with large multi-turn conversation traces hit Cloud Run's 32MB HTTP request body limit when exporting spans. This PR adds transparent batch chunking to `RhesisOTLPExporter` so large batches are split into multiple smaller HTTP requests automatically.

## What Changed

- Added `max_chunk_size` parameter to `RhesisOTLPExporter.__init__` (default 100)
- Added `_chunk_converted_spans()` method that partitions spans into chunks with root spans always in the first chunk (preserves backend `inject_pending_output` ordering)
- Modified `export()` to iterate over chunks with per-chunk deadline and retry budget
- Added 7 new tests in `TestExporterBatchChunking` covering chunking, root-span ordering, partial failure, per-chunk deadline, and configuration

## Additional Context

- No backend, frontend, or schema changes required
- Single-batch exports (the common case) have zero behavioral change
- Follows up on the retry/backoff work in #1605

## Testing

```bash
cd sdk
uv run pytest ../tests/sdk/telemetry/test_exporter.py -v
```

All 41 tests pass (34 existing + 7 new).

### DIFF
--- a/apps/backend/src/rhesis/backend/app/services/telemetry/enrichment/processor.py
+++ b/apps/backend/src/rhesis/backend/app/services/telemetry/enrichment/processor.py
@@ -64,11 +64,28 @@ class TraceEnricher:
             logger.warning(f"No spans found for trace {trace_id}")
             return None
 
-        # Check if already enriched (cache hit)
-        if spans[0].enriched_data:
-            logger.debug(f"Using cached enrichment for trace {trace_id}")
-            # Convert cached dict back to Pydantic model for type safety
-            return EnrichedTraceData(**spans[0].enriched_data)
+        # Smart cache: return cached enrichment only when no new spans have
+        # arrived since the last enrichment pass.  This allows progressive
+        # re-enrichment as child spans (LLM calls, costs, etc.) arrive after
+        # the root span, while avoiding redundant work once everything is
+        # processed.
+        #
+        # `processed_at` is set on every span row by mark_trace_processed.
+        # Newly stored spans default to processed_at=None, so any span with
+        # processed_at=None means new data has arrived that warrants a re-run.
+        root_span = spans[0]  # ordered by start_time asc, so this is the root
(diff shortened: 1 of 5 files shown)


harry-rhesis:
Addressed peqy's review comments in bb93432:

**Comment 1 (validate `max_chunk_size`):** Added `ValueError` guard in `__init__` if `max_chunk_size < 1`, plus `test_max_chunk_size_validation` test. Good catch.

**Comment 2 (partial-send duplication):** Valid observation but not a practical concern -- `BatchSpanProcessor` does not retry on `FAILURE`, so successfully-sent chunks are never re-sent. Added a docstring note to `export()` documenting the partial-send semantics. An idempotency layer / upsert on `(trace_id, span_id)` would be over-engineering for this PR.

harry-rhesis:
**Re: comment 3 (roots exceeding `max_chunk_size`):**

Valid observation -- if `BatchSpanProcessor` flushes 512 independent single-turn traces, most spans are roots and chunk 1 could be as large as the original batch.

However, this edge case doesn't conflict with the problem this PR solves. The 32MB payload issue comes from multi-turn conversations where one trace accumulates hundreds of child spans with large attributes (RAG context, conversation history). In that scenario, root count is low (one per turn) and children dominate the batch.

The suggested fix (splitting roots across multiple chunks) would conflict with the backend's `inject_pending_output` ordering: it pops pending output from the cache on first sight of a `trace_id`, then injects only into root spans in that batch. If roots for the same `trace_id` land in different chunks, only the first chunk gets the output -- the rest is lost.

For the 512-independent-traces scenario, each trace is small (single span), so the serialized payload stays well under 32MB even without chunking.

Documenting this as a known limitation rather than adding complexity that risks breaking output injection.

PR #1705, opened 2026-04-30
Fix run_sync to always use background thread to prevent event loop errors

harry-rhesis:
## Purpose

Fire-and-forget cleanup tasks spawned by LiteLLM's `AsyncHTTPHandler` were causing `RuntimeError: Event loop is closed` because `run_sync` was calling `asyncio.run()` directly. When `asyncio.run()` exits it closes the event loop immediately, killing any background coroutines that were still running.

## What Changed

- Removed the `asyncio.run()` fast path from `run_sync` in `sdk/src/rhesis/sdk/async_utils.py`
- Always uses a dedicated background thread with its own event loop, which keeps running until all tasks complete naturally

## Additional Context

- Root cause surfaced through LiteLLM usage but applies to any async caller that leaves background tasks running after the coroutine returns
- No API changes — `run_sync` signature and behaviour are unchanged for callers

## Testing

```bash
cd sdk
make test
```

### DIFF
--- a/sdk/src/rhesis/sdk/async_utils.py
+++ b/sdk/src/rhesis/sdk/async_utils.py
@@ -1,7 +1,7 @@
 """Async utilities for bridging sync and async code.
 
 Provides run_sync() for executing async coroutines from synchronous contexts,
-with automatic detection of running event loops (Jupyter, FastAPI, etc.).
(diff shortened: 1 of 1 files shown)


harry-rhesis:
@akwasigroch, what do you think, can we merge this?

PR #1714, opened 2026-05-04
feat(frontend): explorer copy and sidebar placement

akwasigroch:
## Purpose
Align the adaptive-testing experience with the Explorer product naming in the UI only (no backend or API path changes), and make the sidebar entry consistent with Architect (beta placement and order).

## What Changed
- Renamed user-facing strings on the test explorer list and detail flows (titles, breadcrumbs, metadata, grid actions, empty state, delete copy, settings/evaluate dialogs).
- Added an Explorer section label above the tests toolbar in tree and list views.
- Moved the Explorer nav item directly under Architect; use `action: <BetaBadge />` like Architect instead of inline beta in the title.

## Additional Context
- Routes remain `/adaptive-testing`; backend behavior string checks are unchanged where required for API compatibility.

## Testing
- Smoke: open Explorer from the sidebar, list page shows "Test explorer" and "Explorer sessions"; create flow shows "New session"; detail page breadcrumbs and explorer panel label render as expected.


harry-rhesis:
Reviewing this as a staff engineer — flagging the scope first, then specific items inline.

## Scope

The PR title/description say UI copy + sidebar placement only ("no backend or API path changes"), but commits 3–4 (`a19d83ff4`, `cf3040ca7`) quietly turn this into a breaking REST rename (`/adaptive_testing` → `/explorer`), a backend module/schema rename, and a docs deletion. Worth either a) splitting the rename into its own PR, or b) updating the PR description and commit messages so reviewers and the changelog accurately reflect what's shipping.

Since the functionality isn't widely used yet, I'm fine treating the API rename as a clean break — no need for an `/adaptive_testing` alias or deprecation cycle.

## Half-rename

End state today: `clientFactory.getExplorerClient().getAdaptiveTestSets()` and `class ExplorerClient { deleteAdaptiveTestSet(...); importAdaptiveTestSetFromSource(...); }`. Mixed naming is a worse spot than either extreme. With a clean break, the cheap path is to also rename schemas (`CreateAdaptiveTestBody`, `AdaptiveSettingsResponse`, `Import/ExportAdaptiveTestSetResponse`), service functions (`create_adaptive_test_set`, `evaluate_tests_for_adaptive_set`, …), TS types (`AdaptiveTestSet`, `AdaptiveSettings*`), client methods, and test class names in this PR. The one string that should stay is `ADAPTIVE_TESTING_BEHAVIOR = "Adaptive Testing"` — it's stored data and is correctly preserved already.

## Action items (details inline)

Merge-blockers:

1. Token-auth gap on `POST /explorer` — see `main.py` comment.
2. `embedding_config.source` change quietly invalidates dedup hashes — see `embeddings.py` comment.
3. Replace the deleted `docs/content/docs/test-sets/adaptive-testing.mdx` with an `/docs/test-sets/explorer` page — the new glossary entry already links to it; see `glossary-terms.jsonl` comment.

Cleanup while you're here:

4. Finish (or revert) the rename in identifiers — see `explorer-client.ts` comment on the class line.
5. Drop the dead `validateTree` / `getTreeStats` methods — backend has no matching routes; only referenced in unit tests.

Commit hygiene:

6. Squash/rewrite the `next` commit and the misleading `feat(explorer): add adaptive testing tree endpoints and schemas` commit (it doesn't add anything).
7. Mark the squashed commit `refactor!:` and add a one-line `CHANGELOG.md` entry.

What I like: the original UI copy reads better, sidebar placement matches the existing Architect/`BetaBadge` pattern, file renames preserve git history, tests were updated alongside, and the `"Adaptive Testing"` behavior string was correctly kept so existing data still resolves.

apps/backend/src/rhesis/backend/app/main.py
```diff
@@ -97,7 +97,7 @@
     "/sources/",
     "/models/",
     "/connector/",
-    "/adaptive_testing/",
+    "/explorer/",
```
harry-rhesis, line 100: **[Blocker] Token auth gap.** Use `"/explorer"` (no trailing slash) here.

`AuthenticatedAPIRoute.get_dependencies` does `any(self.path.startswith(route) for route in token_enabled_routes)`. The router is mounted with `prefix="/explorer"` and the create endpoint registers as `""`, so its `self.path` is the bare `/explorer`. `"/explorer".startswith("/explorer/")` is `False`, so `POST /explorer` falls through to session-only auth and breaks token-based clients trying to create explorer sets.

This bug exists today for `/adaptive_testing` too — the PR is the right place to fix it. Dropping the trailing slash covers both `/explorer` and `/explorer/...`.

apps/backend/src/rhesis/backend/app/services/explorer/embeddings.py
```diff
@@ -382,7 +382,7 @@ def create_test_embedding(
         "model_name": model_name,
         "dimension": dimension,
         "model_id": model_id,
-        "source": "adaptive_testing",
+        "source": "explorer",
```
harry-rhesis, line 386: **[Blocker] Make this an explicit choice, not a side-effect of the rename.**

This dict is fed into `_compute_hash(config)` two lines down, so flipping `"adaptive_testing"` → `"explorer"` produces a different `config_hash` for otherwise identical embeddings. New writes won't dedup against existing rows, and any future query/filter on `embedding_config.source` will see two values for the same logical thing.

Two acceptable paths:

1. Keep `"source": "adaptive_testing"` (it's a stable internal tag, decoupled from the route name) and add a one-line comment explaining why.
2. Change to `"explorer"` **and** ship a one-line Alembic data migration that updates existing rows so the hash space stays consistent.

Either is fine; silently diverging new vs. old writes is not.

docs/content/glossary/glossary-terms.jsonl
```diff
@@ -60,7 +60,7 @@
...
...supports generating stored outputs against your endpoint, then evaluating them with selected metrics.\n\n## Workflow\n\n1. **Define topics** that reflect behavioral areas\n2. **Add tests** within each topic—manually, via import, or suggestions\n3. **Generate outputs** in bulk against your endpoint\n4. **Evaluate** to find gaps and failures\n5. **Iterate** with suggestion flows before persisting\n\n## SDK usage\n\nThe SDK exposes tree APIs under `rhesis.sdk.adaptive_testing` (TestTree, schemas). These call the same backend `/explorer` routes as the UI.\n\n## Difference from static test sets\n\nRegular test sets are a fixed bundle for execution. Explorer sessions emphasize an evolving topic tree, curated outputs, and suggestion-driven growth.\n\n## Best practices\n\n- Name topics by intent or risk area, not tiny prompt variants\n- Re-run evaluation after material model or prompt changes\n- Use export when you need a frozen, standard test set for CI\n", "category": "Advanced Testing", "relatedTerms": ["test-set", "test-generation", "behavior", "topic", "sdk"], "docLinks": ["/docs/test-sets/explorer"], "aliases": ["Test explorer", "Adaptive testing (legacy name)", "adaptive test set"]}
```
harry-rhesis: **[Blocker] Broken link.** `docLinks: ["/docs/test-sets/explorer"]` points to a page that doesn't exist — the previous `docs/content/docs/test-sets/adaptive-testing.mdx` (224 lines, the primary API reference with curl examples) was deleted in this PR and not replaced.

Before merging, please add a replacement page at `docs/content/docs/test-sets/explorer.mdx` (a slimmer version is fine) and re-add the entry to `docs/content/docs/test-sets/_meta.tsx`. Otherwise this glossary entry, search-engine-indexed bookmarks, and any internal `/docs/test-sets/adaptive-testing` links all 404.

apps/frontend/src/utils/api-client/explorer-client.ts
```diff
@@ -31,17 +31,17 @@ import {
...
 
 /**
- * API client for adaptive testing CRUD operations.
+ * API client for Explorer (test tree) HTTP endpoints.
  *
  * Provides methods for managing test tree data within a TestSet,
  * including operations on topics and tests.
  */
-export class AdaptiveTestingClient extends BaseApiClient {
+export class ExplorerClient extends BaseApiClient {
```
harry-rhesis, line 40: **[Cleanup] Finish the rename, here and through the rest of the PR.**

Methods on this class are still `getAdaptiveTestSets`, `deleteAdaptiveTestSet`, `importAdaptiveTestSetFromSource`, `exportRegularTestSetFromAdaptive`, `getAdaptiveSettings`, `updateAdaptiveSettings`. Same pattern on the backend: `routers/explorer.py` exposing schemas like `CreateAdaptiveTestBody`, `AdaptiveSettingsResponse`, and service functions like `create_adaptive_test_set`, `evaluate_tests_for_adaptive_set`. Tests still have `class TestListAdaptiveTestSetsEndpoint` etc.

With a clean-break stance the cheap path is to rename the identifiers too (mostly mechanical since you're already touching every file). The one exception is `ADAPTIVE_TESTING_BEHAVIOR = "Adaptive Testing"` — that's stored data and should stay.

**[Cleanup] Also drop the dead `validateTree` / `getTreeStats` / path-based `getTopic` methods on this client.** Backend has no `/{id}/validate`, `/{id}/stats`, or `GET /{id}/topics/{topic_path}` route — they're only referenced by `__tests__/explorer-client.test.ts`. Better to remove now than carry them through the rename.

harry-rhesis (changes requested):
Inline action items — see top-level comment for the scope/rename overview.

apps/backend/src/rhesis/backend/app/services/explorer/embeddings.py
akwasigroch: Right now embedding service is not used

docs/content/glossary/glossary-terms.jsonl
akwasigroch: Fixed

harry-rhesis (approved):

PR #1717, opened 2026-05-04
Prevent submission of whitespace-only token names

thisisharsh7:
Prevents creation of tokens with names that contain only whitespace.

## Changes
- Trim token name before submitting
- Block submission if trimmed name is empty
- Disable "Create" button when input is only whitespace


harry-rhesis:
@thisisharsh7 thank you for submitting this PR, excellent catch! 🙌

PR #1754, opened 2026-05-11
Fix: prefer SQLALCHEMY_DATABASE_URL env var in alembic migrations

EmanueleDeRossi1:
Alembic was always constructing the database URL from individual  components, ignoring a full DATABASE_URL if set. This caused issues  in Docker and Cloud Run environments where a single connection string  is the standard. Now checks for SQLALCHEMY_DATABASE_URL first and  falls back to the existing behavior if not set.


asad-miah (approved):
looks good

PR #1771, opened 2026-05-15
Fix Redis backend stability: break autoscaler feedback loop on trace ingestion

harry-rhesis:
## Summary

- Cap Redis connection pools across all consumers with explicit `max_connections` limits to prevent unbounded connection growth under contention
- Apply fail-fast Celery broker overrides in the FastAPI web process (2s timeouts, zero retries) so a degraded Redis never blocks HTTP threads
- **Remove `_check_workers_available()` entirely** — it performed a 1s `inspect.ping()` Redis round-trip on every cache miss, triggering a feedback loop that caused Cloud Run to spin up 14+ instances at just 2-3 req/s
- Make telemetry trace ingestion **fire-and-forget**: store spans to DB, dispatch async, return success regardless of broker availability
- Release DB session before Celery `.delay()` to prevent slow broker dispatch from holding connections hostage

## Root cause

A feedback loop where slow Redis amplified request latency and caused autoscaler storms:

1. `_check_workers_available()` does `inspect.ping(timeout=1.0)` — a 1s synchronous Redis round-trip on every cache miss
2. When ping fails → sync fallback adds 500ms-2s of DB work in-request
3. Each request holds a concurrency slot 10-15x longer (130ms → 1.5s)
4. Cloud Run autoscaler overreacts: 0.4 concurrent slots needed at 130ms, 4.5 at 1.5s
5. New instances open more Redis connections, making Redis slower → repeat

One organization sending 192K trace requests/day (99.95% of traffic) made this visible.

## Changes

| File | Change |
|------|--------|
| `app/services/async_service.py` | Remove `_check_workers_available`, `_worker_cache`, `workers_available` param; always try async first, fall back to sync on broker errors |
| `app/routers/telemetry.py` | Fire-and-forget dispatch: remove sync fallback, log warning on broker failure, return success |
| `app/services/telemetry/enrichment/service.py` | Remove `workers_available` parameter from all methods |
| `celery/config.py` | Add `WEB_CELERY_OVERRIDES` (2s timeouts, 0 retries, pool limit 5) |
| `celery/core.py` | Add `apply_web_context_overrides()` for web process |
| `app/main.py` | Call overrides during FastAPI lifespan startup |
| `app/services/cache.py` | `max_connections=3` on sync Redis cache |
| `app/services/garak/cache.py` | `max_connections=3` on async Garak cache |
| `app/services/connector/redis_client.py` | `max_connections=3` on SDK RPC Redis |
| 6 test files | Update tests for simplified async-first interface (-576 lines net) |

## Test plan

- [ ] Verify backend tests pass
- [ ] Deploy and confirm trace ingestion works with workers available
- [ ] Confirm fire-and-forget behavior: `.delay()` fails fast with warning, response still succeeds
- [ ] Monitor Cloud Run instance count: should stay at 1-3 under 2-3 req/s (was spiking to 14+)
- [ ] Monitor P99 latency on POST /telemetry/traces: should stay under 200ms

### DIFF
--- a/apps/backend/src/rhesis/backend/app/main.py
+++ b/apps/backend/src/rhesis/backend/app/main.py
@@ -135,6 +135,12 @@ async def lifespan(app: FastAPI):
     """
     set_logger()
 
+    # Apply fail-fast Celery broker settings for the web process.
+    # Must happen before any Celery task is published from an HTTP handler.
+    from rhesis.backend.celery.core import apply_web_context_overrides
+
+    apply_web_context_overrides()
+
     # Set anyio threadpool size for async-to-thread offloading.
     # Default is 40; 100 is a reasonable production value for 2 vCPU + concurrency 80.
     try:
--- a/apps/backend/src/rhesis/backend/app/routers/telemetry.py
+++ b/apps/backend/src/rhesis/backend/app/routers/telemetry.py
@@ -26,7 +26,6 @@ from rhesis.backend.app.schemas.telemetry import (
     TraceType,
 )
 from rhesis.backend.app.services.async_service import BROKER_ERRORS
-from rhesis.backend.app.services.telemetry.enrichment import EnrichmentService
 from rhesis.backend.app.services.trace_review_override import (
     apply_review_override as trace_apply_review_override,
 )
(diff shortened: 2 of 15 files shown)


harry-rhesis:
## Response to peqy reviews

### Review 1 (on `46947b5`)

1. **`_worker_cache.get("available")` bypasses TTL** — Fixed in `dd29c94`. We removed `_worker_cache` and `_check_workers_available()` entirely.

2. **Guard sync fallback if `fallback_spans` is empty** — Fixed in `dd29c94`. The entire sync fallback path was removed from telemetry ingest; it's fire-and-forget now.

3. **`BROKER_ERRORS` includes broad builtins** — Acceptable. `_enqueue_async` only calls Celery's `.apply_async()`, so `TimeoutError`/`ConnectionError` here are genuinely broker-related. The generic `except Exception` below still catches anything else and falls back to sync, so nothing is silently lost.

4. **Explicitly disable `broker_connection_retry` in web overrides** — Fair point. In Celery 5.x `max_retries=0` already means no retries, but being explicit wouldn't hurt. Low priority.

5. **Infra workflow vars** — Unrelated to this PR (from a different commit already on the branch).

### Review 2 (on `dd29c94`)

1. **Extra `db.commit()` + `db.close()` redundant/interferes with session lifecycle** — Intentional. We commit to guarantee spans are persisted before fire-and-forget dispatch, and close to release the DB connection *before* `.delay()` so a slow broker can't hold it hostage. FastAPI's dependency `finally` block calling close again is a safe no-op.

2. **`BROKER_ERRORS` broad builtins** — Same as review 1, point 3 above.

3. **`apply_web_context_overrides()` could run in worker process** — Not a risk. It's only called from `app/main.py` inside the FastAPI lifespan handler. Workers use `worker.py` which imports the Celery app directly without calling this function.

PR #1773, opened 2026-05-18
1727 differing results using polyphemus between sdk and platform

alexeyhramov123:
Add harmful mode to synthesizers

Introduces a harmful flag across all synthesizers that switches the generation template to produce adversarial attack prompts instead of standard test cases. Refines the adversarial template to generate first-person attack messages rather than meta-descriptions, tuned for Polyphemus.


harry-rhesis:
Thanks @alexeyhramov123 — this is a solid fix for the SDK vs platform mismatch, and the template refinement (first-person attack messages instead of meta-descriptions) is especially important for Polyphemus.

**Follow-ups we discussed for a later change (for visibility, not blocking this PR):**

- **Naming / API:** Longer term we may replace the boolean `harmful` with an explicit **`generation_mode`** (or similar) with two values only: **`Adversarial`** and **`Balanced`**, so the intent is obvious in APIs, docs, and compliance reviews. Mapping would be straightforward: adversarial ≈ `harmful=True`, balanced ≈ `harmful=False`.

- **Parity across synthesizers:** The same mode should apply to **`ConfigSynthesizer` and `PromptSynthesizer`** (and ideally live on `TestSetSynthesizer` / `GenerationConfig` so the backend can pass it through Celery and `/generate/tests` the same way the SDK does).

- **Polyphemus default:** When the resolved model is Polyphemus and the caller **does not** set a mode explicitly, we could default to **Adversarial**; any explicit `Balanced` (or `harmful=False`) should always win so notebooks and integrations stay predictable.

- **Templates:** Still likely **one `base.jinja`** with two branches (or shared macros + mode-specific includes), not three separate top-level templates, unless one mode grows enough to warrant a split.

Happy to help wire any of the above in a follow-up PR if useful.

harry-rhesis:
**Shared “contract” between Balanced and Adversarial**

For a follow-up refactor, we can structure `base.jinja` (and/or small includes) so **both modes share one block of invariant instructions** and only swap the **persona + how to write `prompt`** (neutral QA vs first-person attacker).

**What should stay identical across modes**

- **Task framing:** “Generate EXACTLY `{{ num_tests }}` test cases for …” plus `{{ generation_prompt }}` (and config-driven fields from `ConfigSynthesizer` the same way as today).
- **Output shape:** each row still has `prompt`, `behavior`, `category`, `topic`.
- **Constraint rule:** when `behaviors` / `categories` / `topics` are **explicitly provided** (e.g. from `GenerationConfig` / UI chips), the model must use **only** those values—no extra labels. That logic should live in **one** place, not duplicated per mode.
- **Defaults when lists are omitted:** define defaults **once** (ideally in Python as structured context passed into Jinja, or in a single shared macro) so Balanced and Adversarial never drift. For example, if we standardize on a **single default behavior** unless overridden—**Robustness**—both modes read the same default list; `ConfigSynthesizer` continues to override by passing explicit `behaviors` from config.
- **Categories/topics:** same idea—shared default block or context keys like `default_behaviors`, `default_categories`, rendered identically; mode only changes *tone* and *prompt style*, not whether we respect the config.

**What should differ by mode**

- Opening role (QA agent vs red-team generator).
- Instructions for the **content** of `prompt` (realistic user message vs raw attack line).
- Coverage/diversity rules that are **mode-specific** (e.g. balanced mix vs adversarial-only / attack directions)—kept in small mode-specific sections **after** the shared contract.

That way “Balanced” vs “Adversarial” is mostly a **skin** over the same behavioral and config semantics, with Polyphemus defaulting to Adversarial only affecting which skin is chosen when the user does not pass an explicit mode.

PR #1780, opened 2026-05-19
feat(frontend): UI revamp — Figma-aligned design system, layout and pages

nicolai-rhesis:
## Purpose

Complete frontend UI overhaul to align with the Figma design system. Replaces the Toolpad-based layout with a custom shell, introduces unified design tokens and shared list/detail patterns, and rebuilds list and detail pages across the product to match Figma specifications.

### Design system & layout
- **Theme** (`theme.ts`): Figma tokens (`GREYSCALE`, `BORDER_RADIUS`, `BACKDROP_COLORS`, `ELEVATION`), primary blue `#0080AF`, typography variants, and component overrides for Button, Chip, Drawer, Table, TextField, Autocomplete (56px height alignment).
- **Layout shell**: custom `AppShell` (CSS grid, collapsible sidebar) replaces `@toolpad/core/DashboardLayout`. New `Sidebar` with Define / Generate / Improve / CONNECT sections, Figma SVG nav icons (24×24 viewport), org menu popover, and polished collapsed state.
- **Shared components**: `PageLayout`, `BaseDrawer`, `EntityCard`, `Fab` / `FabGroup`, `FilterButton`, `SearchPill`, `StatusChip`, `BadgeChip`, `GridBadge`, `Tag`, `EntityEmptyState`, `FilterDrawerShell` / `FilterSection`, `EditableSection`, `ViewField`, `DetailTabNav`, `SectionCard`, `FigmaPaginationFooter`, `SortOnlyColumnMenu`.
- **Tags vs badges**: split `Tag` (editable rectangular tags) and `GridBadge` (pill metadata in grids) per Figma; 12px in grids, 14px on detail pages.
- **CssBaseline** mounted to remove default body margin frame.

### Navigation & routing
- Sidebar reorganized: Insights under Improve, Test Sets pluralized, API Tokens → API, Develop → CONNECT, official MCP icon, Models under CONNECT.
- **Insights**: dashboard KPIs, recent runs, and activity merged into `/insights` Overview tab; `/dashboard` and `/test-results` redirect; post-login landing stays on `/architect`.

### List pages (unified toolbar: search pill, filter drawer, FAB, `EntityEmptyState`, themed grid card)
- Tests, Test Sets, Test Runs, Behaviors, Metrics, Models, MCP, Knowledge, Tasks, Endpoints, Explorer, API Tokens.

### Detail pages (tabbed layout, per-card edit, header FABs, metadata strip)
- Test detail (Basic Information, Linked Entities, Tasks) + `GET /tests/{test_id}/test_sets` backend endpoint.
- Test set detail (Basic Information, Linked Tests, Tasks).
- Endpoint detail (Overview, Connection, Mappings, Test) — split from monolithic `EndpointDetail`.
- Org settings (tabbed) and team (filter drawer, invite FAB drawer, OData filters).

### Page-specific highlights
- **Tests grid**: Figma-aligned DataGrid, `TestFilterDrawer`, empty state CTA, badge chips via `BadgeChip`.
- **Behaviors / Metrics / Models / MCP**: `EntityCard`-based grids, delete-only cards, filter drawers; model cards open connection dialog on click (no edit icon).
- **Knowledge**: upload and MCP import moved to drawers; layout aligned with tests.
- **Traces**: filter drawer + unified toolbar; DataGrid column resize enabled by default with fixed-width columns on key grids.
- **Multi-turn test config**: integrated with parent `EditableSection` single save flow.
- **Dark mode**: improved contrast for detail UI, editable sections, and pagination.

### Backend (small)
- `GET /tests/{test_id}/test_sets` with pagination for linked test sets on test detail.
- Auth redirect URLs updated for `/insights` (replaces `/dashboard`).

### Tooling & tests
- `check-hardcoded-styles.js` updated for new Figma palette and token allowances.
- Removed `@toolpad/core` dependency; added layout/component unit tests and updated E2E specs (navigation, insights, org settings, tasks, team).

## Additional Context

- Branch: `feat/frontend-ui-revamp`
- ~304 files changed (+23k / −16k lines); MUI v7 + Emotion stack preserved, no new UI library dependencies.
- Internal Figma audit doc removed from repo (public repo).

## Testing

1. `cd apps/frontend && npm run type-check` — should exit 0
2. `npm run lint` — 0 errors
3. `node ../../scripts/check-hardcoded-styles.js` — 0 violations
4. `npm run test` — unit tests for AppShell, PageLayout, Sidebar
5. E2E (with dev stack): `npm run test:e2e` — navigation, insights, org settings, tasks, team flows updated
6. Manual smoke: sidebar collapse/expand, list pages (toolbar, filter drawer, FAB), test/test-set/endpoint detail tabs, `/insights` overview, org settings and team pages

## CI note
- **Lint** (required): depends on `npm ci`; fixed by syncing `package-lock.json` with Node 24.
- **E2E** (3 shards + merge-reports): expected to fail on this revamp until a follow-up PR updates Playwright specs/selectors for the new shell and routes.


harry-rhesis:
## Backend Security and Feature Enhancement

While reviewing the backend changes in this PR, I identified and fixed a security vulnerability along with adding a missing feature:

### 🔐 Security Fix
- **Fixed RLS vulnerability** in `get_test_sets_for_test()` function
- The association table join was missing `organization_id` filtering, potentially allowing cross-tenant data access
- Now properly filters by both `test_id` AND `organization_id` in the association table

### ✨ Feature Addition  
- **Added OData filtering support** to the new endpoint `GET /tests/{test_id}/test_sets`
- Users can now filter test sets containing a specific test (e.g., `$filter=status eq 'active'`)
- Brings feature parity with the existing `GET /test_sets/{test_set_id}/tests` endpoint

### Changes Made
- **File**: `apps/backend/src/rhesis/backend/app/crud.py` - Updated `get_test_sets_for_test()` 
- **File**: `apps/backend/src/rhesis/backend/app/routers/test.py` - Added `filter` parameter to endpoint

The changes follow established backend patterns and maintain backward compatibility.

**Commit**: 82bf9c3d6

harry-rhesis (commented):
First off — thank you @nicolai-rhesis for putting this together. This is a huge, careful migration and the new design system is going to be a fantastic upgrade for our users. The decomposition of the old `EndpointDetail` god-file, the introduction of a shared `FilterDrawer` / `GridToolbar` / `SectionCard` vocabulary, the new `PageLayout` + `AppShell` primitives, and the consistent Figma alignment across pages are all big wins. The amount of code deleted (-20k lines) speaks for itself.

Reviewing as a staff frontend eng. Below are findings ordered by how blocking I think they are. **Nothing here should keep this from eventually merging** — but the "must-fix" bucket I'd want resolved before we flip this on for real users, because they affect correctness (dark mode, a11y, hydration) rather than aesthetics.

---

## 🔴 Must-fix before turning this on for users

### 1. Dark mode is broken across most new components

The theme defines `palette.greyscale` as a mode-aware token, but ~40 new/modified files import `GREYSCALE` directly and read the **light** ramp unconditionally (`GREYSCALE.light.body`, `GREYSCALE.light.title`, etc.). Examples:

- `components/common/FilterDrawer.tsx` (8 hits, including `filterChipSx`)
- `components/common/DetailTabNav.tsx` (`GREYSCALE.light.title` on lines 83 and 95, plus a hard-coded `#b6bdc9`)
- `app/(protected)/projects/components/ProjectFilterDrawer.tsx`
- `app/(protected)/endpoints/[identifier]/page.tsx` (metadata strip)

The right pattern is already used in `ViewField.tsx`:

```ts
color: theme => theme.palette.greyscale.subtitle
```

The pattern `theme.palette.mode === 'light' ? GREYSCALE.light.body : GREYSCALE.dark.body` (used 80+ times in `Sidebar.tsx`) is **also** the wrong solution — it ignores the theme-aware token already declared in `theme.ts`.

**Recommendation:** sweep `GREYSCALE.light.*` → `theme.palette.greyscale.*`, then add an ESLint `no-restricted-syntax` rule banning `GREYSCALE.light.` and `GREYSCALE.dark.` in component code to prevent regressions.

### 2. Custom tab nav has no keyboard support

`DetailTabNav.tsx` rolls its own ARIA tablist with `role=\"tab\"` + roving `tabIndex` but no `onKeyDown` handler. ARIA Authoring Practices require `ArrowLeft`/`ArrowRight`/`Home`/`End` to move focus between tabs. Today keyboard users can land on the active tab but can't move to its siblings. This affects **every** detail page (endpoints, tests, test-sets, …) so the regression is wide.

**Recommendation:** either lean on MUI's `<Tabs>` with the underline styled to match Figma, or add explicit keyboard handling. The hand-rolled ARIA setup is one of those things that almost never gets done right outside of axe-core's known patterns.

### 3. Sidebar popovers / clickable rows are not keyboard reachable

`Sidebar.tsx` uses plain `<Box onClick={…}>` for the org-menu trigger, brand row, user-avatar row, and every popover menu item — no `tabIndex`, no `role`, no Enter/Space handling. The pop-overs aren't `role=\"menu\"` with `MenuItem`s either.

**Recommendation:** swap the inner rows for `<MenuItem>` (and wrap the trigger in `<ButtonBase>`/`<IconButton>`). You get focus management, Enter/Space, and Escape-to-close for free.

### 4. `params` is unwrapped via `useEffect + setState` in 13 page files

```ts
const [identifier, setIdentifier] = useState<string>('');
useEffect(() => { params.then(p => setIdentifier(p.identifier)); }, [params]);
```

In React 19 / Next 15 the idiomatic call is `const { identifier } = React.use(params);`. The current shape forces an extra render where `identifier === ''`, has no cleanup, and can `setState` on an unmounted component on rapid navigation. One of the files (`tasks/[identifier]/page.tsx`) already does this correctly — let's make it the rule.

### 5. `QueryClient` is constructed at module scope

```ts
// apps/frontend/src/app/(protected)/layout.tsx
const queryClient = new QueryClient({ … });
```

TanStack's docs explicitly recommend constructing the client *inside* the component (`useState(() => new QueryClient())`). It's a client component today so the practical impact is small, but if this is ever rendered server-side (streaming, RSC interleave) one client is shared across requests — that's a real cross-user cache leak. Cheap fix, no reason not to do it now.

### 6. Silent `catch {}` blocks in destructive paths

`useEndpointDetail.ts` (`fetchProjects`), `endpoints/[identifier]/page.tsx` (project lookup), `EndpointHeaderActions.tsx` (delete handler) all swallow errors with empty `catch {}` or `catch { /* silent */ }`. After a refactor of this scale we will be flying blind in prod when something regresses. At minimum `console.error` + push to the error tracker.

---

## 🟡 Should-fix soon (before too many follow-ups land on top)

### 7. The filter-drawer pattern is duplicated 13 times

Every filter drawer (`Project`, `Endpoint`, `Test`, `TestSet`, `Behavior`, `Metric`, `MCP`, `Task`, `TestRun`, `Trace`, `Token`, `Source`, `Team`) repeats this exact shape:

```ts
const [draft, setDraft] = React.useState<F>(filters);
React.useEffect(() => { if (open) setDraft(filters); }, [open, filters]);
const handleReset = () => setDraft(EMPTY_FILTERS);
const handleApply = () => { onApply(draft); onClose(); };
```

…with drifting names (`EMPTY_FILTERS` vs `EMPTY_TEST_FILTERS` vs `EMPTY_TRACE_FILTERS`, `hasActiveTestFilters` vs `hasActiveProjectFilters`). Extracting a `useFilterDrawerDraft<F>(filters, open)` hook + a thin shell would delete a few hundred lines and stop the naming from drifting.

### 8. `Box component=\"button\"` is missing `type=\"button\"` in 9 filter drawers

```tsx
<Box
  component=\"button\"
  onClick={() => toggleEnvironment(env)}
  sx={filterChipSx(…)}
>
```

These render as `<button>` defaulting to `type=\"submit\"`. None of the drawers are inside a `<form>` today, so it's latent — but `filterChipSx` is reusable, and the moment one ends up in a form the page submits. `GridToolbar.PrimarySegmentedPills` and `DetailTabNav` already get this right; just copy the pattern.

### 9. `Sidebar.tsx` is 1037 lines and conflates layout / nav / branding / two pop-overs

The Settings/Team popover (Lines 685–791) and the Dark Mode/Sign Out popover (Lines 926–1032) are 80-line copies of each other. Extract `<SidebarMenuItem icon label onClick />` and split `SidebarBrand`, `SidebarOrgMenu`, `SidebarUserMenu` into their own files. Also: `LEFT_PANEL_PATH` is a hand-inlined SVG — `LeftPanelCloseIcon` / `LeftPanelOpenIcon` already exist in `@mui/icons-material`.

### 10. `NavigationProvider` papers over a hydration mismatch

Returning `{ navigation: [] }` until `mounted` causes an empty sidebar on first paint, then a snap to the real nav on hydration. That's a visible flash on every page load and a sign we're routing around a hydration issue rather than fixing it. If the upstream data is user-dependent and unsafe to render on the server, render a stable skeleton — but don't hand the user an empty nav and then swap.

### 11. `EditableSection.defaultIsDirty` uses `JSON.stringify`

```ts
function defaultIsDirty<T>(draft: T, initial: T) {
  return JSON.stringify(draft) !== JSON.stringify(initial);
}
```

Non-deterministic for object key order, drops `undefined`, breaks on `Date` / `Set` / `Map`. Org-settings and endpoint pages happen to be safe today, but the first time someone passes a draft containing a `Date` we'll get a false-clean state. Pull in `fast-deep-equal` or document the constraint loudly.

### 12. Context values are not memoized (re-render storms)

- `AppShell.tsx` line 58: `<SidebarCollapseContext.Provider value={{ collapsed, toggle }}>` — both fresh each render.
- `useEndpointDetail.ts` (lines 181–195) returns a fresh object every render, including `editorWrapperStyle` (a brand-new object each render). Once that's in `EndpointDetailContext`, all 4 tabs re-render on every keystroke in the test input.

Wrap with `useMemo` + `useCallback`. For `useEndpointDetail` consider splitting state vs. actions contexts so tabs only subscribe to what they need.

### 13. Disabled FAB tooltips silently break

```tsx
<Tooltip title={tooltip}><MuiFab disabled={…}>{…}</MuiFab></Tooltip>
```

MUI doesn't fire events on disabled elements, so the tooltip never shows once disabled. `EndpointHeaderActions` has `disabled={isDeleting}` FABs with tooltips — those tooltips break the moment the user clicks delete. Wrap with `<span style={{ display: 'inline-flex' }}>`.

### 14. `SearchPill`'s search button is a no-op

The teal `IconButton` next to the input has `aria-label=\"Search\"` but no `onClick`. Search fires on every keystroke from the `InputBase`. To screen-reader users that's a button that does nothing. Either drop the button (use a decorative `SearchIcon` adornment) or wire it up to refocus the input.

---

## 🟢 Polish / consistency (good follow-ups)

### 15. `Tag` and `GridBadge` are 95% the same component

Same `{ items, maxVisible, gap, emptyLabel }` API, same `+N` overflow, same empty-state, same theme-mode dance. The only difference is `borderRadius` (xs vs pill) and font weight. Collapse into one component with a `shape=\"rect\" | \"pill\"` prop.

### 16. `BadgeChip.tsx` shim has zero callers — delete it now while the redesign is in flight.

### 17. `ViewField.inputSx` typed as `React.CSSProperties` — should be `SxProps<Theme>` so callers can use `theme =>` callbacks.

### 18. `GridBadge` hand-types the theme as `(theme: { palette: { mode: string } })`. Use `(theme: Theme)` from `@mui/material/styles` — otherwise the codebase loses access to the `palette.greyscale` augmentation, and the pattern propagates.

### 19. `'#fff'` literals where `primary.contrastText` should go — `Fab.tsx`, `FilterButton.tsx`, `SearchPill.tsx`, `GridToolbar.tsx`, `Sidebar.tsx`. The theme already exposes a contrast token for both modes.

### 20. `FilterButton` test selector is fragile

```ts
expect(container.querySelector('[aria-hidden=\"true\"]')).toBeInTheDocument();
```

That matches the active-filter dot **and** every MUI SVG icon (they're all `aria-hidden`). The test passes for the wrong reason. Give the dot a `data-testid` and assert on that.

### 21. Inconsistent spacing units

Mix of `px: 4` (theme spacing) and `px: '14px'` (pixel strings) — sometimes within the same component. Pick one and stick to it; right now the theme spacing scale is undermined and global density tweaks become impossible.

### 22. `AppShell` knows that `/architect` wants full-bleed

```ts
const FULL_BLEED_PATH_PREFIXES = ['/architect'] as const;
```

A layout shell shouldn't enumerate routes. The cleaner Next App-Router answer is a nested `architect/layout.tsx` that overrides the padding.

### 23. `BreadcrumbItem` carries both Toolpad-era and new naming

```ts
interface BreadcrumbItem { label?: string; title?: string; href?: string; path?: string; }
```

Toolpad is removed in this PR (`@toolpad/core` deleted from `package.json`). Let's commit to `{ label, href }` and update callers — otherwise the next contributor will keep adding aliases.

### 24. `pathname?.startsWith('/onboarding')` also matches `/onboarding-foo`

Use `pathname === '/onboarding' || pathname.startsWith('/onboarding/')`. Tiny but a real bug-class.

### 25. Test coverage of new primitives is thin

`AppShell`, `PageLayout`, and `Sidebar` each have a small smoke test. Missing:
- `<DetailTabNav>` (the most a11y-risky new component — no keyboard tests)
- `<FilterDrawerShell>` / `<FilterSection>`
- `<EditableSection>` save / cancel / dirty flow
- `useEndpointDetail` (save, duplicate, runTest)
- Sidebar dark-mode toggle actually flips mode (this is the bit most likely to silently break given finding #1)

Given the scale of the refactor I'd add at least one interaction test for each new primitive in `components/common/`.

### 26. Inconsistent default vs named exports across new primitives. Pick one — named-only plays nicer with auto-import.

### 27. `getEnvironmentChipColor()` always returns `'default'` with a 7-member return type. Either kill it or actually map environments → colors.

---

## Quick scoreboard

| Area | Bar |
|---|---|
| Theming / dark mode | Broken in 40+ files; needs lint enforcement |
| Accessibility | Significant gaps: tablist arrow keys, clickable Box rows, disabled FAB tooltips, missing `type=\"button\"` |
| Component reuse | Filter drawer duplicated 13×; Tag/GridBadge duplicated; sidebar god-component |
| App Router idioms | `params.then` instead of `use()` in 13 pages; module-scoped `QueryClient` |
| Error handling | Silent `catch {}` in destructive paths |
| Test coverage | Thin for new primitives; a11y / interaction tests missing |

If we can land the 🔴 bucket before this goes to users (especially #1 and #2), I think the rest is fair game as follow-ups. The 🟡 bucket is worth filing tickets for so it doesn't get lost under the next round of feature work.

Again — really nice work on this. The end state is going to be much easier to maintain than what we had.

PR #1793, opened 2026-05-21
added audit-hash-key into worker workflow

asad-miah:
- added audit has key into the worker deployment


harry-rhesis:
@asadaaron thank you for fixing this one!

PR #1804, opened 2026-05-21
Fix preflight dry-run for multi-turn metrics

harry-rhesis:
## Summary
- Multi-turn metrics (e.g., `ConversationalJudge`) require `conversation_history` as a mandatory parameter in their `evaluate()` method
- The preflight `_evaluate_metrics_dry_run` was not supplying it, causing these metrics to fail with a missing argument error during preflight checks
- Passes a dummy `ConversationHistory` to the evaluator so all metric types can be validated

## Test plan
- [ ] Run preflight checks on a test set that includes multi-turn metrics — should no longer crash
- [ ] Verify single-turn metrics still pass preflight as before

### DIFF
--- a/apps/backend/src/rhesis/backend/app/services/preflight.py
+++ b/apps/backend/src/rhesis/backend/app/services/preflight.py
@@ -609,10 +609,12 @@ async def _evaluate_metrics_dry_run(
     model = get_evaluation_model_with_override(db, user, model_id=evaluation_model_id)
     org_id = str(user.organization_id) if user.organization_id else None
 
-    dummy_conversation = ConversationHistory.from_messages([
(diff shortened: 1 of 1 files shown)


harry-rhesis:
Closing this PR as superseded by #1806.

The fix here passes a dummy `ConversationHistory` into `_evaluate_metrics_dry_run` so multi-turn metrics don't crash on a missing `conversation_history` argument. However, #1806 ("improve preflight check performance and reliability") was merged the following day and **removed the full LLM dry-run approach entirely**. Metric validation now uses a lightweight instantiation check (`prepare_metrics`) that never calls `evaluator.evaluate()`, so the missing-argument error this PR targets can no longer occur.

The issue is fixed — just by a different mechanism. No further action needed here.

PR #1805, opened 2026-05-21
Add metric-endpoint compatibility preflight check

harry-rhesis:
## Purpose
Adds a new preflight check that validates metric data requirements against endpoint capabilities before test execution begins. Previously, metrics requiring context (e.g., RAG faithfulness) or tool calls would silently fail at execution time if the endpoint didn't provide that data.

## What Changed
- Added `check_metric_compatibility` preflight check that inspects each metric's requirements (`context_required`, `ground_truth_required`, tool_calls class names) against the endpoint's `response_mapping`
- Extracted shared metric resolution logic into `_resolve_metrics()` helper, used by both compatibility and functionality checks
- Added `_infer_endpoint_capabilities()` to derive endpoint data capabilities from its response mapping
- Added `_check_metric_endpoint_issues()` to identify specific incompatibilities per metric

## Additional Context
- Incompatibilities are reported as **warnings** (not failures) so users can still proceed if they know what they're doing
- Ground truth check queries `Prompt.expected_response` via the test set's tests
- Tool calls check uses a known set of class names (`TOOL_CALLS_CLASS_NAMES`)
- All queries follow existing RLS patterns (no explicit org filters)

## Testing
1. Configure an endpoint without `context` in its `response_mapping`
2. Assign a metric with `context_required=True` to the test set
3. Run preflight — should see a WARNING for metric compatibility
4. With matching endpoint capabilities — should see PASSED

### DIFF
--- a/apps/backend/src/rhesis/backend/app/services/preflight.py
+++ b/apps/backend/src/rhesis/backend/app/services/preflight.py
@@ -10,6 +10,7 @@ from sqlalchemy.orm import Session
 from rhesis.backend.app.models.behavior import Behavior
 from rhesis.backend.app.models.endpoint import Endpoint
 from rhesis.backend.app.models.metric import Metric, behavior_metric_association
+from rhesis.backend.app.models.prompt import Prompt
 from rhesis.backend.app.models.test import Test
 from rhesis.backend.app.models.test_set import TestSet, test_test_set_association
 from rhesis.backend.app.models.user import User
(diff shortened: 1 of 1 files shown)


harry-rhesis:
Closing in favour of #2067, which ports this feature onto the refactored preflight package introduced by #1806 and addresses both peqy review comments (ground-truth partial-coverage check and warning message wording).

PR #1814, opened 2026-05-26
feat(behaviors): add tags for grouping behaviors

nicolai-rhesis:
## Purpose

Let users group behaviors by tag — e.g. by department (Marketing, Customer Support, Legal) or by user story (US 1, US 2, US 3) — so they can quickly slice the directory by logical category. Implements the chip-section pattern from [Figma node 841:38558](https://www.figma.com/design/RCN0J2AjA0UlStdPpdjUCu/Frontend?node-id=841-38558&m=dev).

## What changed

**Backend**
- `Behavior` model now mixes in `TagsMixin` (polymorphic via the existing `TaggedItem` table — no migration needed).
- `schemas.Behavior` exposes `tags: list[Tag]`; `BehaviorWithMetricsSchema` inherits it.
- `tests/backend/routes/test_behavior.py`: new `TestBehaviorTags` suite — empty default, list/detail response shape, assign reflects, remove reflects.

**Frontend**
- `BehaviorCard` renders a Tags chip section below Metrics (caps at 5 visible + `+N more`).
- `BehaviorDrawer` includes an Autocomplete (`multiple`, `freeSolo`) seeded by tag names already in use on other behaviors, so typing "Mar" suggests "Marketing".
- `BehaviorsClient` diffs old vs new tag names on save and applies them via `TagsClient.assignTagToEntity` / `removeTagFromEntity`. Works for both create and edit. Tag matches also flow through the search box.
- `BehaviorFilterDrawer` has a new Tags section with multi-select chip toggles (OR semantics). Applied client-side alongside the existing metric-count filter.
- Tag suggestions and filter chips share a single memoised `availableTagNames` derived from the loaded list — no new endpoint.

## Testing

- Frontend: \`cd apps/frontend && npx jest BehaviorCard.test --no-coverage\` — 11/11 pass (3 new tests for the Tags section).
- TypeScript: \`npx tsc --noEmit\` clean.
- ESLint: clean on all modified files.
- Backend: ruff check + format clean on modified files. Full backend pytest run blocked locally by a pre-existing `pyproject.toml` issue (\`exclude-newer = "1 week"\`) unrelated to this PR; CI will validate.

## Manual test plan

- [ ] Load `/behaviors`; Metrics section unchanged.
- [ ] Confirm a Tags section renders below Metrics with "No tags assigned" for untagged behaviors.
- [ ] Open the drawer for an existing behavior; add an existing tag (suggestion) and a new tag; save; card reflects both.
- [ ] Edit the same behavior; remove a tag; save; card reflects removal.
- [ ] Create a new behavior with two tags; verify both are attached after save.
- [ ] Open the filter drawer; pick one or more tags; verify list narrows; reset clears.
- [ ] Search by tag name in the toolbar; verify it matches.

## Out of scope

- No alembic migration — uses the existing polymorphic `TaggedItem` table.
- No server-side `?tag=` filter; client-side over the loaded list (100 max). Revisit if behaviors are paginated.
- No tag display in `SelectBehaviorsDialog` (the behavior picker dialog).

### DIFF
--- a/apps/backend/src/rhesis/backend/app/models/behavior.py
+++ b/apps/backend/src/rhesis/backend/app/models/behavior.py
@@ -8,12 +8,19 @@ from .mixins import (
     CommentsMixin,
     CountsMixin,
     OrganizationAndUserMixin,
+    TagsMixin,
     TasksMixin,
 )
 
 
 class Behavior(
-    Base, ActivityTrackableMixin, OrganizationAndUserMixin, CommentsMixin, TasksMixin, CountsMixin
+    Base,
+    ActivityTrackableMixin,
+    OrganizationAndUserMixin,
+    TagsMixin,
+    CommentsMixin,
+    TasksMixin,
+    CountsMixin,
 ):
     __tablename__ = "behavior"
     name = Column(String, nullable=False)
--- a/apps/backend/src/rhesis/backend/app/routers/behavior.py
+++ b/apps/backend/src/rhesis/backend/app/routers/behavior.py
@@ -87,6 +87,7 @@ def read_behaviors(
         sort_order=sort_order,
         filter=filter,
         nested_relationships={"metrics": ["metric_type", "backend_type"]},
+        selectin_chains=[["_tags_relationship", "tag"]],
         organization_id=organization_id,
         user_id=user_id,
     )
--- a/apps/backend/src/rhesis/backend/app/schemas/behavior.py
+++ b/apps/backend/src/rhesis/backend/app/schemas/behavior.py
@@ -1,8 +1,9 @@
-from typing import Optional
+from typing import List, Optional
 
 from pydantic import UUID4
 
 from rhesis.backend.app.schemas import Base
+from rhesis.backend.app.schemas.tag import Tag
 
 
 # Behavior schemas
@@ -23,4 +24,4 @@ class BehaviorUpdate(BehaviorBase):
 
 
 class Behavior(BehaviorBase):
-    pass
+    tags: Optional[List[Tag]] = []
--- a/apps/backend/src/rhesis/backend/app/schemas/behavior.py
+++ b/apps/backend/src/rhesis/backend/app/schemas/behavior.py
@@ -1,6 +1,6 @@
 from typing import List, Optional
 
-from pydantic import UUID4
+from pydantic import UUID4, Field
 
 from rhesis.backend.app.schemas import Base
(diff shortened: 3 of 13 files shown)


harry-rhesis:
Nice work @nicolai-rhesis — clean use of `TagsMixin` and the tests cover the full lifecycle well.

I pushed one follow-up commit (`b6fb0c0`) that addresses a performance issue with the list endpoint:

**Problem:** `_tags_relationship` is a polymorphic one-to-many (no `secondary` table), so `with_optimized_loads` classifies it as one-to-many and skips it. When Pydantic serialises the response, it triggers N + N×M lazy loads (one query per behavior to load `TaggedItem` rows, then one per item to resolve the `Tag`). With 100 behaviors × 4 tags = ~500 queries per request.

**Fix:** Added `QueryBuilder.with_selectin_chain()` — a method for chained `selectinload` on polymorphic collections. The behavior router now passes `selectin_chains=[["_tags_relationship", "tag"]]` to `get_items_detail`, collapsing those into exactly 2 batched queries regardless of list size.

Stress-tested locally: 100 behaviors with ~4 tags each → **82ms median** (30 sequential requests). The `selectin_chains` param on `get_items_detail` is additive (`None` default, no-op when unset) so existing callers are unaffected.

Backend review looks good overall — approved from my side.

harry-rhesis:
A few small frontend notes:

**Reference identity bug in the drawer effect (medium)**
`onEdit` passes `(behavior.tags ?? []).map(t => t.name)` inline, creating a new array reference on every parent render. Since `initialTagNames` is in the `useEffect` dep array inside `BehaviorDrawer`, every re-render while the drawer is open re-fires the effect and resets `currentTagNames` — discarding any tags the user has typed but not yet saved. The stored `editingBehavior.tagNames` is already stable; passing that instead of the inline `.map()` would fix it.

**Sequential tag sync (low)**
`syncBehaviorTags` awaits each remove and add sequentially in `for...of` loops. `Promise.all` for the removes and separately for the adds would parallelise those into two batched rounds instead of N serial round-trips.

**Partial failure leaves state stale (low)**
If `syncBehaviorTags` throws mid-loop, the `getBehaviorWithMetrics` re-fetch that follows never runs, so the card shows stale tag data. Always re-fetching in a `finally` block (or moving the re-fetch outside the success path) would keep the UI consistent even on partial failure.

Everything else looks clean — imports, memoisation, filter logic, search matching, and the card tests all look good.

PR #1821, opened 2026-05-27
fix(frontend): resolve BACKEND_URL at runtime for /api/auth-config

nicolai-rhesis:
## Purpose

`dev-app.rhesis.ai` shows "Failed to load authentication options" and no
login buttons. Browser hits `GET /api/auth-config` and gets a 500.
Frontend Cloud Run revision logs show:

```
Error: getaddrinfo EAI_AGAIN backend
```

Despite `BACKEND_URL=https://dev-api.rhesis.ai` being correctly set on
the Cloud Run service.

### Root cause

Next.js external `rewrites()` are evaluated at **build time**, not at
runtime. The returned `destination` strings are serialised into
`.next/routes-manifest.json` during `docker build`. `BACKEND_URL` is
**not** passed as a `--build-arg` from `.github/workflows/frontend.yml`,
so during the image build `process.env.BACKEND_URL` is empty and the
fallback in `next.config.mjs` kicks in:

```js
const backendUrl = process.env.BACKEND_URL || 'http://backend:8080';
```

The manifest is then permanently baked with
`destination: 'http://backend:8080/auth/providers'`. Cloud Run setting
the correct `BACKEND_URL` at runtime can't undo that — by then the
manifest already exists in the image. Hence `EAI_AGAIN backend`.

This regressed for users when PR #1815 switched `AuthForm` from
`getClientApiBaseUrl()` to the same-origin `/api/auth-config` path,
because that path was the one wired through the broken rewrite.

## What Changed

- **New Route Handler** `apps/frontend/src/app/api/auth-config/route.ts`
  - Forwards `GET /api/auth-config` to `getServerBackendUrl() + /auth/providers`
  - Preserves the `?org=` query param
  - Marked `export const dynamic = 'force-dynamic'`
  - Returns 502 with a JSON body if the backend is unreachable
  - Runs **per request**, so `process.env.BACKEND_URL` is read from the
    Cloud Run runtime env — one image now works across local / dev / stg / prd
- **Removed** the `/api/auth-config` rewrite block from `next.config.mjs`
- **Added build-time warning** in `next.config.mjs rewrites()` when
  `BACKEND_URL` is empty or points at the bare `backend` host, so the
  same regression is visible in CI build logs next time instead of only
  at runtime

## Why a Route Handler instead of fixing the rewrite

Either approach would unstick today's symptom, but:

- Passing `BACKEND_URL` as a `--build-arg` couples every backend URL
  change to a frontend rebuild (and to image-per-environment).
- A Route Handler lets us promote the *same* image across dev → stg → prd
  the way Cloud Run runtime env was always intended to be used.

The generic `/api/:path*` rewrite is the same latent footgun and should
be migrated similarly in a follow-up; see "Follow-ups" below.

## Verification

Local checks pass:

- `npx prettier --write` — no changes
- `npx tsc --noEmit` — clean
- `npx eslint src/app/api/auth-config/route.ts next.config.mjs` — clean

After this merges to `main` and `[Deploy] Frontend` runs against dev:

- [ ] `curl -s https://dev-app.rhesis.ai/api/auth-config | jq .providers`
      returns the provider list (200)
- [ ] `dev-app.rhesis.ai` renders login options
- [ ] Frontend Cloud Run revision logs no longer contain
      `getaddrinfo EAI_AGAIN backend`

## Follow-ups (separate PRs)

- Audit stg and prd frontend Cloud Run revisions for the same baked-in
  `backend` hostname in `routes-manifest.json` before they're hit by
  the same regression
- Replace the generic `/api/:path*` rewrite with either per-route
  handlers or a single passthrough Route Handler so no API path
  depends on build-time `BACKEND_URL` resolution


nicolai-rhesis:
## Self-review before human review

Posting a critical pass over my own change so the human reviewer has something to push back on, not just a green diff.

### Things I'd want a reviewer to challenge

1. **Header & cookie forwarding is a behavior regression vs the rewrite.**
   The old `next.config.mjs` rewrite forwarded every request header (Cookie, Accept-Language, Authorization, User-Agent…) and every response header (Set-Cookie, Cache-Control, …) by default. The new Route Handler only forwards `content-type` on the response and forwards **no** request headers.

   For `/auth/providers` specifically this is fine — it's an unauth endpoint that takes only `?org=` and returns JSON, no cookies in either direction. But if we generalise this pattern (see point 4), we have to think about which headers to pass through. Worth a `// behavior note:` comment in the handler, or worth doing properly now.

2. **No timeout / abort on the upstream fetch.**
   If the backend hangs, this handler hangs for whatever Node's default socket timeout is. We should probably add an `AbortController` with ~5s, otherwise a slow backend can pin Cloud Run frontend instances. Not a regression (the rewrite had the same property), but a good "while we're here" fix.

3. **The build-time `console.warn` false-positives on docker-compose builds.**
   In local docker-compose, `BACKEND_URL=http://backend:8080` is **correct** — the compose network resolves `backend`. My warning says "unreachable from Cloud Run regardless of the runtime BACKEND_URL", which is misleading in that context. Either:
   - Tighten the warning to only fire when we can tell we're building for Cloud Run (e.g. when `FRONTEND_ENV` is `dev`/`stg`/`prd`), or
   - Reword to "will only resolve in environments where `backend` is a known host (docker-compose); will fail on Cloud Run."

   I went with option B in spirit but the wording is too absolute.

4. **The `/api/:path*` rewrite is still in `next.config.mjs` with the same build-time bake-in trap.**
   I called this out as a follow-up, but a reviewer could reasonably ask "why ship the partial fix?" Two answers, neither great:
   - Grep shows the only client-side `fetch('/api/...')` calls in the repo target `/api/auth-config`, `/api/feedback`, `/api/users/request-polyphemus-access` — all of which have Route Handlers. So in practice the `/api/:path*` rewrite is dead code today.
   - I'd rather not expand the blast radius of this PR before we've verified the Route-Handler approach works on dev.

   If the reviewer prefers, I'm happy to delete the generic rewrite in this PR.

5. **The Route Handler probably shadows the rewrite already; removing the rewrite is belt-and-suspenders.**
   In App Router, when `rewrites()` returns an array (not `{beforeFiles, afterFiles, fallback}`), it has `afterFiles` semantics — filesystem-style routes (including `route.ts`) take precedence. So just adding `src/app/api/auth-config/route.ts` would likely have fixed the live issue even without touching `next.config.mjs`. I removed the rewrite anyway because:
   - "Likely" is not "definitely" — I haven't seen a definitive Next.js doc statement that App Router route handlers count as filesystem routes for rewrite precedence (Pages Router definitely did).
   - Two sources of truth for the same path is a maintenance hazard.

   But it's worth knowing the rewrite removal is *belt-and-suspenders*, not the fix itself.

6. **No `Cache-Control: no-store` on the response.**
   `export const dynamic = 'force-dynamic'` prevents Next.js from caching, and `cache: 'no-store'` prevents fetch caching, but neither sets a downstream cache header. If there's ever a CDN in front of Cloud Run, the provider list could be cached cross-org. Cheap to add:
   ```ts
   headers: {
     'content-type': ...,
     'cache-control': 'no-store',
   },
   ```

7. **Test coverage.**
   No test for the new Route Handler. The closest neighbour (`apps/frontend/src/app/api/feedback/route.ts`) doesn't have one either, so I'm following the existing pattern, but a tiny `vitest`/`jest` that mocks `fetch` and asserts the `?org=` passthrough would catch the next regression.

### Things I deliberately did NOT do in this PR

- Did **not** add `BACKEND_URL` as a `--build-arg` in `Dockerfile` + `.github/workflows/frontend.yml`. That would have fixed the symptom too, but it couples every backend URL change to a rebuild and forces image-per-environment. The Route Handler approach lets us promote the same image dev → stg → prd, which is what Cloud Run runtime env was meant for.
- Did **not** touch stg/prd. Need to verify the fix on dev first.
- Did **not** delete the generic `/api/:path*` rewrite (see point 4).

### Verification status

Local only:
- `prettier --write` — no changes
- `tsc --noEmit` — clean
- `eslint apps/frontend/src/app/api/auth-config/route.ts apps/frontend/next.config.mjs` — clean

Runtime verification will happen after `[Deploy] Frontend` against dev. The acceptance criteria are in the PR description.

Happy to address any combination of points 1–7 before this gets undrafted.

harry-rhesis:
## Review: centralized runtime proxy

Pushed a follow-up commit that addresses the broader fragility beyond the auth-config fix.

### What changed

**New: `src/utils/backend-proxy.ts`** — single `proxyToBackend()` utility that every `/api/*` route handler uses. Resolves `BACKEND_URL` at request time via `getServerBackendUrl()`, so no build args are ever needed. Includes:
- `AbortController` with configurable timeout (default 10s, auth-config uses 5s)
- `cache-control: 'no-store'` on all responses
- Request header forwarding (`Authorization`, `Content-Type`, `Accept`, etc.)
- Response header forwarding (`Content-Type`, `Content-Disposition`, `Location`)
- `redirect: 'manual'` so presigned-URL 302s pass through to the browser
- Structured error responses: 504 on timeout, 502 on unreachable

**New: `src/app/api/[...path]/route.ts`** — catch-all Route Handler that replaces the build-time `next.config.mjs` rewrite for all `/api/*` paths. More specific handlers (`auth-config`, `feedback`, `auth/[...nextAuth]`) take precedence automatically.

**Simplified: `auth-config/route.ts`** — from 49 lines to 18, now just delegates to `proxyToBackend()` with `backendPath: '/auth/providers'`.

**Removed: all `/api/*` rewrites from `next.config.mjs`** — the build-time footgun is gone entirely. Left a comment explaining why rewrites must never be used for API paths.

**Fixed: `proxy.ts`** — line 71 now uses `getServerBackendUrl()` instead of raw `process.env.BACKEND_URL` for consistent localhost IPv4 resolution on macOS.

### Why this prevents future regressions

The `rewrites()` function is deleted, so there's no build-time destination to accidentally bake in. Every `/api/*` path flows through a Route Handler that reads `BACKEND_URL` at runtime. New backend endpoints work automatically through the catch-all — no per-path config needed.

### Points from the self-review that are now addressed

- **Point 2 (timeout):** ✅ `AbortController` with configurable timeout
- **Point 4 (generic rewrite):** ✅ removed entirely, replaced by catch-all handler
- **Point 6 (cache-control):** ✅ `no-store` on every proxied response
- **Point 1 (header forwarding):** ✅ explicit allowlists for request and response headers

harry-rhesis:
@nicolai-rhesis, thanks for this PR. Made some minor changes to it. Merging.

PR #1848, opened 2026-06-01
Feat/Service Deployment Workflow For K8s

### LINKED ISSUE #1851: Implement GitHub Workflow for Container Image Build and Push to GCP
**Description**

Create  GitHub Actions workflows (for individual and all services) to automatically build the application container image and push it to the GCP container registry for Kubernetes deployments.

**Acceptance Criteria**

* GitHub Actions workflow is configured.
* Container image is built on code changes to the target branch.
* Image is pushed successfully to the designated GCP registry.
* Authentication is handled securely using GitHub secrets/workload identity.
* Image tags include version or commit SHA for traceability.

**Tasks**

* [ ] Configure GCP authentication.
* [ ] Create GitHub Actions workflow.
* [ ] Build container image.
* [ ] Push image to GCP registry.
* [ ] Validate image availability for Kubernetes deployments.

asad-miah:
This PR introduces changes from the `feat/service-deployment-workflow-for-k8s` branch.

## 📁 Files Changed (      10 files)

```
.github/workflows/backend-k8s.yml
.github/workflows/chatbot-k8s.yml
.github/workflows/deploy_all_for_K8s.yml
.github/workflows/docs-k8s.yml
.github/workflows/frontend-k8s.yml
.github/workflows/k8s-build-push.yml
.github/workflows/otel-collector-k8s.yml
.github/workflows/polyphemus-k8s.yml
.github/workflows/telemetry-processor-k8s.yml
.github/workflows/worker-k8s.yml
```

## 📋 Commit Details

```
0d474b5cf - added: docs services (Md Asaduzzaman Miah, 2026-06-01 17:44)
5b7d9bf45 - fix: otel-collector and processor build (Md Asaduzzaman Miah, 2026-06-01 17:15)
a79062ac0 - fix: no builder found (Md Asaduzzaman Miah, 2026-06-01 17:02)
3cb940309 - workflow-for-each-service (Md Asaduzzaman Miah, 2026-06-01 16:42)
911095db1 - refactor: added graphics to the deployments. (Md Asaduzzaman Miah, 2026-06-01 15:29)
dc31e2dff - implement deploy all the k8s workflow (Md Asaduzzaman Miah, 2026-05-27 17:21)
```

## ✅ Checklist

- [ ] All tests pass locally


.github/workflows/backend-k8s.yml
```diff
@@ -0,0 +1,111 @@
...
+
+      - name: Validate environment parameter
+        uses: ./.github/actions/validate-environment
+        with:
+          environment: ${{ inputs.environment || github.event.inputs.environment }}
+
+  build:
+    needs: validate-inputs
+    runs-on: ubuntu-latest
+    permissions:
```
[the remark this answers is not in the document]
asad-miah: addressed.

.github/workflows/backend-k8s.yml
```diff
@@ -0,0 +1,110 @@
...
+    environment: ${{ inputs.environment || github.event.inputs.environment || 'dev' }}
+
+    env:
+      ENVIRONMENT: ${{ inputs.environment || github.event.inputs.environment || 'dev' }}
+      IMAGE: rhesis-backend
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Resolve GCP credentials
```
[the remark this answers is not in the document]
asad-miah: https://github.com/rhesis-ai/rhesis/issues/1852

.github/workflows/backend-k8s.yml
```diff
@@ -0,0 +1,110 @@
...
+        options:
+          - dev
+          - stg
+          - prd
+      tag:
+        description: 'Image tag override (default: short git SHA)'
+        required: false
+        default: ''
+        type: string
+  workflow_call:
```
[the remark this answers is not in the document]
harry-rhesis: Intentional — each service workflow supports both triggers so it can be run individually for targeted deploys (e.g. hotfix a single service) as well as being called from `deploy_all_for_K8s` for a full stack deployment.

.github/workflows/backend-k8s.yml
```diff
@@ -0,0 +1,110 @@
...
+    permissions:
+      contents: read
+    environment: ${{ inputs.environment || github.event.inputs.environment || 'dev' }}
+
+    env:
+      ENVIRONMENT: ${{ inputs.environment || github.event.inputs.environment || 'dev' }}
+      IMAGE: rhesis-backend
+
+    steps:
+      - uses: actions/checkout@v6
```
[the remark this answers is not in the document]
harry-rhesis: Confirmed valid — `actions/checkout@v6` is the current latest release (v6.0.2, released 2026-01-09). No issue here.

.github/workflows/backend-k8s.yml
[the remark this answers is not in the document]
harry-rhesis: To clarify the impact: the `trap ... ERR` is scoped to the *Resolve GCP credentials* shell step only. If a later step (e.g. docker build, kubectl apply) fails or the job is cancelled, the trap never fires and `sa-key.json` stays on disk for the remainder of the runner's lifetime.

In practice the risk is **low** because GitHub-hosted runners (`ubuntu-latest`) are ephemeral VMs — the disk is wiped after each job. However, if the runner were ever self-hosted or reused, the key file would persist across jobs and could be picked up by a subsequent unrelated workflow run.

The safe fix is `trap 'rm -f sa-key.json' EXIT` (EXIT fires unconditionally on shell exit regardless of how the step ends), or writing the key to `$RUNNER_TEMP` which GitHub clears automatically. Both are low-effort changes. The longer-term fix (OIDC/WIF) eliminates the key from disk entirely and is tracked in #1852.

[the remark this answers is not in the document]
harry-rhesis: @asadaaron can you apply the easy fix here — change `trap 'rm -f sa-key.json' ERR` to `trap 'rm -f sa-key.json' EXIT` in the *Resolve GCP credentials* step? One-line change across all the `*-k8s.yml` workflows that use the SA key.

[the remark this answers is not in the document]
harry-rhesis: Actually, a better approach than fixing `trap` in every workflow individually: extract the credential setup into a **composite action** (e.g. `.github/actions/gcp-auth/action.yml`).

Composite actions support a `post:` step that GitHub runs automatically after the job finishes — regardless of success, failure, or cancellation. The key cleanup lives in one place and all workflows get it for free just by calling the action.

```yaml
# .github/actions/gcp-auth/action.yml
name: "GCP Auth"
description: "Write SA key and clean up on exit"
inputs:
  gcp_sa_key:
    required: true
runs:
  using: composite
  steps:
    - name: Write SA key
      shell: bash
      run: |
        echo "${{ inputs.gcp_sa_key }}" | base64 -d > sa-key.json
        chmod 600 sa-key.json
        gcloud auth activate-service-account --key-file=sa-key.json
  post:
    - name: Cleanup SA key
      shell: bash
      run: rm -f sa-key.json
```

Each `*-k8s.yml` workflow then just calls `uses: ./.github/actions/gcp-auth` instead of duplicating the credential block. No `trap` needed anywhere.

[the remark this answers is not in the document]
asad-miah: replaced with : trap 'rm -f sa-key.json' EXIT

harry-rhesis (commented):
Thanks @asadaaron -- solid work here. The per-service workflow structure, single approval gate, concurrency guard, input validation composite action, and the SA key cleanup are all well-thought-out. The PR also responded quickly to review feedback. Good foundation to build on.

I have added inline comments on specific code lines below, and a few general comments on architectural items that do not block this PR but need follow-up before the prd path is used in anger. Each is labeled **[Blocks merge]** or **[Follow-up PR]** so the priority is clear.

.github/workflows/backend-k8s.yml
```diff
@@ -0,0 +1,123 @@
...
+        uses: actions/checkout@v6
+
+      - name: Validate environment parameter
+        uses: ./.github/actions/validate-environment
+        with:
+          environment: ${{ inputs.environment || github.event.inputs.environment }}
+
+  gate:
+    needs: validate-inputs
+    if: github.event_name == 'workflow_dispatch'
```
harry-rhesis, line 55: **[Follow-up PR] Gate is only enforced on the `workflow_dispatch` path**

The `gate` job is correctly guarded by `if: github.event_name == 'workflow_dispatch'`. However, the `build` job has no `environment:` declaration. When triggered via `workflow_call` from `deploy_all`, the gate is skipped AND the build job runs with no job-level environment binding.

Protection is currently delegated to `deploy_all`'s own `gate` job, which is acceptable as a first step. The gap to address in a follow-up is that `build` should declare `environment:` directly so that environment protection rules and environment-scoped secret isolation apply at the job level regardless of trigger path:

```yaml
  build:
    needs: [validate-inputs, gate]
    environment: ${{ inputs.environment || github.event.inputs.environment }}
```

.github/workflows/backend-k8s.yml
```diff
@@ -0,0 +1,123 @@
...
+    if: |
+      always() &&
+      needs.validate-inputs.result == 'success' &&
+      (needs.gate.result == 'success' || needs.gate.result == 'skipped')
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+
+    env:
+      ENVIRONMENT: ${{ inputs.environment || github.event.inputs.environment || 'dev' }}
```
harry-rhesis: **[Follow-up PR] Silent fallback to `dev` is unsafe**

The `|| 'dev'` default silently resolves to the dev environment if both `inputs.environment` and `github.event.inputs.environment` are empty. In a malformed `workflow_call` invocation this produces no error and uses dev credentials without the caller realizing it.

A deployment workflow with no environment specified is a configuration error, not a valid state. Remove the default and fail fast:

```yaml
ENVIRONMENT: ${{ inputs.environment || github.event.inputs.environment }}
```

Then add a guard at the top of the credentials step:
```bash
if [ -z "$ENVIRONMENT" ]; then
  echo "ERROR: ENVIRONMENT is not set"
  exit 1
fi
```

.github/workflows/backend-k8s.yml
```diff
@@ -0,0 +1,123 @@
...
+      - name: Approve deploy
+        run: echo "Deploy approved for ${{ github.event.inputs.environment }}"
+
+  build:
+    needs: [validate-inputs, gate]
+    if: |
+      always() &&
+      needs.validate-inputs.result == 'success' &&
+      (needs.gate.result == 'success' || needs.gate.result == 'skipped')
+    runs-on: ubuntu-latest
```
harry-rhesis: **[Follow-up PR] No `timeout-minutes` on the build job**

A hung `docker build` or `docker push` (network stall, Artifact Registry issue) will run until the GitHub Actions default of 6 hours. During that time the concurrency group for this environment is locked and runner minutes are consumed.

Add a job-level timeout appropriate for your typical build times:

```yaml
  build:
    needs: [validate-inputs, gate]
    runs-on: ubuntu-latest
    timeout-minutes: 30
```

.github/workflows/backend-k8s.yml
```diff
@@ -0,0 +1,123 @@
...
+        run: |
+          docker build --platform linux/amd64 --target backend \
+            -t "$REGISTRY/$IMAGE:$TAG" \
+            -t "$REGISTRY/$IMAGE:latest" \
+            -f apps/backend/Dockerfile .
+          echo "Built $REGISTRY/$IMAGE:$TAG"
+
+      - name: Push Container
+        run: |
+          docker push "$REGISTRY/$IMAGE:$TAG"
```
harry-rhesis: **[Follow-up PR] Non-atomic push and missing digest logging**

Two separate `docker push` calls mean that if the second fails, `:$TAG` exists in the registry but `:latest` is stale. This is a silent inconsistency with no alert.

Additionally, the pushed image digest (`sha256:...`) is never logged. For incident response -- particularly when `:latest` has been overwritten -- the digest is the only durable identifier of what is actually running.

Suggested follow-up:
1. Switch to `docker/build-push-action` which pushes both tags atomically and supports layer caching (significant CI minute savings across 8 services)
2. Capture and log the digest:

```bash
DIGEST=$(docker inspect --format='{{index .RepoDigests 0}}' "$REGISTRY/$IMAGE:$TAG")
echo "Digest: $DIGEST"
echo "IMAGE_DIGEST=$DIGEST" >> $GITHUB_ENV
```

.github/workflows/frontend-k8s.yml
```diff
@@ -0,0 +1,136 @@
...
+          echo "Image:    $IMAGE"
+
+      - name: Configure Docker for Artifact Registry
+        run: gcloud auth configure-docker europe-west4-docker.pkg.dev --quiet
+
+      - name: Build Container
+        run: |
+          docker build --platform linux/amd64 \
+            --build-arg FRONTEND_ENV="$FRONTEND_ENV" \
+            --build-arg NEXT_PUBLIC_API_BASE_URL="$NEXT_PUBLIC_API_BASE_URL" \
```
harry-rhesis: **[Follow-up PR] Environment-specific URLs baked into the image violate build artifact immutability**

`NEXT_PUBLIC_API_BASE_URL` and `NEXT_PUBLIC_APP_URL` are passed as build args, producing a different image per environment. This means the same artifact cannot be promoted dev to stg to prd -- a separate rebuild is required for each environment even if the application code is identical.

This is a Twelve-Factor App principle V violation (strictly separate build and release stages).

The recommended fix is to inject these values at runtime via a Kubernetes ConfigMap or environment variables in the deployment manifest, and use Next.js runtime configuration (`publicRuntimeConfig` or `serverRuntimeConfig`) rather than build-time `NEXT_PUBLIC_*` substitution. This allows one image to be built once and promoted across all environments.

harry-rhesis:
**[Follow-up PR -- Critical] These workflows build and push images. They do not deploy anything.**

The cluster runs ArgoCD with `automated.selfHeal: true` tracking Git manifests. The Kustomize overlays (`apps/worker/k8s/kustomization.yaml`) and Helm values (`charts/rhesis/values-*.yaml`) currently pin `newTag: latest`. Pushing a new `:latest` image to Artifact Registry does not change any file in Git, so ArgoCD sees no diff and triggers no rollout. A green run of these workflows leaves the cluster unchanged.

To actually deploy after a push, one of the following is needed:

**Option A -- GitOps write-back (recommended for this ArgoCD setup):**
After push, update the image tag in the relevant env overlay and commit to Git. ArgoCD picks it up automatically:
```bash
cd kubernetes/clusters/$ENVIRONMENT/...
kustomize edit set image rhesis-backend=$REGISTRY/$IMAGE@$DIGEST
git commit -am "chore: update backend image to $TAG"
git push
argocd app wait rhesis-backend --timeout 120
```

**Option B -- ArgoCD Image Updater:**
Configure [Argo CD Image Updater](https://argocd-image-updater.readthedocs.io/) to watch the registry and write tags back to Git automatically when a new image is pushed.

**Option C -- explicit sync step:**
Add `argocd app sync` + `argocd app wait` after the push step, though this is the least GitOps-native approach.

Until one of these is in place, I would suggest renaming the workflows from `[Deploy]` to `[Build & Push]` to avoid operational confusion where an on-call engineer runs a "Deploy" workflow, sees green, and assumes the cluster has been updated.

harry-rhesis:
**[Follow-up PR -- High] `:latest` tag in manifests makes rollback impossible and silently runs stale code in prd**

Two related issues in the existing manifests:

1. `apps/worker/k8s/kustomization.yaml` pins `newTag: latest`, and `charts/rhesis/values-prd.yaml` sets `imagePullPolicy: IfNotPresent`. That combination means: once a node has pulled a `:latest` image, a pod restart or rescheduling on that node will run the cached (stale) image without pulling the new one. In production, this can mean different nodes in the same deployment running different code with no alert.

2. `:latest` is a mutable pointer with no history. There is no tag to roll back to. If a bad image is pushed, the only recovery is to rebuild and push a fixed one.

Recommended fixes for a follow-up PR:
- Switch from `:latest` to immutable tags (short git SHA, or better, the image digest) in all Kustomize overlays and Helm values
- Change `imagePullPolicy` in prd from `IfNotPresent` to `Always` as an immediate short-term fix (accepts the latency cost in exchange for correctness)
- Once immutable tags are in use, the GitOps write-back step in the workflow (from the previous comment) handles promotion and rollback naturally by reverting a Git commit

harry-rhesis:
**[Follow-up PR -- High] Registry and image name mismatch between the workflows and the existing manifests**

The workflows push to `europe-west4-docker.pkg.dev/<project>/rhesis/rhesis-worker` (Artifact Registry), but the existing raw Kubernetes manifest at `apps/worker/k8s/deployment.yaml` references `gcr.io/PROJECT_ID/rhesis-worker` (GCR, with a placeholder project ID). The Helm chart values (`charts/rhesis/values-*.yaml`) reference the Artifact Registry path, so the Helm path is aligned.

This means the raw `apps/*/k8s` manifests are not aligned with where images are actually being pushed. Two manifest systems coexist with different registries and different image reference conventions.

Recommended for a follow-up PR:
- Align `apps/worker/k8s/deployment.yaml` (and any other raw manifests) to use the Artifact Registry path
- Decide whether raw Kustomize overlays or Helm charts are the canonical deployment path for each service and remove or clearly deprecate the other. Having both active is a latent incident waiting to happen if the wrong one is applied.

harry-rhesis:
**[Follow-up PR -- High] No image scanning, signing, or SBOM before images are pushed to the registry**

Images are built and pushed directly to production-capable registries with no vulnerability scanning step between build and push. This means a dependency with a known CRITICAL CVE could reach prd undetected.

Recommended for a follow-up PR -- add a scan step gating the push:

```yaml
- name: Scan image for vulnerabilities
  uses: aquasecurity/trivy-action@main
  with:
    image-ref: "${{ env.REGISTRY }}/${{ env.IMAGE }}:${{ env.TAG }}"
    format: table
    exit-code: 1
    severity: CRITICAL
    ignore-unfixed: true
```

Longer term (separate follow-up):
- Image signing with `cosign` to create a verifiable chain of custody from build to deploy
- SBOM generation (Syft or Trivy) attached as a build artifact for compliance and incident response

The scan is a one-step addition that prevents the worst-case supply chain failures.

harry-rhesis:
**[Follow-up PR -- Medium] No `concurrency` on individual service workflows**

`deploy_all_for_K8s.yml` correctly prevents concurrent full-stack deploys via its `concurrency` group. However, the individual service workflows (`backend-k8s.yml`, `worker-k8s.yml`, etc.) have no `concurrency` block of their own.

If `backend-k8s.yml` is triggered manually at the same time `deploy_all` is running (or two manual triggers are fired quickly), two jobs will race to push `:$TAG` and overwrite `:latest` for the same environment. The last one to finish wins, with no indication that a race occurred.

Recommended for a follow-up PR, add to each service workflow:

```yaml
concurrency:
  group: build-${{ github.workflow }}-${{ inputs.environment || github.event.inputs.environment }}
  cancel-in-progress: false
```

This is a one-line addition to each workflow (or a single line once the workflows are consolidated).

harry-rhesis:
**[Follow-up PR -- Medium] 8 near-identical workflows -- consolidate into one reusable workflow**

This PR ships 8 workflows that are ~95% identical. The only per-service differences are `IMAGE`, the Dockerfile path, and occasionally `--target`. Every cross-cutting change -- the `trap` fix, adding a timeout, adding scanning, adding the concurrency group, wiring the deployment step -- must be applied to all 8 files. This review thread is already evidence of that cost.

Recommended for a follow-up PR: extract into a single reusable `build-push-k8s.yml` callable workflow:

```yaml
# .github/workflows/build-push-k8s.yml
on:
  workflow_call:
    inputs:
      service:     { required: true,  type: string }
      image:       { required: true,  type: string }
      dockerfile:  { required: true,  type: string }
      context:     { required: false, type: string, default: "." }
      target:      { required: false, type: string, default: "" }
      environment: { required: true,  type: string }
      tag:         { required: false, type: string, default: "" }
```

Each current `*-k8s.yml` then becomes a ~15-line caller, and `deploy_all` stays as-is. The result is roughly 1,100 lines collapsed to ~200. Any future improvement (caching, scanning, signing, deployment step) ships once and applies everywhere.

.github/workflows/backend-k8s.yml
```diff
@@ -0,0 +1,123 @@
...
+          fi
+
+      - name: Authenticate to Google Cloud
+        run: |
+          gcloud auth activate-service-account --key-file=sa-key.json
+          echo "GOOGLE_APPLICATION_CREDENTIALS=$(pwd)/sa-key.json" >> $GITHUB_ENV
+
+      - name: Set registry and tag
+        run: |
+          REGISTRY="europe-west4-docker.pkg.dev/${{ env.PROJECT_ID }}/rhesis"
```
harry-rhesis: **[Follow-up PR -- Medium] Hardcoded Artifact Registry region**

`europe-west4-docker.pkg.dev` is hardcoded here and again on line 105 (`gcloud auth configure-docker`). This appears identically across all 8 service workflows (~16 hardcoded occurrences in total). If the registry region ever changes (e.g. a second region for DR, or a region migration), every file needs updating.

Move it to a top-level env variable or a workflow input:

```yaml
env:
  ENVIRONMENT: ${{ inputs.environment || github.event.inputs.environment }}
  IMAGE: rhesis-backend
  REGISTRY_REGION: europe-west4-docker.pkg.dev
```

Then use `$REGISTRY_REGION` in both places:
```bash
REGISTRY="$REGISTRY_REGION/${{ env.PROJECT_ID }}/rhesis"
```
```yaml
- name: Configure Docker for Artifact Registry
  run: gcloud auth configure-docker $REGISTRY_REGION --quiet
```

Better still, this becomes a single definition once the workflows are consolidated into one reusable callable (see consolidation comment).

.github/workflows/backend-k8s.yml
```diff
@@ -0,0 +1,123 @@
...
+            TAG=$(echo "${{ github.sha }}" | cut -c1-8)
+          fi
+          echo "TAG=$TAG" >> $GITHUB_ENV
+
+          echo "Registry: $REGISTRY"
+          echo "Tag:      $TAG"
+          echo "Image:    $IMAGE"
+
+      - name: Configure Docker for Artifact Registry
+        run: gcloud auth configure-docker europe-west4-docker.pkg.dev --quiet
```
harry-rhesis: Same hardcoded `europe-west4-docker.pkg.dev` as line 91. Both occurrences should reference the same variable so a region change is a one-line edit.

.github/workflows/backend-k8s.yml
harry-rhesis: To expand on this: the registry region should ideally be derived from the deployment environment, the same way `PROJECT_ID` already is. In GCP, the Artifact Registry hostname is region-bound (`<region>-docker.pkg.dev`). If `dev`, `stg`, and `prd` environments run in different GCP regions -- or if you later move to a multi-region registry for prd (e.g. `europe-docker.pkg.dev`) for lower-latency pulls -- a hardcoded region breaks silently.

The pattern to follow is the same as `PROJECT_ID`: resolve it inside the `Resolve GCP credentials` step:

```bash
if [ "$ENVIRONMENT" = "dev" ]; then
  echo "REGISTRY_REGION=europe-west4-docker.pkg.dev" >> $GITHUB_ENV
  echo "PROJECT_ID=${{ secrets.TF_PROJECT_ID_DEV }}" >> $GITHUB_ENV
  ...
elif [ "$ENVIRONMENT" = "stg" ]; then
  echo "REGISTRY_REGION=europe-west4-docker.pkg.dev" >> $GITHUB_ENV
  echo "PROJECT_ID=${{ secrets.TF_PROJECT_ID_STG }}" >> $GITHUB_ENV
  ...
else
  echo "REGISTRY_REGION=europe-docker.pkg.dev" >> $GITHUB_ENV  # e.g. multi-region for prd
  echo "PROJECT_ID=${{ secrets.TF_PROJECT_ID_PRD }}" >> $GITHUB_ENV
  ...
fi
```

Then `REGISTRY` and the `gcloud auth configure-docker` call both reference `$REGISTRY_REGION`. The current values can all stay `europe-west4` for now -- the point is to make region a first-class per-environment config value rather than a constant baked into the logic.

.github/workflows/backend-k8s.yml
asad-miah: addressed

.github/workflows/backend-k8s.yml
asad-miah: addressed

.github/workflows/backend-k8s.yml
asad-miah: addressed

.github/workflows/frontend-k8s.yml
asad-miah: addressed.

.github/workflows/backend-k8s.yml
asad-miah: addressed

.github/workflows/backend-k8s.yml
asad-miah: addressed

.github/workflows/k8s-build-push.yml
```diff
@@ -0,0 +1,215 @@
...
+            rhesis-polyphemus)          YQ_PATH=".polyphemus.image.tag" ;;
+            rhesis-otel-collector)      YQ_PATH=".otelCollector.image.tag" ;;
+            rhesis-telemetry-processor) YQ_PATH=".telemetryProcessor.image.tag" ;;
+            rhesis-docs)                YQ_PATH=".docs.image.tag" ;;
+            *)                          echo "ERROR: Unknown IMAGE: $IMAGE"; exit 1 ;;
+          esac
+
+          # Fetch main so we patch the branch ArgoCD watches, regardless of trigger branch.
+          git fetch origin main
+          git checkout main
```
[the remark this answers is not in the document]
asad-miah: this is intended behavior when someone work in a feature branch. then before merging to the main, he wants to see whether it breaks something in dev or not. In that case, this workflow will serve this purpose. This actually mimics the existing practice.

asad-miah:
> **[Follow-up PR -- Medium] 8 near-identical workflows -- consolidate into one reusable workflow**
> 
> This PR ships 8 workflows that are ~95% identical. The only per-service differences are `IMAGE`, the Dockerfile path, and occasionally `--target`. Every cross-cutting change -- the `trap` fix, adding a timeout, adding scanning, adding the concurrency group, wiring the deployment step -- must be applied to all 8 files. This review thread is already evidence of that cost.
> 
> Recommended for a follow-up PR: extract into a single reusable `build-push-k8s.yml` callable workflow:
> 
> ```yaml
> # .github/workflows/build-push-k8s.yml
> on:
>   workflow_call:
>     inputs:
>       service:     { required: true,  type: string }
>       image:       { required: true,  type: string }
>       dockerfile:  { required: true,  type: string }
>       context:     { required: false, type: string, default: "." }
>       target:      { required: false, type: string, default: "" }
>       environment: { required: true,  type: string }
>       tag:         { required: false, type: string, default: "" }
> ```
> 
> Each current `*-k8s.yml` then becomes a ~15-line caller, and `deploy_all` stays as-is. The result is roughly 1,100 lines collapsed to ~200. Any future improvement (caching, scanning, signing, deployment step) ships once and applies everywhere.

-> addressed

.github/workflows/k8s-build-push.yml
```diff
@@ -0,0 +1,155 @@
...
+        type: string
+        default: ''
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    environment: ${{ inputs.environment }}
+    permissions:
+      contents: read
```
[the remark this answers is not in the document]
asad-miah: intent is only build+push to Artifact Registry

charts/rhesis/templates/_helpers.tpl
```diff
@@ -139,14 +139,19 @@ Usage: include "rhesis.image" (dict "imageValues" .Values.backend "global" .Valu
...
+  {{- if .global.requirePinnedTag -}}
+    {{- fail "global.imageTag must be set — CI publishes immutable git SHA tags; pin via GitOps before deploy" -}}
+  {{- end -}}
+{{- end -}}
+{{- if eq ($tag | default "") "latest" -}}
+  {{- fail "global.imageTag must not be 'latest' — use an immutable git SHA or digest tag" -}}
 {{- end -}}
 {{- if $registry -}}
-{{- printf "%s/%s:%s" $registry $repository $tag -}}
+{{- printf "%s/%s:%s" $registry $repository ($tag | default "") -}}
```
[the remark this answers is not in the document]
asad-miah: fixed: rhesis.image in _helpers.tpl to avoid a trailing colon when the tag is empty

charts/rhesis/templates/_helpers.tpl
```diff
@@ -139,14 +139,27 @@ Usage: include "rhesis.image" (dict "imageValues" .Values.backend "global" .Valu
...
+  {{- end -}}
+{{- end -}}
+{{- if eq ($tag | default "") "latest" -}}
+  {{- fail "global.imageTag must not be 'latest' — use an immutable git SHA or digest tag" -}}
 {{- end -}}
 {{- if $registry -}}
+  {{- if $tag -}}
 {{- printf "%s/%s:%s" $registry $repository $tag -}}
+  {{- else -}}
+{{- printf "%s/%s" $registry $repository -}}
```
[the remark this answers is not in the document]
asad-miah: latest is fine for the fallback in dev and stg.

harry-rhesis:
**Approval gate — remove for now, wire in later**

The environment-level approval gate is not needed yet. Please make the gate jobs always auto-pass by removing the `environment:` binding (which is what triggers GitHub's approval/wait-timer protection rules) from three places:

1. **`k8s-build-push.yml`** — drop `environment: ${{ inputs.environment }}` from the `build` job. This also closes peqy's "N approvals" concern.

2. **`build-push-all.yml`** — drop `environment: ${{ github.event.inputs.environment }}` from the `gate` job. The job itself (and its `echo`) can stay as a logging checkpoint.

3. **Each `*-k8s.yml`** — drop `environment: ${{ github.event.inputs.environment }}` from the per-service `gate` job. The `if: github.event_name == 'workflow_dispatch'` guard stays.

The `environment:` lines inside `workflow_dispatch.inputs` blocks are input *definitions*, not protection bindings — leave those as-is.

When we're ready to re-introduce approval gates (before prd is used in anger), we wire the `environment:` back in — one line per workflow — and configure the required reviewers in the GitHub environment settings.

harry-rhesis (commented):
Thanks @asadaaron -- good iterative work responding to the review rounds. The consolidation into a single reusable workflow, the SA key handling, the digest logging, and the rename from Deploy to Build & Push are all solid. A few remaining items below, labeled by priority.

---

**[Blocks merge] dev/stg tag mismatch -- pipeline and Helm values are now out of sync**

The workflow pushes `registry/image:<git-sha>` (immutable, no `:latest`). But `values-dev.yaml` and `values-stg.yaml` both set `global.imageTag: ""`, and the `rhesis.image` helper in `_helpers.tpl` renders an untagged image ref when the tag is empty. Kubernetes resolves an untagged ref as `:latest`, which the pipeline no longer publishes.

The result: a successful build-and-push run for dev/stg leaves no new image that the cluster will actually pull. prd is safe (requirePinnedTag: true hard-fails on empty tag), but dev/stg will silently run stale images.

Fix options:
- Set `requirePinnedTag: true` for dev and stg as well, and rely on the GitOps write-back step (to be added in a follow-up) to pin the tag before deploy.
- Or explicitly document that dev/stg tags are expected to be set externally, with a comment in the values files explaining the placeholder.

Either way, `imageTag: ""` in dev/stg with the comment "Do not deploy on floating latest" is currently contradictory.

---

**[Follow-up PR] `REGISTRY_REGION` is the wrong name**

In `k8s-build-push.yml` the variable `REGISTRY_REGION` holds the full Artifact Registry hostname (`europe-west4-docker.pkg.dev`), not a region (`europe-west4`). It is then used as both the Docker registry host and the argument to `gcloud auth configure-docker`. Rename to `REGISTRY_HOST` so it matches what it actually contains. Small change but it will matter when someone is debugging a configure-docker failure at 2am.

---

**[Follow-up PR] Trivy scan is present but commented out**

`k8s-build-push.yml` has the scan step included but disabled with a block comment. That is a good placeholder. Before this workflow is used for stg or prd, please either enable it or open a tracking issue so the scan gate does not get forgotten. A CRITICAL-severity image reaching prd undetected is the scenario we are trying to prevent.

---

**[Approved, confirming scope] No deploy step is intentional**

Confirmed: these workflows stop at `docker push`. There is no `argocd app sync`, no GitOps write-back, and no `kustomize edit set image` commit. The gate job in `build-push-all.yml` even echoes this explicitly. The scope is Build and Push only, with deploy deferred to a follow-up. That is fine -- the naming is now honest and the architecture is the right foundation to add the deploy step onto.

The approval gate comment has been posted separately.

.github/workflows/k8s-build-push.yml
```diff
@@ -0,0 +1,155 @@
...
+        description: 'Image tag override (default: short git SHA)'
+        required: false
+        type: string
+        default: ''
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    environment: ${{ inputs.environment }}
```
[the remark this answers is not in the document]
asad-miah: addressed

.github/workflows/k8s-deploy.yml
```diff
@@ -0,0 +1,103 @@
...
+      contents: none
+
+    steps:
+      - name: Resolve ArgoCD credentials for environment
+        id: argocd-creds
+        run: |
+          ENV="${{ inputs.environment }}"
+          case "$ENV" in
+            dev)
+              echo "server=${{ secrets.ARGOCD_SERVER_DEV }}"         >> "$GITHUB_OUTPUT"
```
harry-rhesis: **[Blocks merge] ArgoCD auth token written to `$GITHUB_OUTPUT`**

Secrets written to `$GITHUB_OUTPUT` are masked in logs but the value is accessible to all subsequent steps via `steps.argocd-creds.outputs.token` and is stored in the workflow run object. GitHub's own guidance is to never route secrets through step outputs.

Fix: set the secrets directly in the `env:` block of each step that needs them, resolving the environment case there:

```yaml
      - name: Set image tag on ArgoCD Application
        env:
          ARGOCD_SERVER: ${{ inputs.environment == 'prd' && secrets.ARGOCD_SERVER_PRD || inputs.environment == 'stg' && secrets.ARGOCD_SERVER_STG || secrets.ARGOCD_SERVER_DEV }}
          ARGOCD_AUTH_TOKEN: ${{ inputs.environment == 'prd' && secrets.ARGOCD_AUTH_TOKEN_PRD || inputs.environment == 'stg' && secrets.ARGOCD_AUTH_TOKEN_STG || secrets.ARGOCD_AUTH_TOKEN_DEV }}
```

Or keep the `case` logic but write only `server` to output (not `token`), and resolve the token separately in `env:`.

.github/workflows/k8s-deploy.yml
```diff
@@ -0,0 +1,103 @@
...
+      - name: Sync ArgoCD Application
+        env:
+          ARGOCD_SERVER: ${{ steps.argocd-creds.outputs.server }}
+          ARGOCD_AUTH_TOKEN: ${{ steps.argocd-creds.outputs.token }}
+        run: |
+          argocd app sync rhesis \
+            --server "$ARGOCD_SERVER" \
+            --auth-token "$ARGOCD_AUTH_TOKEN" \
+            --grpc-web \
+            --timeout 120
```
harry-rhesis, line 87: **[Blocks merge] No rollout health check after sync**

`argocd app sync --timeout 120` initiates the sync and waits up to 2 minutes for the sync operation to complete, but it does not wait for pods to become healthy. If new pods start CrashLooping or fail liveness probes after being scheduled, this step still exits 0 and the deploy job shows green.

Add `argocd app wait` after the sync to gate on pod health:

```yaml
      - name: Wait for rollout to be healthy
        env:
          ARGOCD_SERVER: ${{ steps.argocd-creds.outputs.server }}
          ARGOCD_AUTH_TOKEN: ${{ steps.argocd-creds.outputs.token }}
        run: |
          argocd app wait rhesis \
            --server "$ARGOCD_SERVER" \
            --auth-token "$ARGOCD_AUTH_TOKEN" \
            --grpc-web \
            --health \
            --timeout 300
```

Without this, a failed rollout is only visible in ArgoCD UI, not in the GitHub Actions run that triggered it.

.github/workflows/k8s-deploy.yml
```diff
@@ -0,0 +1,103 @@
...
+            *)
+              echo "ERROR: Unknown environment: $ENV"
+              exit 1
+              ;;
+          esac
+
+      - name: Install ArgoCD CLI
+        run: |
+          ARGOCD_VERSION="v2.14.4"
+          curl -sSL -o /usr/local/bin/argocd \
```
harry-rhesis, line 52: **[Follow-up PR] ArgoCD CLI downloaded without checksum verification**

The binary is fetched directly from GitHub releases over HTTPS (good), but its sha256 is not verified before execution. The CLI then runs with deployment credentials in a privileged path.

Add a checksum check after the download:

```bash
ARGOCD_VERSION="v2.14.4"
ARGOCD_SHA256="<paste sha256 from the release page for argocd-linux-amd64>"
curl -sSL -o /usr/local/bin/argocd \
  "https://github.com/argoproj/argo-cd/releases/download/${ARGOCD_VERSION}/argocd-linux-amd64"
echo "${ARGOCD_SHA256}  /usr/local/bin/argocd" | sha256sum -c -
chmod +x /usr/local/bin/argocd
```

Alternatively, use the official `argoproj/argocd-installer` or pin the action to a SHA.

.github/workflows/k8s-deploy.yml
```diff
@@ -0,0 +1,103 @@
...
+        run: |
+          if [ -z "$SERVICE" ]; then
+            PARAM="global.imageTag=$TAG"
+            echo "Setting global.imageTag=$TAG on app rhesis [${{ inputs.environment }}]"
+          else
+            PARAM="${SERVICE}.image.tag=$TAG"
+            echo "Setting ${SERVICE}.image.tag=$TAG on app rhesis [${{ inputs.environment }}]"
+          fi
+
+          argocd app set rhesis \
```
harry-rhesis: **[Follow-up PR] ArgoCD application name `rhesis` is hardcoded**

If the application name differs per environment (e.g. `rhesis-dev`, `rhesis-prd`) this breaks silently -- `argocd app set` on the wrong app name will error but with a confusing message.

Promote it to an input with a sensible default:

```yaml
    inputs:
      app_name:
        description: 'ArgoCD application name'
        required: false
        type: string
        default: 'rhesis'
```

Then reference `${{ inputs.app_name }}` in both the `app set` and `app sync` calls.

.github/workflows/k8s-build-push.yml
```diff
@@ -0,0 +1,164 @@
...
+
+          docker build --platform linux/amd64 \
+            "${TARGET_ARGS[@]}" \
+            "${BUILD_ARG_FLAGS[@]}" \
+            -t "$REGISTRY/$IMAGE:$TAG" \
+            -f "${{ inputs.dockerfile }}" \
+            "${{ inputs.context }}"
+          echo "Built $REGISTRY/$IMAGE:$TAG"
+
+      # - name: Scan image for vulnerabilities
```
harry-rhesis, line 148: **[Follow-up PR] Trivy scan is commented out**

The scan step was added and then commented out (commit message says "ommit: trivy security scan"). It should be re-enabled before this workflow is used against prd. The code is already correct -- just needs to be uncommented and the `aquasecurity/trivy-action` SHA pinned.

If there is a specific reason it was disabled (e.g. too many false positives on initial run, local registry auth issues with Trivy), please add a comment explaining the blocker and open a follow-up issue so it does not get forgotten.

harry-rhesis (commented):
The changes since the last review are substantial and address the vast majority of the earlier findings in one commit. Great work turning this around quickly.

What is now resolved:
- Consolidated into one reusable `k8s-build-push.yml` (1,100 lines collapsed)
- ArgoCD deploy step wired via `k8s-deploy.yml` (critical gap closed)
- `REGISTRY_HOST` resolved per environment alongside `PROJECT_ID`
- `timeout-minutes: 30` on the build job
- SA key written to `$RUNNER_TEMP`, `chmod 600`, deleted immediately after `gcloud auth` within the same step
- Fail fast if `ENVIRONMENT` is unset
- Digest logged after push
- No `:latest` push
- `imagePullPolicy: Always` in prd
- `imageTag: ""` + `requirePinnedTag` in all env overlays
- Helm helper rejects empty and `latest` tags
- Frontend env URLs moved out of build args

A few remaining items below, inline on the two central files.

harry-rhesis:
**[Follow-up PR -- Low] No `concurrency` block on individual service workflows**

This was raised earlier and is still outstanding. `deploy-all-k8s.yml` has a `concurrency` group, but the individual `*-k8s.yml` workflows do not. Running `backend-k8s.yml` manually at the same time as `deploy-all` will race on the build and push for that service.

Add to each service workflow (or pick it up as part of the consolidation cleanup):

```yaml
concurrency:
  group: k8s-build-${{ github.workflow }}-${{ inputs.environment || github.event.inputs.environment }}
  cancel-in-progress: false
```

harry-rhesis:
**[Informational] ArgoCD `app set` patches live state, not Git**

The `k8s-deploy.yml` approach uses `argocd app set --helm-set` to patch the imageTag directly on the ArgoCD application, without committing anything back to Git. This means:

- `values-dev.yaml`, `values-stg.yaml`, `values-prd.yaml` will always show `imageTag: ""` in Git
- The actual deployed tag only exists in ArgoCD application state (`argocd app get rhesis -o json`)
- A `git blame` or `git log` on the values file gives no history of what was deployed when
- A full cluster re-bootstrap from Git (disaster recovery scenario) would deploy with empty imageTag and fail on `requirePinnedTag`

This is a known trade-off of the "Option C" approach vs. Git write-back. It is not a blocker for day-to-day operations since ArgoCD retains history. But worth being aware of for DR planning, and it means the Git repo is no longer the complete source of truth for what is running.

If this becomes a concern, the path forward is git write-back: after push, run `kustomize edit set image` or update the values file and commit, then ArgoCD syncs from Git state (Option A from earlier comments).

harry-rhesis:
**Final review -- two items to fix before merging, plus a DR note**

---

### 1. `ARGOCD_AUTH_TOKEN` must not go through `$GITHUB_OUTPUT`

Secrets written to step outputs are stored in the workflow run object and accessible to all subsequent steps. The fix is to resolve them directly in the `env:` block of each step that needs them, using GitHub Actions inline conditionals:

```yaml
      - name: Set image tag on ArgoCD Application
        env:
          ARGOCD_SERVER: ${{ inputs.environment == 'prd' && secrets.ARGOCD_SERVER_PRD || inputs.environment == 'stg' && secrets.ARGOCD_SERVER_STG || secrets.ARGOCD_SERVER_DEV }}
          ARGOCD_AUTH_TOKEN: ${{ inputs.environment == 'prd' && secrets.ARGOCD_AUTH_TOKEN_PRD || inputs.environment == 'stg' && secrets.ARGOCD_AUTH_TOKEN_STG || secrets.ARGOCD_AUTH_TOKEN_DEV }}
        run: |
          ...

      - name: Sync ArgoCD Application
        env:
          ARGOCD_SERVER: ${{ inputs.environment == 'prd' && secrets.ARGOCD_SERVER_PRD || inputs.environment == 'stg' && secrets.ARGOCD_SERVER_STG || secrets.ARGOCD_SERVER_DEV }}
          ARGOCD_AUTH_TOKEN: ${{ inputs.environment == 'prd' && secrets.ARGOCD_AUTH_TOKEN_PRD || inputs.environment == 'stg' && secrets.ARGOCD_AUTH_TOKEN_STG || secrets.ARGOCD_AUTH_TOKEN_DEV }}
        run: |
          ...
```

The `Resolve ArgoCD credentials` step and its output can then be removed entirely.

---

### 2. Add `argocd app wait --health` after the sync

`argocd app sync --timeout 120` waits for the sync operation to finish, not for the pods to be healthy. If new pods crash after being scheduled, the sync step still exits 0 and the workflow shows green while the service is down.

Add a wait step immediately after sync:

```yaml
      - name: Wait for rollout to be healthy
        env:
          ARGOCD_SERVER: ...
          ARGOCD_AUTH_TOKEN: ...
        run: |
          argocd app wait rhesis \
            --server "$ARGOCD_SERVER" \
            --auth-token "$ARGOCD_AUTH_TOKEN" \
            --grpc-web \
            --health \
            --timeout 300
```

This makes the workflow fail if pods do not become healthy within 5 minutes, giving the on-call engineer a clear signal rather than a false green.

---

### DR note (informational, not a code change)

Because `argocd app set` patches live ArgoCD state and never commits back to Git (main is branch-protected), the deployed image tag is not persisted in the repo. In a full cluster rebuild scenario, ArgoCD would restore from Git with `imageTag: ""` and the deploy workflow would need to be re-run manually.

**Runbook for cluster DR:**
1. Rebuild the cluster and restore ArgoCD from Git as normal
2. Find the last deployed tag -- it is available in Artifact Registry (list tags by push date) or in ArgoCD application history before the cluster was lost
3. Re-run the `[Build & Deploy]` workflow with that tag as the override input -- no rebuild needed, the image already exists in the registry
4. ArgoCD syncs and the cluster is back to the last known good state

This is an accepted trade-off given the branch protection constraint. A future improvement would be a dedicated unprotected `gitops/` branch that CI commits tag updates to, so Git becomes the full source of truth and step 2-3 above become automatic.

PR #1878, opened 2026-06-03
Disable RLS during alembic migrations

EmanueleDeRossi1:
## Purpose
Alembic migrations crash on any database where RLS was enabled by https://github.com/rhesis-ai/rhesis/pull/1849 because the migration connection never sets `app.current_organization` — the session variable RLS policies depend on.

The error:

```
psycopg2.errors.UndefinedObject: unrecognized configuration parameter "app.current_organization"
```

## What Changed
Added `SET LOCAL row_security = off` at the start of the migration transaction in `env.py`. `SET LOCAL` scopes the change to the migration transaction only — no other connections are affected.

This is safe: whoever runs migrations already holds admin credentials with full database access; RLS provides no additional protection in that context.

## Testing
Deploy to an environment with RLS enabled and run `alembic upgrade head`.

### DIFF
--- a/apps/backend/src/rhesis/backend/alembic/env.py
+++ b/apps/backend/src/rhesis/backend/alembic/env.py
@@ -2,6 +2,7 @@ from logging.config import fileConfig
 
 from alembic import context
 from dotenv import load_dotenv
+import sqlalchemy as sa
 from sqlalchemy import create_engine
 
 from rhesis.backend.app.config.settings import get_database_settings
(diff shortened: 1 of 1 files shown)


harry-rhesis:
@EmanueleDeRossi1 thanks for this, but RLS should not be disabled. The permissions need to be set correctly for the admin user. Please consult @asadaaron.

EmanueleDeRossi1:
Thank you @harry-rhesis! Yes, and as peqy commented, this won't do anything: `SET LOCAL row_security = off` is superuser-only. On the other had, if you are super-user, you already bypass RLS anyways. Completely circular.

The real fix is granting permissions to the admin users, as you said. Closing this.

PR #1888, opened 2026-06-04
Fix sticky ActionBar and polish test generation wizard UI

nicolai-rhesis:
## Purpose
Fix the ActionBar not sticking to the bottom of the browser window on the `/test-sets/new-generated` wizard, and apply a set of related UI polish improvements to the same flow.

## What Changed
- **Fix sticky ActionBar**: Propagate a full-height flex chain through `AppShell`, `PageLayout` (new opt-in `fullHeight` prop), and all 3 wizard screens (`TestInputScreen`, `TestGenerationInterface`, `TestConfigurationConfirmation`). Each screen is now a flex column with a scrollable content region and the ActionBar pinned at the bottom via `position: relative` — no fragile `calc(100vh - 220px)` magic number
- **Fix over-rounded corners**: Skeleton loading cards and sample cards used `borderRadius: theme => theme.shape.borderRadius` which MUI's sx system multiplies (8 × 8px = 64px). Changed to `BORDER_RADIUS.sm` / `BORDER_RADIUS.xs` string constants that pass through as-is
- **Endpoint drawer**: Replace the centered Dialog with a `BaseDrawer` (slides in from the right), consistent with the rest of the app
- **EndpointsIcon**: Swap generic `ApiIcon` for the same `EndpointsIcon` SVG used in the sidebar navigation
- **Copy updates**: "Review Test Cases" → "Review Samples", "Get Endpoint Responses" / "Activate Live Responses" → "Show Live Responses"
- **Label consistency**: Scaffold category headings ("What to test", "Scenarios to cover", "Constraints & context") changed from `variant="caption"` to `variant="subtitle2"` to match the other labels in the card
- **Scaffold chips**: Clicking a chip now always appends to the end of the textarea instead of inserting at the cursor position; removes the cursor-tracking state and event handlers

## Testing
1. Navigate to `/test-sets/new-generated`
2. Confirm the Back/Continue bar is pinned to the bottom of the window even when the form content is shorter than the viewport
3. Resize the window short — content area should scroll, bar stays fixed
4. Walk through all 3 steps and verify the bar stays at the bottom on each
5. Click "Show Live Responses" — should open a right-side drawer, not a modal
6. Click scaffold chips — text should append to the end of the description textarea


harry-rhesis:
## Project boundary: generated test sets must be created with a required, explicit `project_id`

While reviewing the backend changes (the new pre-created "pending" test set flow), the project scoping for generated test sets is incorrect. Decisions for this PR:

- **The project must not be selectable in the generation wizard** — remove `ProjectSelector` from the flow.
- **Project assignment must be explicit / body-driven** — not inferred from the global active-project header.
- **A project is required** — generation must be rejected when no project is in context (no org-wide / `project_id = NULL` generated test sets).

### Root cause

The wizard has a `ProjectSelector` (`selectedProjectId`), but that selection is dropped on the final create call:

- `GenerateTestsRequest` has **no** `project_id` field, so the selected project can't ride in the body.
- The finish call builds its client with no project — `new ApiClientFactory(sessionToken)` → `getTestSetsClient()` → `new TestSetsClient(sessionToken, undefined, undefined)` — so `base-client` falls back to `readActiveProjectId()` (the **global active-project cookie**), not the wizard selection.
- Net effect: `POST /test_sets/generate` sends `X-Project-Id =` whatever the global active project is (or none) → `create_pending_test_set` flushes → `auto_stamp` stamps that project (or `NULL`). The pending test set ends up under the wrong project, or org-wide.

Note the streaming preview steps already pass `project_id` via `generateTestPipelineStream({ project_id: selectedProjectId })`, so only the final create is inconsistent.

### Required changes

**Backend**
- Add `project_id` (required `UUID4`) to `GenerateTestsRequest` (`apps/backend/.../schemas/services.py`).
- In `POST /test_sets/generate` (`routers/test_set.py`): validate `project_id` is present (return **400** if missing) and pass it explicitly into `create_pending_test_set`.
- `create_pending_test_set` (`services/test_set.py`): set `project_id` on the `TestSet` model **explicitly** instead of relying on the `auto_stamp` listener. Update the docstring to state the project is required.
- Forward `project_id` to the worker so the attach path (`_attach_tests_to_existing_test_set`) operates under the same project scope.

**Frontend**
- Remove `ProjectSelector` from the generation wizard (project is no longer user-selectable).
- Include `project_id` in the `GenerateTestsRequest` body sent by `testSetsClient.generateTestSet(...)` (`TestGenerationFlow.tsx`), sourced from the page/route project context rather than the global cookie.

### Secondary boundary risks (worth addressing here)

- `_mark_test_set_generation_failed` and `_attach_tests_to_existing_test_set` locate the row via the project-auto-filtered `db.query(TestSet).filter(TestSet.id == ...)`. If project scope ever fails to propagate to the worker, the filter collapses to `project_id IS NULL`, the row isn't found, **and** the failure marker also can't find it — leaving the test set stuck `in_progress` forever. The failure path should locate the row under `bypass_tenant_filter()` (it already knows the exact id + org) so it can always flip the status.
- No tests cover `create_pending_test_set` / `_attach_tests_to_existing_test_set` / `_mark_test_set_generation_failed`. Add a scope-boundary test asserting the generated test set lands under the required project.

nicolai-rhesis:
Backend schema (schemas/services.py): Added project_id: Optional[UUID4] to GenerateTestsRequest. It's optional at schema level because the same schema is shared with the synchronous sampling endpoint; enforcement is in the router.

Backend router (routers/test_set.py): POST /test_sets/generate now returns 400 immediately when project_id is absent. It passes project_id=str(request.project_id) explicitly to create_pending_test_set so the test set row is never implicitly scoped by the header alone.

Backend service (services/test_set.py): create_pending_test_set gains a required project_id parameter and stamps it directly onto the TestSet model constructor — bypassing auto_stamp entirely for this field.

Backend tasks (tasks/test_set.py): Both _attach_tests_to_existing_test_set and _mark_test_set_generation_failed now wrap the db.query(TestSet) lookup in bypass_tenant_filter(). This guarantees the row is found even when project scope fails to propagate to the Celery worker, eliminating the risk of a test set being stuck in_progress forever.

Frontend interface (interfaces/test-set.ts): Added project_id?: string to GenerateTestsRequest.

Frontend flow (TestGenerationFlow.tsx): handleGenerate now includes project_id: selectedProjectId in the request body, and fails fast with a user-facing error message if no project is active — giving a clear signal rather than a cryptic backend 400.

harry-rhesis:
Thanks for the quick turnaround, @nicolaibohn — this looks great. Confirmed the project boundary is now solid:

- `project_id` is explicit and body-driven (schema + router + `create_pending_test_set` stamps it directly on the model rather than relying on `auto_stamp`).
- `POST /test_sets/generate` returns **400** when no `project_id` is supplied, and the frontend guards before calling.
- `_attach_tests_to_existing_test_set` and `_mark_test_set_generation_failed` now wrap the lookup in `bypass_tenant_filter()`, so the "stuck `in_progress` forever" failure path is closed. Nice — that's the more robust of the two options I floated.

One small follow-up: **dead code**. With the project no longer user-selectable, `apps/frontend/src/app/(protected)/test-sets/new-generated/components/shared/ProjectSelector.tsx` is no longer rendered anywhere (no remaining `<ProjectSelector>` usages), and `setSelectedProjectId` is never called — `selectedProjectId` is fixed to `readActiveProjectId()`. Could you delete `ProjectSelector.tsx` and drop the unused setter so the intent ("project comes from the active-project context, not a picker") is clear in the code?

PR #1916, opened 2026-06-09
Fix: Self Hosted Runner Test

asad-miah:
This PR introduces changes from the `fix/self-hosted-runner-test` branch.

## 📁 Files Changed (      20 files)

```
.github/workflows/backend-k8s.yml
.github/workflows/chatbot-k8s.yml
.github/workflows/deploy-all-k8s.yml
.github/workflows/docs-k8s.yml
.github/workflows/frontend-k8s.yml
.github/workflows/k8s-deploy.yml
.github/workflows/otel-collector-k8s.yml
.github/workflows/polyphemus-k8s.yml
.github/workflows/telemetry-processor-k8s.yml
.github/workflows/worker-k8s.yml
charts/rhesis/templates/configmap.yaml
charts/rhesis/values-dev.yaml
charts/rhesis/values-prd.yaml
charts/rhesis/values-stg.yaml
charts/rhesis/values.yaml
infrastructure/config/service-secrets-config.sh.example
infrastructure/config/service-secrets-create.sh
kubernetes/clusters/dev/external-secrets/rhesis-app-secrets.yaml
kubernetes/clusters/prd/external-secrets/rhesis-app-secrets.yaml
kubernetes/clusters/stg/external-secrets/rhesis-app-secrets.yaml
```

## 📋 Commit Details

```
305cf3171 - feat(infra): add license and support email secrets (Md Asaduzzaman Miah, 2026-06-09 10:32)
c3517c9ed - feat(helm): add config vars and tune dev HPA (Md Asaduzzaman Miah, 2026-06-09 10:32)
1b3a6f518 - ci: fix deployment concurrency and tag handling (Md Asaduzzaman Miah, 2026-06-09 10:32)
```

## ✅ Checklist

- [ ] All tests pass locally


.github/workflows/k8s-deploy.yml
```diff
@@ -40,7 +40,16 @@ on:
 
 jobs:
   deploy:
-    runs-on: ubuntu-latest
+    # Self-hosted runner inside the target cluster's VPC so ArgoCD's internal
+    # hostname (e.g. dev-argocd.rhesis.ai) resolves via kube-dns. Build/push
+    # jobs stay on ubuntu-latest because Artifact Registry is public.
+    runs-on: ${{ inputs.environment == 'prd' && 'arc-runner-prd' || inputs.environment == 'stg' && 'arc-runner-stg' || 'arc-runner-dev' }}
```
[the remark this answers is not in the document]
asad-miah: not needed. the env will be selected from the dropdown, so it is safe. there is no chance of having different value.

kubernetes/clusters/dev/external-secrets/rhesis-app-secrets.yaml
```diff
@@ -318,3 +318,12 @@ spec:
         key: grafana-datasource-password
 
 ##--------------------------------------------------------
+## Rhesis license
+##--------------------------------------------------------
+    - secretKey: RHESIS_LICENSE_PRIVATE_KEY
+      remoteRef:
+        key: dev-rhesis-rhesis-license-private-key
```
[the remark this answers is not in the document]
asad-miah: managed by self-managed files to update the secret manager.

.github/workflows/k8s-deploy.yml
harry-rhesis, line 60: **[HIGH]** The nested ternary works, but an unrecognized environment value (e.g. a typo like `prod`) silently falls through to `arc-runner-dev`, deploying to the wrong cluster with no error.

Consider adding a validation step at the top of the job, or confirm that the callers' `validate-environment` action gates this path in all cases (including `deploy-all-k8s`, which does not call `validate-environment` today).

.github/workflows/k8s-deploy.yml
```diff
@@ -61,30 +73,87 @@ jobs:
...
+          # diff and nothing rolls. Instead we set every deploy-all service's tag
+          # explicitly, which overwrites those stale overrides and guarantees a rollout.
+          # chatbot is intentionally excluded (9GB image, built/deployed separately via
+          # chatbot-k8s.yml) and left on its own pinned tag.
+          SET_FLAGS=()
+          case "$SERVICE" in
+            "")
+              for KEY in backend.image.tag frontend.image.tag docs.image.tag \
+                         polyphemus.image.tag otelCollector.image.tag \
+                         telemetryProcessor.image.tag "workers[0].image.tag"; do
```
harry-rhesis, line 111: **[MEDIUM]** This service list also appears implicitly in the `needs` block of `deploy-all-k8s.yml`. When a new service is added, both must be updated in lockstep. Consider defining the list in one place (e.g. a job output or reusable composite action) to avoid drift.

.github/workflows/k8s-deploy.yml
```diff
@@ -61,30 +73,87 @@ jobs:
...
+          # Retry up to 5 times with 30s back-off to handle residual
+          # "another operation is already in progress" races not caught by
+          # the concurrency group (e.g. a manual sync in the ArgoCD UI).
+          for attempt in $(seq 1 5); do
+            argocd app sync "${{ inputs.app_name }}" \
+              --server "$ARGOCD_SERVER" \
+              --auth-token "$ARGOCD_AUTH_TOKEN" \
+              --grpc-web \
+              --timeout 300 \
+              ${RESOURCE_FLAG:+"$RESOURCE_FLAG"} && break
```
harry-rhesis: **[MEDIUM]** The retry catches all sync failures, not just the concurrent-operation race. A legitimate configuration error (bad tag, invalid Helm values) will retry 5 times with 30s delays (2.5 min wasted) before failing.

Consider inspecting stderr for `another operation is already in progress` and failing fast on other errors.

.github/workflows/k8s-deploy.yml
```diff
@@ -96,4 +165,5 @@ jobs:
             --auth-token "$ARGOCD_AUTH_TOKEN" \
             --grpc-web \
             --health \
-            --timeout 300
+            --timeout 1200 \
```
harry-rhesis, line 194: **[MEDIUM]** Bumped from 300s to 1200s (20 min). If a pod is crash-looping, CI hangs for 20 minutes before reporting failure. Is this driven by a specific slow-starting service (e.g. the worker with heavy ML imports)?

Consider making the timeout an input parameter with a sensible default, or add a comment explaining which service needs the extended window.

charts/rhesis/values-dev.yaml
```diff
@@ -190,8 +190,8 @@ workers:
         cpu: "1000m"
     hpa:
```
harry-rhesis, line 191: **[MEDIUM]** With `minReplicas: 1` and the note about heavy ML imports on startup, a single worker crash means zero capacity until HPA scales up and the replacement finishes its slow startup. Fine for dev cost savings, but worth confirming the readiness probe grace period accounts for this.

harry-rhesis (commented):
Thanks for the quality of this work, Asad. The comments are thorough (especially the `global.imageTag` shadowing analysis), the ArgoCD CLI is pinned by both version and SHA256, and the concurrency/retry choices are well-reasoned. A few items below.

## Summary

Three logical changes: (1) migrating deploys to self-hosted ARC runners with concurrency serialization and retry logic, (2) fixing the `deploy` input gate so reusable workflows correctly skip deploy when called with `deploy: false`, and (3) adding license/support-email secrets and config vars.

Overall quality is high. The inline comments below are mostly hardening suggestions, nothing blocking.

## Items not attached inline

**Stale header comment (`k8s-deploy.yml` lines 6-7):** The file header still says "leave `service` empty → sets global.imageTag", but the behavior now sets per-service tags explicitly. This contradicts the correct comment at `deploy-all-k8s.yml` line 126. Updating the header to say "sets per-service image tags for all deploy-all services (chatbot excluded)" would prevent confusion.

**Values formatting (`values.yaml`, `values-dev/stg/prd.yaml`):** The `nextPublicSupportEmail` lines removed the blank line that previously separated the Frontend and Backend config sections. Minor, but the blank line served as a visual separator in every per-env file.

**`nextPublicQuickStart` consistency:** This key is only in `values.yaml` (base defaults) and the configmap template, but not in any per-env file. Meanwhile `nextPublicSupportEmail` is set in every per-env file. If `nextPublicQuickStart` might need a per-env override (e.g. `"true"` in dev), worth adding it now.

## Positive Observations

- ArgoCD CLI pinned by version AND SHA256 checksum. Most teams only pin the version.
- The `deploy: false` gate fix is correct and well-commented. The explanation of `github.event_name` inheritance in reusable workflows is valuable.
- The `global.imageTag` shadowing analysis is excellent and prevents future "why didn't my deploy roll out?" incidents.
- Concurrency group with `cancel-in-progress: false` is the right call for deploys.
- Non-root ARC runner adaptation (`$HOME/bin` + `GITHUB_PATH`) done correctly.
- ExternalSecrets follow a consistent naming convention across all three environments.

.github/workflows/k8s-deploy.yml
asad-miah: with timeout 300, the all deployment fails after two or 3 service deployment. Backend and worker needs longer time to deploy.

charts/rhesis/values-dev.yaml
asad-miah: fine for dev and stg for cost saving.

.github/workflows/k8s-deploy.yml
asad-miah: GitHub Actions parses uses: at workflow load time, before any runtime context (including matrix) is available. The expression ${{ matrix.workflow }} in uses: is therefore always unrecognized, regardless of what the matrix contains. This limitation is not fixable — reusable workflow calls simply can't be dynamic.
improved the docs for the changes.

.github/workflows/k8s-deploy.yml
asad-miah: addressed in commit f55e011e0f255b00461b21f74654606c462c1857

.github/workflows/k8s-deploy.yml
harry-rhesis: The fast-fail fix in `ccd1d64` looked correct (capture output, grep for "another operation", fail immediately on anything else), but it was reverted in `3b7893f` with no explanation. The current retry loop still retries all sync failures indiscriminately, so a config error (bad tag, invalid Helm values) burns through 5 attempts with 30s delays before surfacing the real problem.

Could you share what went wrong with the reverted approach? Happy to help find an alternative if the output capture caused issues in practice.

asad-miah: It just exit with code 20 
https://github.com/rhesis-ai/rhesis/actions/runs/27271146557/job/80541458247

I have fixed it and reimplemented and tested it. 
The bug was a bash -e interaction with command substitution.

PR #1952, opened 2026-06-11
Fix traces page empty state and project-scoped fetch race

akwasigroch:
## Purpose

The Traces page could incorrectly show the empty state even when traces existed, and could fail to load traces on first visit because requests were sent before the active project was known.

## Problem

Two related bugs caused the Traces page to appear empty while the Network tab showed successful trace responses:

1. **Project scope race on first fetch** — Traces are project-scoped (fail-closed RLS). `TracesClient` fetched immediately on mount using only the session token. On first paint the active-project cookie/context may not be resolved yet, so `ApiClientFactory` was called without a project ID. That request returned no traces (or was rejected), and the UI treated the result as a real empty project.

2. **Empty hint unmounted the list** — `TracesClientWrapper` conditionally unmounted `TracesClient` when `showEmptyHint` was true. If the first fetch completed empty (from the race above), the empty-state card replaced the list component entirely. Later fetches that returned traces still succeeded in the Network tab, but the list was no longer mounted to render them — leaving the page stuck on the empty hint.

## What Changed

- Wait for `ActiveProjectContext` to finish loading before fetching traces; pass the resolved project ID into `ApiClientFactory`.
- Track `hasFetchedOnce` so the unfiltered empty hint only appears after a completed fetch with a known project.
- Keep `TracesClient` mounted in `TracesClientWrapper` and hide it visually while the empty hint is shown, so a subsequent successful fetch can clear the empty state.

## Additional Context

Hotfix for `release/v0.9.0`.

## Testing

- [ ] Open Traces on a project that has traces (hard refresh / first visit) — list should load instead of empty hint.
- [ ] Switch projects — traces should refresh for the selected project.
- [ ] Project with no traces — empty hint should still appear after fetch completes.
- [ ] Apply filters with no matches — filtered empty state should still work.

### DIFF
--- a/apps/frontend/src/app/(protected)/traces/components/SpanDetailsPanel.tsx
+++ b/apps/frontend/src/app/(protected)/traces/components/SpanDetailsPanel.tsx
@@ -15,13 +15,10 @@ import {
   TableBody,
   TableRow,
   TableCell,
-  Tabs,
-  Tab,
 } from '@mui/material';
 import ExpandMoreIcon from '@mui/icons-material/ExpandMore';
 import GridBadge from '@/components/common/GridBadge';
-import InfoOutlinedIcon from '@mui/icons-material/InfoOutlined';
-import AssessmentOutlinedIcon from '@mui/icons-material/AssessmentOutlined';
+import DetailTabNav from '@/components/common/DetailTabNav';
 import {
   SpanNode,
   TraceDetailResponse,
@@ -36,8 +33,6 @@ import { formatDuration } from '@/utils/format-duration';
 import TestResultTab from './TestResultTab';
 import TraceMetricsTab from './TraceMetricsTab';
 import TraceReviewsTab from './TraceReviewsTab';
-import RateReviewIcon from '@mui/icons-material/RateReview';
-import ForumOutlinedIcon from '@mui/icons-material/ForumOutlined';
 import { TasksAndCommentsWrapper } from '@/components/tasks/TasksAndCommentsWrapper';
 
 interface SpanDetailsPanelProps {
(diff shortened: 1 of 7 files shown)


nicolai-rhesis:
## Trace drawer UI follow-up (`34e901be8`)

Aligned the trace detail drawer with the Figma drawer shell pattern (Test Run drawer reference) and polished a few UX details from review:

### Drawer shell
- **85% width** (resizable from the left edge; default opened wider for the split-pane layout)
- **Footer-only Close** — removed the top-right X; dismiss via the outlined Close button at the bottom
- **`BaseDrawer` `contentLayout="fill"`** — full-height workspace layout so the split pane doesn’t fight the form-style inner scroll/gap

### Visual / layout
- Replaced the **“Trace Details” header + icon chips** with Figma-style **label + link pills** for project/endpoint/test run context, and **`GridBadge`** for spans/duration/environment
- Wrapped the split content in a **bordered section card** (radius, elevation, 30px padding)
- Swapped MUI icon tabs for **`DetailTabNav`** on both panels (18px bold, underline active state)
- **Tighter tab spacing** in the narrow split panels (`tabGap="20px"`, horizontal scroll instead of wrapping)
- Shortened left-panel labels to **Conversation · Tree · Sequence · Graph** (dropped redundant “View”)

### Tasks & Comments tab
- **30px gap** between Tasks and Comments cards (margin collapse was making borders touch)
- **30px padding** on the tab content in the span details panel

### Shared components
- **`DetailTabNav`**: optional `tabGap` and `scrollable` props (defaults unchanged for full-page detail views)

Happy to tweak spacing or width if anything still feels off in the split-pane layout.

PR #1955, opened 2026-06-11
Fix TraceDrawer project scope for deep-link fetches

nicolai-rhesis:
## Purpose
TraceDrawer fetches traces with `?project_id=` but without `X-Project-Id`. Backend RLS fail-closes project-scoped rows when the active-project cookie is not set yet (deep links, first paint).

## What Changed
- Pass `projectId` to `ApiClientFactory` in `fetchTrace` and `refreshTrace`
- Apply the same scope to the test-run experiment metadata fetch

## Additional Context
- Follow-up to #1952 review feedback (merged before this fix landed)

## Testing
- Open a trace via deep link (`?open_trace=...&project_id=...`) before the project switcher resolves
- Confirm trace details load instead of 404/empty state
- Refresh trace in drawer and verify span data updates

### DIFF
--- a/apps/frontend/src/app/(protected)/traces/components/TraceDrawer.tsx
+++ b/apps/frontend/src/app/(protected)/traces/components/TraceDrawer.tsx
@@ -115,7 +115,7 @@ export default function TraceDrawer({
     setSelectedSpan(null);
 
     try {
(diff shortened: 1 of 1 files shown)


akwasigroch:
Closing, the fix was applied in the release branch

PR #2009, opened 2026-06-19
Add Haystack integration (auto-instrumentation + Penelope Target)

### LINKED ISSUE #1869: Add Haystack integration (auto-instrumentation + Penelope Target)
## Summary

Add full Haystack integration: automatic telemetry instrumentation of Haystack pipelines via pipeline hooks and a `HaystackTarget` for Penelope conversation testing of NLP pipelines.

## Problem / Opportunity

Haystack is widely used for building NLP and RAG pipelines. Users need automatic observability and a way to run Penelope conversation tests against their Haystack pipelines.

Part of container issue: #1083

### Auto-instrumentation

Implement `HaystackIntegration` in `sdk/src/rhesis/sdk/telemetry/integrations/haystack.py` using Haystack's pipeline event hooks (e.g. `ComponentBase` tracing or `Pipeline.run` instrumentation) to automatically capture:
- Pipeline input/output content
- Token usage from component outputs
- Model name and provider detection
- Latency and errors per pipeline run

Register in `sdk/src/rhesis/sdk/telemetry/integrations/__init__.py`.

### Penelope Target

Implement `HaystackTarget` in `penelope/src/rhesis/penelope/targets/haystack.py` wrapping a Haystack `Pipeline` so Penelope can run multi-turn conversation simulations against Haystack NLP pipelines.

Export from `penelope/src/rhesis/penelope/targets/__init__.py`.

## Acceptance Criteria

- [ ] `HaystackIntegration` auto-instruments Haystack pipeline runs
- [ ] Token usage, model, latency captured automatically
- [ ] `HaystackTarget` wraps Haystack `Pipeline` for Penelope testing
- [ ] Multi-turn conversations work via conversation history management
- [ ] Unit tests for both integration and target
- [ ] Existing LangChain/LangGraph tests still pass

Ayush7614:
## Purpose

Add Haystack pipeline auto-instrumentation and a Penelope target for multi-turn NLP/RAG pipeline testing, aligned with the same `ai.*` schema standard used by LangChain, LangGraph, and MAF integrations.

Closes #1869

### SDK telemetry (`ai.*` schema conformance)
- Added `HaystackTranslatingExporter` (MAF-style) to rewrite native Haystack OTel spans into Rhesis `ai.*` before export
- Added `haystack/mapping.py` for span name, attribute, and event translation (pipeline → `ai.agent.invoke`, components → `ai.llm.invoke`, `ai.retrieval`, `ai.tool.invoke`, etc.)
- Enabled Haystack content tracing by default so prompts, completions, and tool I/O are captured
- Native tracing + translation is the default path; `Pipeline.run` patching remains as fallback only
- Refactored integration into `integrations/haystack/` package (`integration.py`, `mapping.py`, `translator.py`)
- Hardened `tracing_helpers.py` with `sanitize_for_tracing()` for binary/sensitive/signed-URL data

### Penelope target
- Added `input_mapping` and `run_inputs` for nested `{component: {socket: value}}` pipeline inputs
- Added native async path via `a_send_message` → `pipeline.run_async`
- Documented that `files` attachments are not supported (explicit error)

### Examples
- `examples/telemetry/haystack_example.py` — transparent `auto_instrument("haystack")` flow
- `penelope/examples/haystack_minimal.py` — `PenelopeAgent.execute_test()` demo

### Tests
- Rewrote integration tests to exercise the translating exporter and assert `ai.*` attributes (no mocking of the primary tracing path for schema checks)

## Testing

```bash
cd sdk
uv run pytest ../tests/sdk/telemetry/integrations/test_haystack_integration.py -v

cd penelope
uv run pytest ../tests/penelope/targets/test_haystack.py -v
```

All 17 Haystack-related tests passing.


Ayush7614:
cc: @harry-rhesis

Ayush7614:
cc: @Arman-Beykmohammadi

Arman-Beykmohammadi:
Hi @Ayush7614,

I went back through this after the rework, and honestly the turnaround is great. The translating exporter, the content sanitizer, the nested `input_mapping` on the target, the native async path, and the examples all landed cleanly, and the tests now check real `ai.*` output instead of mocking it. Nice work taking the earlier feedback and running with it.

One thing worth knowing: while this PR was open, `main` moved. We now have a shared `genai.py` module with the common translator machinery (`TranslatedSpan`, `translate_events`, `get/set_processor_exporter`, `content_capture_enabled`), a Pydantic AI integration already built on it, and a first-party Haystack agent under `agents/dr-rhesis` whose playground is stubbed with a comment saying it's waiting on Haystack `auto_instrument`. So the bar for "consistent with the rest of the codebase" shifted a bit under you. Good news is the branch is even with the main tip and the merge is clean, so none of this is a rebase headache.

Here's what I'd like to sort out before we merge, roughly in priority order:

**1. Build the translator on `genai.py` rather than private copies.** Right now the Haystack translator and exporter-swap logic are a third hand-rolled copy of what MAF and Pydantic AI already do, and the drift is real, not hypothetical: the Haystack version reads the exporter only via `span_exporter` and wraps only `BatchSpanProcessor`, so on a `SimpleSpanProcessor` setup (common in local/dev) it fails closed while the other two keep working. Same `auto_instrument("haystack")` call, different result depending on the framework. Reusing the shared helpers fixes that and means the next exporter fix lands once instead of three times. Since the shared module is already on the branch from the merge, this is mostly deleting code.

**2. Honor `RHESIS_DISABLE_CONTENT_CAPTURE` and stop writing to `os.environ`.** MAF, Pydantic AI, and the shared genai path all gate prompt/completion capture on `genai.content_capture_enabled()`. This integration force-enables it and also sets `HAYSTACK_CONTENT_TRACING_ENABLED=true` in the process env, which outlives `disable()` and leaks into anything else the app is tracing. For anyone who turned that flag off for privacy reasons, we'd be quietly overriding them. Routing through the shared check keeps the opt-out working.

**3. Add `haystack` to the blocked set in `tests/penelope/targets/test_optional_imports.py`.** That test is what actually guarantees Penelope imports cleanly without the optional extras installed. Haystack isn't in the list yet, so the "works without `haystack-ai`" promise isn't pinned and could regress silently. One-line change.

**4. Fresh CI + a human pass on the two new commits.** The earlier peqy approval is on a commit this branch no longer resembles, and the translator plus the ~400-line `mapping.py` (the code most likely to hide edge cases) hasn't been looked at by anything since. Worth a real review before this goes in. Merging while the branch is even with main also saves us another drift-and-remerge round.

**5. Wire up `agents/dr-rhesis` as the first real consumer** (here or in an immediate follow-up). Pointing that agent at `auto_instrument("haystack")` and unstubbing its playground gives us an actual in-repo pipeline exercising the mapping end to end, which will catch component-classification gaps that mocked unit tests won't.

A few minor things, no need to block on them: `set_token_attributes` in `tracing_helpers.py` is defined but never called now (the mapping uses `extract_token_usage` directly), the `Secret.from_env_var(...) or Secret.from_token(...)` line in the telemetry example never actually falls back since `from_env_var` always returns a Secret, and rejecting `files` on the target diverges from the `files_to_content_blocks` convention added in #2115. The last one is a reasonable call for a pipeline target, just add a line to the docstring noting it's deliberate.

Thanks again for pushing this so far. Points 1–3 are small and mechanical given what's already on the branch, and once they're in I think this is in good shape to land. Happy to pair on the genai.py refactor if it helps.

Ayush7614:
Thanks @Arman-Beykmohammadi for the thorough re-review — and @peqy for the re-audit. All points are now addressed on `feature/haystack-integration` (latest commit `ed2d970f7`). Summary:

**Arman's points 1–3 (priority):**
1. **Built on `genai.py`.** The Haystack translator and exporter-swap no longer hand-roll the shared machinery. `integration.py` wraps both `BatchSpanProcessor` *and* `SimpleSpanProcessor` via `set_processor_exporter`/`get_processor_exporter`, and `translator.py` reuses `genai.TranslatedSpan` (the private `_TranslatedSpan` was dropped). This fixes the `SimpleSpanProcessor` fail-closed drift between integrations.
2. **Honor `RHESIS_DISABLE_CONTENT_CAPTURE`, no `os.environ` writes.** Content tracing is gated on `content_capture_enabled()` (mirrors MAF/Pydantic AI). I also removed a stray `os.environ.setdefault("HAYSTACK_CONTENT_TRACING_ENABLED", "true")` that had crept into `agents/dr-rhesis/app.py` — it would have reintroduced exactly the process-global leak you flagged. No env writes remain in the integration path.
3. **`haystack` added to blocked imports.** `tests/penelope/targets/test_optional_imports.py` now blocks `haystack`, so the "works without `haystack-ai`" guarantee is pinned.

**Point 4 (fresh CI + human pass):** CI is green on the current tip and the translator/`mapping.py` have a real test pass behind them (real `ai.*` output assertions, not mocks). This thread is the human pass — happy to pair on anything that still looks off.

**Point 5 (wire up `dr-rhesis`):** Done in `agents/dr-rhesis` — `app.py` calls `auto_instrument("haystack")` after `RhesisClient` is constructed, the playground example is unstubbed, and the architecture docs are updated.

**Minor items:**
- `set_token_attributes` in `tracing_helpers.py` was removed (the mapping uses `extract_token_usage` directly).
- Fixed the dead `Secret.from_env_var(...) or Secret.from_token(...)` fallback in the telemetry example (the `from_env_var` is always truthy, so it never fell back) — now falls back only when the resolved value is `None`.
- Added a docstring line on the target noting `files` is deliberately rejected (diverges from `files_to_content_blocks` by design, for a pipeline target).
- **peqy follow-ups also addressed:** synthesized events now omit `timestamp=0` to match `genai.translate_events()`; `ai.agent.name` is reserved for pipeline/agent spans instead of being overloaded onto every component span; and `enable()` now reverts Haystack tracing/content capture on the early-return (fail-closed) paths.

Left as follow-up (not blocking, out of scope for this PR): the `sanitize_for_tracing` recursion-depth guard and the deeper `ai.agent.name`-vs-component-attribute schema question are tracked separately.

Ayush7614:
Thanks @peqy — addressed the final follow-up in commit \`40a781740\`:

- Hardened \`sanitize_for_tracing()\` with a recursion **depth limit** (\`MAX_SANITIZE_DEPTH = 32\`); deeply nested lists/dicts now return \`[max depth exceeded]\` instead of risking stack overflow.
- Added **cycle detection** via a \`seen\` id-set; reference cycles in dicts return \`[circular reference omitted]\` (ids are tracked in a try/finally so the set is correctly unwound during recursion).
- Public signature is unchanged (public wrapper delegates to a private \`_sanitize\` worker that carries \`depth\`/\`seen\`), so all callers — including \`truncate_content\`/\`add_agent_io_events\` used by LangChain/LangGraph/MAF/Haystack — are unaffected.

Verified: \`ruff\` clean and the full \`tests/sdk/telemetry/\` suite passes (383 passed, 1 skipped). This was the only outstanding (non-blocking) item, so the review follow-ups should now be fully cleared.

Arman-Beykmohammadi:
@Ayush7614 Thank you so much for the work. It looks very promising.

Just one last comment, can you please also make sure the lint checks will pass properly? It's a requirement for PRs to be merged, to have all the checks pass.

best,
Arman

Ayush7614:
cc: @Arman-Beykmohammadi

Arman-Beykmohammadi:
Thanks for putting this together @Ayush7614, and sorry for the slow turnaround. There's a lot of solid mechanics in here (patch/unpatch state handling, the fallback logic, the disable path), but I can't merge it as it stands: the branch is ~2 months behind `main` and in that time the way we do Haystack, and the way we do framework integrations in general, both changed. Most of what follows is context you couldn't have had when you opened this, so please read the first two sections before touching any code.

---

## 1. The direction for Haystack changed, and it's no longer an SDK-side integration

Since this PR was opened, Haystack support in Rhesis moved out of the SDK and into a **Haystack connector package**, `rhesis-haystack`, which lives in the deepset `haystack-core-integrations` monorepo (currently in review as deepset-ai/haystack-core-integrations#3669). It follows the `LangfuseConnector` pattern that Haystack itself endorses, i.e. a tracing component you wire into the pipeline rather than something that monkey-patches `Pipeline.run` from the outside.

You can see the intended usage on `main` in `agents/visit-prep`:

```python
# agents/visit-prep/examples/run_scenarios_traced.py
os.environ.setdefault("HAYSTACK_CONTENT_TRACING_ENABLED", "true")  # before any haystack import
from haystack_integrations.tracing.rhesis import RhesisTracing

tracing = RhesisTracing("Visit-Prep", turn_span_name="function.visit_prep_turn")
tracing.start_conversation(str(uuid.uuid4()))
...
tracing.flush()
```

and in `agents/visit-prep/pyproject.toml`, which depends on `rhesis-haystack` and pins `haystack-ai>=3.0,<4`.

So `sdk/src/rhesis/sdk/telemetry/integrations/haystack.py` as written would be a second, competing Haystack tracing path that fights the connector for the global Haystack tracer (`haystack.tracing.enable_tracing()` is process-global and last-writer-wins). **I'd like to drop the SDK-side `HaystackIntegration` from this PR entirely** and keep the scope to the Penelope target, which is genuinely still missing and still wanted. More on the reshaped scope in section 7.

> Heads up on one specific thing: **`agents/visit-prep` on your branch is the old version.** The one on `main` is the one we're moving forward with (Haystack 3.x multi-agent, wired to `rhesis-haystack`). When you rebase, take `main`'s copy wholesale and don't resolve any conflict in favour of your branch's version. It's also the best reference for what a real Haystack pipeline input/output actually looks like, which matters a lot for section 5.

## 2. The branch is stale in a way that would silently delete newer code

`sdk/src/rhesis/sdk/telemetry/integrations/__init__.py` in this diff only knows about `langchain`, `langgraph` and `autogen`. On `main` it also registers `agent_framework` (plus the `"maf"` alias) and `pydantic_ai`. Merging this file as-is would drop both from `get_all_integrations()`, which would break MAF and Pydantic AI auto-instrumentation with no test failure to catch it, since the registry isn't asserted against anywhere.

Same story for `sdk/pyproject.toml`: the extras block here predates the ones on `main`.

Please `git rebase origin/main` (not merge) and re-verify every touched file against current `main` before pushing again.

## 3. The native-tracer path doesn't work on the Haystack version we target

This is the bug that worries me most, because it fails quietly.

```python
# sdk/src/rhesis/sdk/telemetry/integrations/haystack.py:95-101
def _enable_haystack_tracing() -> Any:
    from haystack import tracing
    from haystack.tracing import OpenTelemetryTracer
```

`OpenTelemetryTracer` was removed from `haystack.tracing` in Haystack 3.x. `haystack/tracing/opentelemetry.py` no longer exists in core; it moved out to the separate `opentelemetry-haystack` integration package. On 2.x that import works, on 3.x it raises `ImportError`.

Combined with `_create_callback()`:

```python
# haystack.py:119-126
try:
    return _enable_haystack_tracing()
except Exception as exc:
    logger.debug(f"Haystack tracing API unavailable, using pipeline patch: {exc}")
    _patch_pipeline_run()
    return "haystack_patched"
```

on Haystack 3.x (i.e. every user we actually have, per `visit-prep`) this always falls through to the monkey-patch, and it does so at `debug` level, so nobody finds out. `enable()` still returns `True` and logs `✓ Observing haystack`. A silent degradation to a strictly worse instrumentation mode should at minimum be a `warning`, and arguably should fail closed. Compare `PydanticAIIntegration.enable()` on `main`, which returns `False` rather than emitting spans it knows the backend can't consume.

Related version-matrix issue: `hasattr(Pipeline, "run_async")` is only true on Haystack 3.x. On 2.x the async entry point is the separate `AsyncPipeline` class, which this never patches. Since we've standardised on `>=3.0,<4`, the `>=2.0.0` pins in both `pyproject.toml` files are wrong and the 2.x compatibility branches are dead weight.

## 4. Span names and attributes would be rejected by the backend

```python
# haystack.py:49-53
with observe_framework_call(
    f"haystack.pipeline.run {pipeline_name}",
    ...
```

`sdk/src/rhesis/sdk/telemetry/attributes.py::validate_span_name` documents what the backend enforces: `ai.<domain>[.<action>]` or `function.<name>`, all lowercase, no spaces, and framework concepts like `pipeline` are in `FORBIDDEN_SPAN_DOMAINS`. `haystack.pipeline.run qa-pipeline` fails on the prefix, on the domain and on the space, and the backend answers 422. So even where the patch path does run, the spans don't land.

This is exactly the problem `sdk/src/rhesis/sdk/telemetry/integrations/genai.py` exists to solve on `main`: frameworks emit their own span shape, and we translate to the `ai.*` / `function.*` schema in a translating exporter before export. `agent_framework/` and `pydantic_ai/` are both structured as `integration.py` + `mapping.py` + `translator.py` around it. Any future framework work should reuse that, not invent a parallel path.

A few more things in that area:

- **No content-capture opt-out.** `add_agent_io_events(span, data, None)` stringifies the entire pipeline input. For a RAG pipeline that's documents, embeddings and whatever PII the user passed in, all onto a span attribute. `genai.py` exposes `content_capture_enabled()` / `RHESIS_DISABLE_CONTENT_CAPTURE` for precisely this and every integration on `main` honours it.
- **`HAYSTACK_CONTENT_TRACING_ENABLED` is never mentioned.** Haystack reads it once, at `ProxyTracer` construction time, i.e. at import. Without it set before `import haystack`, the native path produces spans with no input/output content at all. This has to be documented wherever we tell people to turn Haystack tracing on.
- **`disable()` doesn't undo the native path.** `HaystackPatchState.reset()` only unpatches `Pipeline.run`. If `_enable_haystack_tracing()` succeeded, the tracer stays installed globally after `disable()`. It needs a `haystack.tracing.disable_tracing()`.
- **A single flat pipeline span is not much of an integration.** One `transform` span per `Pipeline.run` gives no per-component visibility, no retrieval spans, no token counts, no cost. The value is in the component-level spans, which is what the connector route gets us.
- Minor: `getattr(self, "metadata", {}).get("name")` blows up if `metadata` is `None`; use `(getattr(self, "metadata", None) or {})`. And the wrappers should use `functools.wraps`.

## 5. `tracing_helpers.py` overlaps with what already exists

The module is reasonable in isolation but it's a second home for things that already have one:

- `MAX_CONTENT_LENGTH = 4000` here vs `MAX_CONTENT_LENGTH = 8000` in `telemetry/attributes.py`, which is explicitly commented as framework-agnostic and shared.
- `set_agent_attributes`, `set_token_attributes` and `infer_model_provider` aren't called by anything in this PR. `infer_model_provider`'s substring heuristic will also silently mislabel most non-OpenAI/Anthropic/Google models.
- `observe_framework_call` sets `"ai.operation.duration_ms"` as a bare string literal. It isn't in `AIAttributes`, and OTel spans already carry start/end timestamps, so downstream this is redundant and off-schema.

If a shared helper module is still needed after the rescope, it should live next to `genai.py` and reuse its constants rather than restate them.

## 6. `HaystackTarget` needs to work against a real pipeline

This part I do want, but the input and output handling is modelled on a pipeline shape that Haystack doesn't really produce.

**Input.** `Pipeline.run` takes `{component_name: {socket_name: value}}`. The flat form only resolves when the socket name is unambiguous across the whole graph. `{self.input_key: message}` therefore breaks on anything with more than a trivial component set, including our own reference agent:

```python
# agents/visit-prep/src/visit_prep/pipeline.py
return {COORDINATOR: {"messages": [...], "slots": ..., "chief_complaint": ..., "slot_status": ...}}
```

Please add a `component` argument (nest under it when set) and/or an `input_builder: Callable[[str, list], dict]` escape hatch for pipelines that need more than one field per turn.

**Output.** Results are nested by component name too, so the top-level probe over `["answer", "reply", "result", "output", "generator", "llm"]` almost never hits and `_extract_output` falls through to `return str(result)`. Penelope then treats a Python dict repr as the agent's reply and every evaluation downstream is garbage. It needs to descend one level into component outputs, and handle the common Haystack shapes explicitly: `{"replies": [ChatMessage, ...]}` and `{"answers": [GeneratedAnswer, ...]}`, taking `.text` off the last `ChatMessage`. Right now the `replies` branch does `str(nested)` on a list and yields `[ChatMessage(...)]`.

**Other things in `penelope/src/rhesis/penelope/targets/haystack.py`:**

- No `a_send_message` override, so async runs go through the base `asyncio.to_thread` fallback and block a worker thread even though Haystack 3.x has `Pipeline.run_async`. `PydanticAITarget` on `main` is the pattern to copy.
- `files` is accepted and silently dropped. Either handle it (see `_file_compat` usage in `targets/pydantic_ai.py`) or return an explicit `TargetResponse(success=False, ...)` so it fails loudly.
- History is corrupted on failure: the user turn is appended before `pipeline.run`, and if the run raises it stays in `_session_histories`. Buffer it and only commit both turns after a successful run.
- History is `list[dict[str, str]]`, but Haystack agents want `list[ChatMessage]`. Worth either converting or documenting that `history_key` is only for pipelines taking plain dicts.
- `metadata["raw_response"]` stores the whole result dict. For a RAG pipeline that can be megabytes of `Document` objects with embeddings, and it goes into Penelope's turn metadata. Truncate it the way the other targets do.
- Once the target lands, add it to the docstring list and `__all__` in `penelope/src/rhesis/penelope/targets/__init__.py` (which has also changed on `main`, see section 2).

## 7. Tests

The tests currently lock in the behaviour that needs to change, which is worth fixing before anything else.

- `test_send_message_success` asserts `run({"query": "What is NLP?"})`, i.e. the flat input shape from section 6.
- The SDK tests inject `MagicMock`s into `sys.modules["haystack"]`, and `test_enable_patches_pipeline_run` explicitly patches `_enable_haystack_tracing` to raise, so no test ever exercises the real Haystack path. That's how the 3.x `ImportError` slipped through.
- Nothing asserts what actually gets emitted. For an instrumentation change the core test is: run something, collect spans with `InMemorySpanExporter`, assert names and `ai.*` attributes. The MAF and Pydantic AI translator tests on `main` do this and are the model to follow.
- No coverage for `run_async`, for `disable()` leaving the global tracer installed, or for content capture being off.
- The PR description's test command runs `test_langgraph_integration.py`, which isn't related to this change.

## 8. Suggested path forward

Concretely, what I'd like the next revision to look like:

1. `git rebase origin/main`. Take `main`'s `agents/visit-prep`, `sdk/src/rhesis/sdk/telemetry/integrations/__init__.py`, `penelope/src/rhesis/penelope/targets/__init__.py` and both `pyproject.toml` files as the base.
2. Drop `sdk/src/rhesis/sdk/telemetry/integrations/haystack.py` and `sdk/src/rhesis/sdk/telemetry/integrations/tracing_helpers.py`, and the SDK-side registration and extras. Haystack tracing is handled by `rhesis-haystack`.
3. Keep and rework `HaystackTarget` per section 6: nested-input support, real output extraction, `a_send_message` via `run_async`, explicit `files` behaviour, transactional history, bounded metadata.
4. Pin `haystack-ai>=3.0,<4` in `penelope/pyproject.toml` to match `visit-prep`.
5. Rewrite the Penelope tests against a small real `Pipeline` (a trivial two-component graph is enough, no API key needed) rather than a bare `MagicMock`, and add one integration-style test that runs `visit-prep`'s coordinator pipeline shape through the target.
6. Update the PR title and description, since the scope changes. Conventional Commits, per `CONTRIBUTING.md`.

If you'd rather split it, a Penelope-target-only PR is the fastest thing to land and I'm happy to review that quickly. Then we can talk separately about whether anything from the tracing side is still worth carrying over.

Thanks again for the work here, and sorry the ground moved under it.

PR #2032, opened 2026-06-25
feat(frontend): RBAC consumption mechanism + server-driven affordances

harry-rhesis:
## Purpose
First of three PRs delivering the RBAC frontend (see roadmap). This is the **consumption mechanism**: the reusable machinery for rendering authorization decisions, plus **comments** wired end-to-end as the reference. The backend (merged) is the single source of truth and ships dark behind `FeatureName.RBAC`; this layer is strictly cosmetic and fail-closed.

**No behavior change when RBAC is off** — the default license returns full capabilities, so every affordance renders exactly as today.

## What Changed
**One abstraction.** Every check is `can(subject, capability)` over a subject's server-resolved `permitted_actions`, subsuming object-level and scope/role-level:
- **ambient/scope subject** — the active project, via `GET /me/permissions` (`PermissionsContext` + `useCan`) → nav/page/create gating.
- **object subject** — a resource instance carrying `permitted_actions` (`can` + `<Can>`) → per-row edit/delete, with `:own` resolved server-side.

**Backend — server-driven affordances**
- `permitted_actions_for`: generic resolver projecting a caller's effective caps over an object's owner (no per-resource registry — the capability string is the schema).
- `effective_permissions`: single implementation behind both `/me/permissions` and the resolver.
- `populate_permitted_actions` + `WithPermittedActions` mixin; **comment** wired on every returning handler (incl. mutations the frontend patches state from). `ResourceType` enum (no resource-name literals).

**Frontend — rendering machinery**
- `PermissionsContext`/`useAmbientPermissions`, `can`/`useCan`/`<Can>`, `permissions-client`, `FeatureName.RBAC`.
- `EntityAction`/`EntityActionBar` descriptor layer — action sets as data, gated + rendered generically (no per-component `canEdit`/`canDelete` booleans).
- Constants (no magic strings): `Capability` mirror, canonical `EntityType` module (consolidates 3 prior copies), labels/area/level helpers.
- Permission-gated nav (`requiredPermission` + locked `NavItem` state); `AccessDenied`; removed dead `is_superuser`.
- `CommentItem` migrated to the descriptor layer as the reference consumer.

## Additional Context
- Roadmap: PR2 = consumption rollout across entities; PR3 = EE role-authoring UI. This PR is the foundation both build on.
- `EntityType` is a `const`-object + union (not a TS `enum`) to keep ~90 existing string-literal call sites compiling; migrating those is scoped to PR2.

## Testing
- Backend: `pytest` — resolver unit tests (owner/non-owner/admin/cross-resource); comment-route + capabilities suites green; `ruff` clean.
- Frontend: `tsc` clean; `CommentItem` (visibility driven by `permitted_actions`) and `entity-actions` resolver suites pass.


harry-rhesis:
Thanks @peqy — all three addressed in `936ff9e` (`fix(rbac): owner-gate affordances to match endpoint enforcement`).

**Critical — affordance/enforcement mismatch.** Fixed on the *affordance* side rather than the endpoint. The endpoint's owner-only `authorize_object` (no admin bypass) is intentional: in the community tier every project member holds the plain `comment:update`/`delete` cap, so the suggested `authorize(...) OR authorize_object(...)` would let any member edit any comment. Instead, `permitted_actions_for` now **owner-gates** any action that has a `:own` variant in the catalog (derived once by `populate_permitted_actions`): the action is advertised only when the caller owns the object **and** holds the `:own` cap; the plain cap alone never advertises it. So the affordance now exactly matches what the endpoint enforces. Added a regression test (`test_global_cap_does_not_grant_owner_gated_action`).

If we *do* want admins to moderate any comment, that's a deliberate backend policy change (endpoint + community cap grants) — tracked for the rollout PR, not silently flipped here.

**Improvement — docstring shape.** Corrected the `WithPermittedActions` docstrings (backend `schemas/affordances.py` + frontend `types/affordances.ts`) to say `permitted_actions` holds full capability strings (e.g. `comment:update`), matching the implementation.

**Improvement — logout state.** `PermissionsContext` now resets to the fail-closed default when `status !== 'authenticated'`, so permissions don't leak across an auth transition.

Verified: backend 61 passed (incl. the new regression), `ruff` clean; frontend `tsc` clean, `CommentItem` suite green.

harry-rhesis:
Addressed both follow-ups in `7f36168`:

**Fail-open during feature-flag load.** `PermissionsProvider` now reads
`useFeaturesState().loading`: while flags are loading the RBAC status is unknown,
so it stays `loading` (not `DISABLED_STATE`), and `useCan` checks `loading` before
`enabled` → fail-closed until status resolves. EE nav no longer flashes
unlocked→locked; community still renders permissively once flags resolve to off.
Added a test (`is fail-closed while RBAC status is unknown`).

**`own_gated_actions` footgun.** Made it a required parameter of
`permitted_actions_for` (no `frozenset()` default), so a future caller can't omit
it and over-advertise owner-gated actions. Both call sites already pass it.

Verified: backend resolver 11 passed, ruff clean; frontend tsc/eslint/prettier
clean, Can suite 7 passed.

harry-rhesis:
Applied the polish in `1dc21b6`: while `FeaturesContext`/the scope set is still loading, `requiredPermission` nav items now render nothing (return null) instead of flashing locked. The lock/show decision is made only once it's definitive, so community no longer flashes locked and EE never fails open. Steady-state nav is unchanged. tsc/eslint/prettier clean.

PR #2035, opened 2026-06-25
Add Alembic migration head conflict check

EmanueleDeRossi1:
## Purpose
Prevent the common scenario where two developers branch off the same Alembic head and both write a migration with the same `down_revision`, causing a conflict on main.

## What Changed
- Added `.github/workflows/check-alembic-head.yml`: CI check that runs when a PR adds files under `alembic/versions/`

## How it works

```
1. Find new migration files added in this PR
   git diff(base...HEAD) → list added files in alembic/versions/
   if none → exit 0 (nothing to check)

2. Read base branch head(s)
   git worktree add(base_ref) → isolated checkout of main (or PR target)
   alembic heads in worktree → BASE_HEADS (usually one revision)

3. Read PR branch head(s)
   alembic heads in PR checkout → PR_HEADS

4. If PR has multiple heads
   for each new head (not already on base):
     if down_revision ≠ base head → fail with "rebase and fix down_revision"
   fail with "multiple heads — chain linearly or add merge migration"

5. Walk the chain from PR head back toward base
   current = PR_HEAD
   while current's parent(s) are not the base head(s):
     if not a single parent → fail (need merge migration when base has multiple heads)
     current = parent
     chain_len++

6. Verify chain length matches new file count
   if chain_len ≠ number of added files → fail
     (catches orphaned migrations not connected to the new chain)

7. Pass
   all new migrations form one chain landing on base head(s)
```

**Example — conflict (should fail):**
```
main:     ... → A (head)
PR adds:  ... → A → B
          ... → A → C   ← two migrations, same down_revision A → two heads
```

**Example — valid (should pass):**
```
main:     ... → A (head)
PR adds:  ... → A → B → C   ← single linear chain, one new head
```

## Testing
This PR intentionally includes two fake conflicting migrations (`zzz_fake_conflict_migration.py` and `yyy_another_fake_conflict_migration.py`) that both point at main's current head. The check should fail with a clear error. Once confirmed red, a follow-up commit removes them and the check should go green.


EmanueleDeRossi1:
Regarding peqy's comments: 

- `actions/checkout@v6`: this is correct and exists, the most recent version is actually v7. 

- "alembic_heads() still does output=$("$ALEMBIC" heads 2>&1) and then awk '{print $1}'" This is correct, but not so urgent. The suggested fix is wrong though, `-q` does not exist for `alembic heads`.

PR #2054, opened 2026-06-26
Owasp synthesizer

alexeyhramov123:
# Purpose
Extends the OWASP synthesizer with source-grounded generation. The new model we're running on Polyphemus required updating the vLLM deployment along the way.

# What Changed
- **OWASPSynthesizer**:
generation is now driven by the actual OWASP PDF content via a new owasp_extractor service — no hardcoded category descriptions
- **Polyphemus**:
updated vLLM Docker image to 20260622_0916_RC01 and migrated --guided-decoding-backend → --structured-outputs-config.backend (removed in vLLM v0.12). Model is configured via POLYPHEMUS_MODEL_PATH / POLYPHEMUS_DEFAULT_MODEL secrets as before
- **Example**:
added examples/owasp-synthesizer.ipynb covering LLM Top 10, Agentic Top 10, and filtered section usage

### DIFF
--- a/apps/polyphemus/model_deployment/deploy.py
+++ b/apps/polyphemus/model_deployment/deploy.py
@@ -65,7 +65,7 @@ def deploy_model_vllm(
         model_config=model_config,
         enable_lora=enable_lora,
         enforce_eager=enforce_eager,
-        guided_decoding_backend=guided_decoding_backend,
+        structured_outputs_backend=guided_decoding_backend,
     )
 
     logger.info(f"Uploading model: {model_config['model_name']}")
@@ -335,11 +335,14 @@ def main():
     parser.add_argument("--enable-lora", action="store_true", help="Enable LoRA support")
     parser.add_argument("--enforce-eager", action="store_true", help="Enforce eager execution")
     parser.add_argument(
-        "--guided-decoding-backend",
+        "--structured-outputs-backend",
         type=str,
         default="auto",
         choices=["auto", "xgrammar", "outlines"],
-        help="Structured outputs backend (default: auto)",
+        help=(
+            "Structured outputs backend passed to"
+            " --structured-outputs-config.backend (default: auto)"
+        ),
     )
     parser.add_argument(
         "--vllm-logging-level",
@@ -361,7 +364,7 @@ def main():
         skip_existing=not args.force,
         enable_lora=args.enable_lora,
         enforce_eager=args.enforce_eager,
-        guided_decoding_backend=args.guided_decoding_backend,
+        guided_decoding_backend=args.structured_outputs_backend,
     )
 
 
--- a/apps/polyphemus/model_deployment/deploy.py
+++ b/apps/polyphemus/model_deployment/deploy.py
@@ -35,7 +35,7 @@ def deploy_model_vllm(
     service_account: str,
     enable_lora: bool = False,
     enforce_eager: bool = False,
-    guided_decoding_backend: str = "auto",
+    structured_outputs_backend: str = "auto",
 ) -> tuple[aiplatform.Model, aiplatform.Endpoint]:
     """Deploy a model with vLLM on Vertex AI.
 
@@ -44,7 +44,7 @@ def deploy_model_vllm(
         service_account: Service account email for deployment
         enable_lora: Enable LoRA support
         enforce_eager: Enforce eager execution
-        guided_decoding_backend: Structured outputs backend (auto, xgrammar, outlines)
+        structured_outputs_backend: Structured outputs backend (auto, xgrammar, outlines)
 
     Returns:
         Tuple of (Model, Endpoint) objects
@@ -65,7 +65,7 @@ def deploy_model_vllm(
         model_config=model_config,
         enable_lora=enable_lora,
         enforce_eager=enforce_eager,
-        structured_outputs_backend=guided_decoding_backend,
+        structured_outputs_backend=structured_outputs_backend,
     )
 
     logger.info(f"Uploading model: {model_config['model_name']}")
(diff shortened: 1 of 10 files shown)


harry-rhesis (changes requested):
Thanks for this, Alexey -- source-grounding the OWASP synthesizer against the actual PDF is a strong improvement. Inline comments cover the specifics, but the two key items to address are the incomplete parameter rename in `deploy.py` and the `behavior` field, which should be set to `"OWASP LLM Top 10"` rather than `"Robustness"`.

sdk/src/rhesis/sdk/synthesizers/owasp_synthesizer.py
```diff
@@ -463,25 +165,49 @@ def _generate_without_sources(self, num_tests: int = 10, **kwargs: Any) -> List[
...
-            "owasp_description": category.description,
-            "owasp_attack_instructions": category.attack_instructions,
-            "owasp_examples": category.examples,
-            "behavior": category.behavior,
-            "category": "Harmful",
-            "topic": category.topic,
+            "section_id": section.id,
+            "section_name": section.name,
+            "section_content": section.content,
+            "behavior": "Robustness",
```
harry-rhesis: **Please fix:** This should be \`"OWASP LLM Top 10"\` (or the equivalent label for the active report) rather than \`"Robustness"\`. The old code mapped each category to its own behavior (Compliance, Reliability, Robustness). Hardcoding \`"Robustness"\` here collapses all OWASP tests into a single bucket, which breaks analytics grouping and dashboard filters that expect the framework label. Using \`"OWASP LLM Top 10"\` keeps those pipelines working regardless of which per-risk behavior each section would have mapped to.

apps/polyphemus/model_deployment/deploy.py
```diff
@@ -44,7 +44,7 @@ def deploy_model_vllm(
         service_account: Service account email for deployment
         enable_lora: Enable LoRA support
         enforce_eager: Enforce eager execution
-        guided_decoding_backend: Guided decoding backend
+        guided_decoding_backend: Structured outputs backend (auto, xgrammar, outlines)
```
harry-rhesis: **Please fix:** The docstring was updated to "Structured outputs backend" but the parameter on line 38 is still named `guided_decoding_backend`. `config.py` and the CLI flag were correctly renamed to `structured_outputs_backend` -- the function signature here (and `deploy_models` on line 197) should match.

apps/polyphemus/model_deployment/deploy.py
```diff
@@ -361,7 +364,7 @@ def main():
         skip_existing=not args.force,
         enable_lora=args.enable_lora,
         enforce_eager=args.enforce_eager,
-        guided_decoding_backend=args.guided_decoding_backend,
+        guided_decoding_backend=args.structured_outputs_backend,
```
harry-rhesis: **Please fix:** The keyword argument name here should also be `structured_outputs_backend` once the `deploy_models` signature is updated. Currently the old keyword name is passed the new CLI arg value, which works at runtime but is confusing to read.

examples/owasp-synthesizer.ipynb
```diff
@@ -0,0 +1,188 @@
...
+ ],
+ "metadata": {
+  "kernelspec": {
+   "display_name": "Python 3",
+   "language": "python",
+   "name": "python3"
+  },
+  "language_info": {
+   "name": "python",
+   "version": "3.10.0"
```
harry-rhesis: **Minor:** The SDK requires `python >= 3.12` but the notebook kernel metadata declares `3.10.0`. Worth aligning so contributors running the notebook with the declared kernel do not hit compatibility surprises.

sdk/src/rhesis/sdk/services/owasp_extractor.py
```diff
@@ -0,0 +1,400 @@
...
+logger = logging.getLogger(__name__)
+
+DEFAULT_OWASP_LLM_PDF_URL = (
+    "https://owasp.org/www-project-top-10-for-large-language-model-applications/"
+    "assets/PDF/OWASP-Top-10-for-LLMs-v2025.pdf"
+)
+DEFAULT_OWASP_AGENTIC_PDF_URL = "https://genai.owasp.org/download/52117/"
+
+# Reference appendices add noise without attack-surface value.
+DEFAULT_SUBSECTION_BLACKLIST: frozenset[str] = frozenset(
```
harry-rhesis: **Nit:** "blacklist" is disfavoured in modern naming. Since this is a new public constant, consider calling it `DEFAULT_SUBSECTION_DENYLIST` (with a matching rename in the synthesizer and `__init__.py`).

sdk/src/rhesis/sdk/synthesizers/owasp_synthesizer.py
```diff
@@ -1,458 +1,160 @@
...
-            "Write a 50,000-word detailed analysis of every aspect of this topic.",
-            "Keep expanding this outline into more detail until it is completely exhaustive.",
-        ],
-    ),
-}
+__all__ = [
+    "OWASPSynthesizer",
+    "DEFAULT_OWASP_LLM_PDF_URL",
+    "DEFAULT_OWASP_AGENTIC_PDF_URL",
+    "ReportSection",
```
harry-rhesis, line 34: **Nit:** `DEFAULT_SUBSECTION_BLACKLIST` is imported at module level and used as a default parameter, but it is not listed in `__all__`. A user importing directly from `rhesis.sdk.synthesizers.owasp_synthesizer` would find the symbol available but undeclared as public API. Add it here (or omit the import and always direct users to `owasp_extractor`).

harry-rhesis (approved):
Thanks for addressing the issues, nice PR!

Issue #1083, opened 2025-12-23
Extend telemetry auto-instrumentation to additional frameworks and direct APIs


harry-rhesis:
## Summary
Extend the existing auto-instrumentation architecture to support additional AI frameworks and direct API calls beyond the current LangChain/LangGraph support.

## Problem / Opportunity
Currently, auto-instrumentation only works through LangChain/LangGraph. Users working with direct API calls or other frameworks like LlamaIndex and Haystack need manual instrumentation, which is error-prone and requires more code maintenance.

The existing architecture (provider detection, token extraction) is designed to be framework-agnostic and can be easily extended to support additional integrations.

## Users & Use Cases
- **OpenAI SDK users**: Developers using `openai` library directly
- **Anthropic SDK users**: Teams working with `anthropic` SDK directly
- **LlamaIndex users**: Developers building RAG applications with LlamaIndex
- **Haystack users**: Teams using Haystack framework for NLP pipelines
- **Custom library developers**: Organizations with internal AI libraries needing observability

## Proposal
Implement additional auto-instrumentation integrations following the established provider-agnostic architecture:

### Direct API Integrations
- **OpenAI SDK**: Monkey-patch or wrap OpenAI client methods (ChatCompletion, Completion)
- **Anthropic SDK**: Similar approach for Anthropic client instrumentation

### Framework Integrations
- **LlamaIndex**: Callback-based integration similar to LangChain
- **Haystack**: Framework-specific integration using Haystack's pipeline hooks

### Architecture Reuse
All integrations will leverage existing utilities:
- `sdk/src/rhesis/sdk/telemetry/utils/provider_detection.py` - Provider identification
- `sdk/src/rhesis/sdk/telemetry/utils/token_extraction.py` - Universal token counting
- `sdk/src/rhesis/sdk/telemetry/integrations/base.py` - Base integration class

### Non-Goals / Out of Scope
- Synchronous blocking of API calls for instrumentation
- Support for deprecated library versions
- Custom cost models (use LiteLLM's pricing)

## Acceptance Criteria
- [ ] Direct OpenAI SDK instrumentation (`openai` library)
- [ ] Direct Anthropic SDK instrumentation (`anthropic` library)
- [ ] LlamaIndex callback-based integration
- [ ] Haystack framework integration
- [ ] Extensible provider detection system for custom libraries
- [ ] Consistent token extraction across all new integrations
- [ ] Maintain existing LangChain/LangGraph functionality (no regressions)
[...]

Issue #10, opened 2025-05-19
npm warn deprecated eslint@8.57.1


nicolai-rhesis:
npm warn deprecated eslint@8.57.1: This version is no longer supported. Please see https://eslint.org/version-support for other options.

akwasigroch:
Fixed: we are using eslint version 9

PR #2057, opened 2026-06-26
Add Pydantic AI integration (auto-instrumentation + Penelope target)

### LINKED ISSUE #1867: Add Pydantic AI integration (auto-instrumentation + Penelope Target)
## Summary

Add full Pydantic AI integration: automatic telemetry instrumentation of Pydantic AI agent runs and a `PydanticAITarget` for Penelope conversation testing.

## Problem / Opportunity

Pydantic AI is gaining adoption for building type-safe AI agents. Users need automatic observability and a way to run Penelope tests against their Pydantic AI agents.

Related manual instrumentation examples: #1831, #1855

Part of container issue: #1083

### Auto-instrumentation

Implement `PydanticAIIntegration` in `sdk/src/rhesis/sdk/telemetry/integrations/pydantic_ai.py` using Pydantic AI's instrumentation hooks (e.g. `UsageLimits`, `instrument_all()`, or custom middleware) to automatically capture:
- Agent input/output messages
- Token usage per run
- Model name and provider detection
- Latency and errors

Register in `sdk/src/rhesis/sdk/telemetry/integrations/__init__.py`.

### Penelope Target

Implement `PydanticAITarget` in `penelope/src/rhesis/penelope/targets/pydantic_ai.py` wrapping a Pydantic AI `Agent` so Penelope can run multi-turn conversation simulations against Pydantic AI agents.

Export from `penelope/src/rhesis/penelope/targets/__init__.py`.

## Acceptance Criteria

- [ ] `PydanticAIIntegration` auto-instruments Pydantic AI agent runs
- [ ] Token usage, model, latency captured automatically
- [ ] `PydanticAITarget` wraps `pydantic_ai.Agent` for Penelope testing
- [ ] Multi-turn conversations work via message history
- [ ] Unit tests for both integration and target
- [ ] Existing LangChain/LangGraph tests still pass

rhiannalitchfield:
## Purpose

Closes #1867. Adds Pydantic AI support to two parts of the platform: automatic
OpenTelemetry instrumentation in the SDK, and a Penelope target so Pydantic AI
agents can be used in multi-turn adversarial testing.

## What Changed

- `PydanticAIIntegration` (`sdk/src/rhesis/sdk/telemetry/integrations/pydantic_ai.py`)
  - Monkey-patches `Agent.run` (same approach as `LangGraphIntegration`) to
    auto-capture input/output, model name/provider, token usage, and errors as
    an `ai.agent.invoke` OTel span — no manual `@observe` calls required
  - `run_sync()` is *not* patched separately: it's a thin
    `run_until_complete(self.run(...))` wrapper, so patching `run()` alone
    instruments both call paths without producing a duplicate nested span
  - Registered in the integrations `__init__.py` alongside langchain/langgraph/autogen
- `PydanticAITarget` (`penelope/src/rhesis/penelope/targets/pydantic_ai.py`)
  - Wraps a Pydantic AI `Agent` for Penelope, tracking `message_history` per
    `conversation_id` for multi-turn conversations
  - `a_send_message` overridden to use the agent's native `run()` directly
    (skips the default thread-pool fallback)
  - Exported from the targets `__init__.py`
- Added a `pydantic-ai` optional extras group to `sdk/pyproject.toml`, mirroring
  the existing `langchain`/`langgraph`/`autogen` groups
- Unit tests for both: `tests/sdk/telemetry/integrations/test_pydantic_ai_integration.py`
  (span attributes/events, token usage, error status, duplicate-span regression)
  and `tests/penelope/targets/test_pydantic_ai.py` (init/validation, multi-turn
  isolation, sync/async paths, error handling)

## Additional Context

- Discovered and fixed a real cross-test pollution risk while writing tests:
  since the integration patches the shared `pydantic_ai.Agent` class directly,
  the test fixture restores `Agent.run` after every test (not just resetting
  internal flags), otherwise patches leak into other test modules that import
  `Agent` in the same pytest process
- Verified manually against pydantic-ai 1.94.0 with `TestModel`

## Testing

- `uv run pytest tests/sdk/telemetry/integrations/test_pydantic_ai_integration.py tests/penelope/targets/test_pydantic_ai.py -v`
- Full regression: `uv run pytest tests/sdk/ tests/penelope/` (599 passed, 0 failed)
- `ruff check` clean on all new/modified files


rhiannalitchfield:
Thanks for the review! Addressed in 192f888b:

1. **validate_configuration gap** — fixed. Added a `hasattr(self.agent, "run")` check alongside the existing `run_sync` check, and a regression test (`test_target_rejects_agent_without_run`).

2. **Tracer captured at enable-time** — I looked into this one. `trace.get_tracer()` returns an OTel `ProxyTracer`, whose internal `_tracer` property re-checks the global `TracerProvider` lazily on every span-start call until a real one is configured, then caches it. It does not get permanently stuck on NoOp — that's the whole point of the proxy. This is also the same pattern already used by `LangChainIntegration`'s callback (`self.tracer = trace.get_tracer(__name__)` in `__init__`), so I kept it consistent with the existing convention rather than introducing a different resolution strategy for just this one integration.

rhiannalitchfield:
Re: the file-attachment finding — went ahead and implemented real support across all three targets in 4e817d2f rather than leaving it as a known gap:

- Added a shared `files_to_content_blocks()` helper (`penelope/targets/_content_blocks.py`) used by both `LangChainTarget` and `LangGraphTarget` to turn `files` into real multimodal content blocks (image/file/audio/video) instead of silently dropping them.
- `PydanticAITarget` gets the equivalent via `BinaryContent`.
- `LangChainTarget` needed a different approach: its generic `{input_key}` string-templating just stringifies a list of content blocks rather than treating them as multimodal content (verified empirically), so when `files` is present it now bypasses the dict/template path and invokes the runnable with a `HumanMessage` directly. This is best-effort — it works for runnables that accept a message at the top level (bare chat models, `MessagesPlaceholder`-first prompts) but isn't guaranteed for every possible chain shape, and additional `**kwargs` aren't merged into the input in that branch. Documented in the method's docstring.

Added test coverage for the new behavior in all three targets plus the shared helper (`test_content_blocks.py`, `test_langchain.py`, `test_langgraph.py`, updates to `test_pydantic_ai.py`). Full regression: 616 passed.

rhiannalitchfield:
Thanks for catching this — both critical findings were real regressions from the file-attachment commit (4e817d2). Fixed in 0517dc3:

- **Unconditional imports**: moved the `langchain_core` imports in `_content_blocks.py`/`langchain.py` and the `pydantic_ai` import in `pydantic_ai.py` (target) from module level into the functions that actually need them. Verified with a new regression test (`test_optional_imports.py`) that simulates blocking all three optional packages and confirms `import rhesis.penelope.targets` still succeeds and `EndpointTarget` still works.
- Added the missing `pydantic-ai` optional extra to `penelope/pyproject.toml`, mirroring the existing `langchain`/`langgraph` groups (it had a hard implicit dependency with nowhere documented to install it from).
- **kwargs question**: confirmed empirically that wrapping the HumanMessage back into `{input_key: HumanMessage(...), **kwargs}` doesn't actually solve the stringification problem — LangChain's template substitution stringifies whatever's under `input_key` regardless of its type when the slot is a plain string template (`"{input}"`). So there's no way to support both real attachments and templated kwargs simultaneously for an arbitrary chain; dropping kwargs when files are present is the only viable option, not an oversight. Made this more explicit in the docstring.

Full regression: 618 passed.

rhiannalitchfield:
Both findings confirmed real and fixed in 2ffe92a:

1. **Data leak (sdk/.../pydantic_ai.py:47)** — verified empirically: `BinaryContent`'s `repr()` embeds the raw attachment bytes verbatim, and `_extract_user_prompt()` called `str()` on the whole prompt list before truncating to `MAX_CONTENT_LENGTH` (8000 chars), so real file content was ending up in the `ai.prompt.content` span. Fixed: now only text parts are recorded; non-text parts (BinaryContent, ImageUrl, etc.) get a `<attachment: kind>` placeholder instead of their repr.

2. **FileReference shape (penelope/.../pydantic_ai.py:31)** — this turned out to be broader than just `PydanticAITarget`. I found `_file_compat.py`, which documents that Penelope already has two legitimate file shapes reaching targets: plain dicts with inline base64, and `FileReference` (object-storage-backed, bytes fetched on-demand via `read_bytes()`/`aread_bytes()`). I traced the actual call path (`target_interaction.py` → `target.send_message(files=...)`) and confirmed `FileReference` objects really do reach here, not just hypothetically — and the same dict-only assumption existed in `_content_blocks.py` too (used by LangChainTarget/LangGraphTarget), not just `PydanticAITarget`.

   Added `file_bytes_and_type()` / `afile_bytes_and_type()` / `file_extracted_text()` to `_file_compat.py` and updated all three targets to use them — preferring pre-extracted text when available (cheap, no network call) and materializing real bytes via `read_bytes()`/`aread_bytes()` otherwise, with the async path correctly using `aread_bytes()` so it doesn't block the event loop.

Full regression: 632 passed. New tests cover both fixes (`test_file_compat.py`, plus FileReference cases added to `test_content_blocks.py`, `test_pydantic_ai.py`, and a safe-extraction test class in the SDK integration tests).

sdk/src/rhesis/sdk/telemetry/integrations/pydantic_ai.py
```diff
@@ -0,0 +1,226 @@
...
+import logging
+from typing import Any, Callable, Optional
+
+from opentelemetry import trace
+from opentelemetry.trace import SpanKind, Status, StatusCode
+
+from rhesis.sdk.telemetry.attributes import AIAttributes, AIEvents
+from rhesis.sdk.telemetry.context import is_tracing_disabled
+from rhesis.sdk.telemetry.integrations.base import BaseIntegration
+from rhesis.sdk.telemetry.integrations.langchain.extractors import MAX_CONTENT_LENGTH
```
harry-rhesis: **Should fix**: `MAX_CONTENT_LENGTH` is a framework-agnostic constant (the max chars to record in a span event), but importing it from `rhesis.sdk.telemetry.integrations.langchain.extractors` creates a coupling between the Pydantic AI integration and the LangChain subpackage. Consider moving this constant to a shared location like `attributes.py` or `base.py` so that Pydantic AI does not depend on LangChain internals.

sdk/src/rhesis/sdk/telemetry/integrations/pydantic_ai.py
```diff
@@ -0,0 +1,226 @@
...
+        span.set_attribute(AIAttributes.AGENT_NAME, agent_name)
+
+    model = getattr(agent, "model", None)
+    model_name = getattr(model, "model_name", None)
+    if model_name:
+        span.set_attribute(AIAttributes.MODEL_NAME, str(model_name))
+
+    provider = getattr(model, "system", None)
+    if provider:
+        span.set_attribute(AIAttributes.MODEL_PROVIDER, str(provider))
```
harry-rhesis, line 94: **Should fix**: `model.system` is not a documented public attribute in Pydantic AI. If the internal API changes across versions, this will silently stop recording the provider. Consider adding a fallback, e.g. `type(model).__name__` or `model.__class__.__module__.split(".")[0]`, so the provider field degrades gracefully rather than disappearing.

sdk/src/rhesis/sdk/telemetry/integrations/pydantic_ai.py
```diff
@@ -0,0 +1,226 @@
...
+    """Record output content and token usage from an AgentRunResult."""
+    output = getattr(result, "output", None)
+    if output is not None:
+        span.add_event(
+            AIEvents.COMPLETION,
+            {AIAttributes.COMPLETION_CONTENT: str(output)[:MAX_CONTENT_LENGTH]},
+        )
+
+    usage_fn = getattr(result, "usage", None)
+    usage = usage_fn() if callable(usage_fn) else None
```
harry-rhesis: **Should fix**: `usage_fn = getattr(result, "usage", None); usage = usage_fn() if callable(usage_fn) else None` works today because `usage` is a method on `AgentRunResult`. If Pydantic AI changes it to a property in a future release, `callable()` will return `False` and token counts will silently stop being recorded. Consider handling both patterns:

```python
usage_attr = getattr(result, "usage", None)
usage = usage_attr() if callable(usage_attr) else usage_attr
```

sdk/src/rhesis/sdk/telemetry/integrations/pydantic_ai.py
```diff
@@ -0,0 +1,226 @@
...
+                    _record_run_end(span, result)
+                    return result
+                except Exception as e:
+                    _record_run_error(span, e)
+                    raise
+
+        Agent.run = patched_run
+
+        AgentPatchState.set_done()
+        logger.debug("Patched Agent.run for automatic span creation")
```
harry-rhesis, line 238: **Must fix (3 items)**:

1. **`run_stream()` is not instrumented.** `Agent.run_stream()` is a separate code path that does not delegate to `run()`. Users calling it will get no span at all. For consistency with LangGraph (which patches `invoke`, `ainvoke`, `stream`, and `astream`), this should be patched here as well.

2. **Tool calls are invisible.** Tools registered with `@agent.tool` execute inside the patched `Agent.run()`, but only the outer `ai.agent.invoke` span is created. The LangChain/LangGraph integrations create individual `ai.tool.invoke` child spans for each tool call, capturing tool name, input, and output. The Pydantic AI integration should produce equivalent child spans. Since Pydantic AI has no callback system, this likely requires hooking into the internal tool dispatch or leveraging Pydantic AI's native instrumentation / Logfire hooks.

3. **Agent handoffs are not detected.** Pydantic AI supports multi-agent delegation (one agent can call another agent as a dependency). The integration should detect these transitions and create `ai.agent.handoff` spans with `AIAttributes.AGENT_HANDOFF_FROM` / `AIAttributes.AGENT_HANDOFF_TO`, matching the LangChain/LangGraph pattern.

harry-rhesis (commented):
Thank you for the excellent work on this PR, @rhiannalitchfield! And for the thorough back-and-forth with peqy to address every finding. The iterative improvements (file attachment support across all targets, the data-leak fix, FileReference compatibility) show real diligence. The Penelope target and SDK integration are well-structured, consistent with existing patterns, and the test coverage is solid.

Below are recommendations focused specifically on the Pydantic AI implementation. Cross-cutting improvements (async parity for LangChain/LangGraph targets, PatchState deduplication, `__init__.py` import guards) are out of scope for this PR and will be tracked separately.

### Resolve Conflicts
The PR is displaying a few conflicts, it would be important to solve them before continuing. 

### Must Fix

1. **Instrument `run_stream()`** - `Agent.run_stream()` is a separate code path that does not delegate to `run()`. Users calling it will get no span. Patch it alongside `run()` in `_patch_agent()`, consistent with how LangGraph patches all four entry points.

2. **Instrument tool calls** - Tools registered with `@agent.tool` execute inside `Agent.run()` but produce no `ai.tool.invoke` spans. The LangChain/LangGraph integrations create individual child spans for each tool invocation with tool name, input, and output. The Pydantic AI integration should do the same.

3. **Detect agent handoffs** - Pydantic AI supports multi-agent delegation (one agent calling another as a dependency). The integration should detect these transitions and create `ai.agent.handoff` spans with `AGENT_HANDOFF_FROM` / `AGENT_HANDOFF_TO`, consistent with the LangChain/LangGraph pattern.

4. **Update `examples/telemetry/pydantic_ai_example.py`** - The existing example uses manual `@observe` and states that auto-instrumentation is "planned in #1083". This PR delivers that feature, but the example does not showcase it. Update the example to demonstrate `auto_instrument()` with zero-config tracing, matching the pattern of `langchain_example.py`.

5. **Add `penelope/examples/pydantic_ai_minimal.py`** - Both LangChain and LangGraph have minimal Penelope examples (`langchain_minimal.py`, `langgraph_minimal.py`). Add one for Pydantic AI showing `PydanticAITarget` + `PenelopeAgent.execute_test()`.

6. **Add a multi-agent Pydantic AI example** - Demonstrate agent delegation and handoffs with Pydantic AI, showing how the instrumentation traces agent transitions.

### Should Fix

7. **Move `MAX_CONTENT_LENGTH` import** (see inline comment) - This constant is framework-agnostic and should not create a dependency on the LangChain subpackage.

8. **Harden provider name extraction** (see inline comment) - `model.system` is not a documented public API. Add a fallback for resilience.

9. **Handle `usage` as property or method** (see inline comment) - The callable check can silently return `None` if Pydantic AI changes the API.

### Consider

10. **Structured output in completion events** - `str(result.output)` works but loses structure for Pydantic model outputs. Consider recording the output type as a span attribute and using `model_dump_json()` for richer trace data when the output is a Pydantic model.

harry-rhesis:
Ah, @rhiannalitchfield, one last item: would you mind signing off your commits? We use Developer Certificate of Origin (DCO). You can check this reference: https://github.com/rhesis-ai/rhesis/blob/main/CONTRIBUTING.md Thank you so much for contributing to our community!

rhiannalitchfield:
Thank you for the thorough review, and for scoping the cross-cutting items out separately, that really helps keep this PR focused.

Here's how I'm planning to handle each group.

**Doing in this PR**

- **Conflicts**: resolving now. They're all in registration/manifest files (the two `__init__.py` files, both `pyproject.toml` files, and `sdk/uv.lock`), none in the integration logic. `main` added the `agent_framework`/MAF integration after I branched, so the merge keeps both that and the new `pydantic_ai` entries.
- **#1 `run_stream()`**: agreed, it's a genuinely separate path that doesn't delegate to `run()`, so it gets no span today. I'll patch it alongside `run()`.
- **#7 `MAX_CONTENT_LENGTH`**: agreed, moving it to a framework-agnostic location (`attributes.py`) so the Pydantic AI integration stops depending on the LangChain subpackage.
- **#8 provider extraction**: agreed, adding a fallback so `model.system` degrades gracefully instead of silently dropping the provider.
- **#9 `usage` as property-or-method**: agreed, switching to the both-patterns form you suggested.
- **#4 / #5 examples**: I'll update `examples/telemetry/pydantic_ai_example.py` to showcase `auto_instrument()` and add `penelope/examples/pydantic_ai_minimal.py`.
- **#10 structured output**: small enough to fold into this batch. I'll record the output type as a span attribute and use `model_dump_json()` when the output is a Pydantic model.

**Want to confirm the approach before building: #2 and #3 (tool spans + handoffs)**

Before I implement these by patching internal tool dispatch, I want to flag something that may change the cleanest path. Pydantic AI ships its own built-in OpenTelemetry instrumentation: enabling it via `logfire.instrument_pydantic_ai()` or `Agent.instrument_all(InstrumentationSettings(...))` already emits spans for agent runs, model calls, and tool function execution, and by default it writes to the global `TracerProvider`, the same provider this integration uses. Docs: https://ai.pydantic.dev/logfire/

Two questions I'd like your call on:

1. Since the framework already produces tool-call spans natively, would you prefer this integration route through `instrument_all()` rather than re-deriving them by monkey-patching private dispatch internals? The manual approach is what #2 would require (Pydantic AI has no callback system), and it would be re-implementing supported instrumentation against internals that can shift between versions.
2. There's also a latent duplication risk worth surfacing: because both the current `Agent.run` patch and the native instrumentation target the same global `TracerProvider`, a user who enables both would get overlapping `ai.agent.invoke` spans. Settling the mechanism (patch vs. native) resolves that too.

For **#3 (handoffs)**: Pydantic AI doesn't have a first-class handoff primitive in the way the term implies. Multi-agent delegation is just one agent called inside another agent's tool, so if both agents are instrumented the delegation already nests as child spans. I'm happy to map that onto `AGENT_HANDOFF_FROM`/`AGENT_HANDOFF_TO` for attribute parity with LangChain/LangGraph, but wanted to confirm that's the intent rather than synthesize a primitive that isn't there.

Given the above, would you be open to landing #1, #4, #5, #7, #8, #9, #10 plus the conflict resolution in this PR, and tracking #2, #3, and the multi-agent example (#6) as a follow-up once we've settled the patch-vs-native question? That mirrors how you've already scoped the other cross-cutting work out. Happy to go whichever way you prefer.

harry-rhesis:
Great catch, @rhiannalitchfield. Agreed, let's go with Pydantic AI's native instrumentation instead of monkey-patching. It's the pattern we standardized on after you branched: the MAF integration hits the same situation (native `gen_ai.*` spans) and solves it by enabling native instrumentation plus a translator that rewrites `gen_ai.*` into our `ai.*` schema. That gives agent/chat/**tool** spans for free, and it avoids the duplication you flagged. Your handoff read is right too: MAF has no real handoff primitive, so its translator *synthesizes* `ai.agent.handoff` spans from the handoff tool-calls, and we'd do the same for delegation.

For context, we may later adopt the emerging OTel GenAI conventions, but for now we keep our own `ai.*` schema, so the translation layer is the bridge in the meantime.

Plan:

- **This PR:** conflict resolution, #4/#5 (examples), #7, #8, #9, #10.
- **Skip #1 (`run_stream`) here:** no point adding patching we're about to remove, so fold it into the migration.
- **Follow-up:** replace the `Agent.run` patch with a `PydanticAITranslatingExporter` mirroring MAF (covers #2, #3, `run_stream`, #6). Both frameworks emit `gen_ai.*`, so much of MAF's `mapping.py` should be reusable. Let's coordinate with #2070 to avoid two GenAI translators.

Two things to confirm while scoping it:

1. **Logfire dependency.** Confirm we can enable instrumentation via `Agent.instrument_all(InstrumentationSettings(...))` so spans flow into the global `TracerProvider` that `RhesisClient` already configures, without making `logfire` a hard runtime dep. If spans only emit with Logfire installed, we'll need an optional extra and a fallback.
2. **Message shape.** GenAI allows two forms: per-message span *events* (`gen_ai.user.message`, etc.) or span *attributes* with JSON arrays (`gen_ai.input.messages` / `gen_ai.output.messages`). Our UI reads events, and MAF handles both. We just need to know which Pydantic AI emits to gauge how much of `mapping.py` is reusable.

The Penelope `PydanticAITarget` is fine to ship as-is. Thanks for surfacing this before sinking time into the harder items, and thank you for the contribution. If it's easier to talk any of this through, you can also reach me on our Discord: rhesis.discord.ai.

rhiannalitchfield:
Researched both scoping questions:

**1. Logfire dependency: not required.** Pydantic AI's instrumentation is plain OpenTelemetry; Logfire is only one exporter option (docs: "You can also emit OpenTelemetry data from Pydantic AI without using Logfire at all," with a `raw_otel.py` example using just `opentelemetry-sdk` + the OTLP exporter). `InstrumentationSettings(tracer_provider=None)` defaults to the global `TracerProvider`, so `Agent.instrument_all(InstrumentationSettings())` flows spans into the same provider `RhesisClient` configures. No optional `logfire` extra or fallback needed for the core path; the only requirement is the OTel SDK, which the rhesis SDK already pulls in.

**2. Message shape: span attributes (JSON arrays), not events.** Current Pydantic AI emits `gen_ai.input.messages` / `gen_ai.output.messages` on model-request spans and `pydantic_ai.all_messages` on agent-run spans. There's no per-message event mode anymore; the current `InstrumentationSettings` has no `event_mode` param. So the translator will reuse MAF's attribute-handling path (parse the JSON arrays into our `ai.*` events) rather than its event path.

Two related notes for the follow-up:
- I'd pin `version` explicitly in `InstrumentationSettings` (default is the latest, currently `5`), so the `gen_ai.*` attribute schema can't shift under us on a Pydantic AI bump.
- `include_binary_content` defaults to `True`, which is the same leak surface I closed in 2ffe92a. The translator should set it to `False` (or scrub) so that fix carries over to the native path.

Happy to coordinate with #2070 on the shared translator.

harry-rhesis:
@rhiannalitchfield, sounds awesome thank you for the detailed follow-up! I'll defer to your call here. If you feel like it's better to address https://github.com/rhesis-ai/rhesis/issues/2070 first, and then rebase this work, we could do that! Otherwise we can do it the other way around. Finalize the implementation here, work on https://github.com/rhesis-ai/rhesis/issues/2070 and then iron things out. Might be better because then we know for sure what can be generalized. What do you think?

rhiannalitchfield:
Agreed, let's finalize #2057 first, then the translator, then #2070. One reason to specifically avoid doing #2070 ahead of the translator: two of its deliverables are written against the current monkey-patch that the translator removes. The `PatchState` base extraction pulls from `AgentPatchState`, which goes away when `Agent.run` patching is replaced by the exporter, and the `ai.llm.invoke` child spans for Pydantic AI come for free from the native `gen_ai.*` model-call spans once the translator lands. So doing those two first would just be rework.

Proposed sequence:
1. Finalize #2057 (Penelope target, #7/#8/#9/#10, examples). It doesn't touch #2070's code, so no rebase risk.
2. Translator migration (`PydanticAITranslatingExporter`, mirroring MAF): removes the `Agent.run` patch, covers #2, #3, `run_stream`, #6.
3. #2070 last. Its Pydantic-specific items are handled by then, so it reduces to the genuinely cross-cutting work (import guards, `LangChainTarget`/`LangGraphTarget` native async, shared `GraphPatchState` base). Might be worth noting that dependency on the #2070 issue so nobody starts the `PatchState`/`ai.llm.invoke` items early.

Sounds good to move ahead finalizing here?

harry-rhesis:
Sounds like an excellent plan, @rhiannalitchfield! Thanks for going the extra mile!

rhiannalitchfield:
Finalization is pushed. Summary of the two pushes:

**Push 1 (109ce51b): finalize per the agreed scope**
- **Conflicts resolved**: merged `upstream/main` keeping both the MAF integration/target (including the `MAFTarget` rename and its deprecation alias) and the new `pydantic_ai` entries. Both lockfiles regenerated; that also fixed a latent gap where `penelope/uv.lock` never picked up the `pydantic-ai` extra added earlier.
- **#7**: `MAX_CONTENT_LENGTH` moved to `telemetry/attributes.py` (framework-agnostic home), re-exported from `langchain.extractors` for backwards compatibility.
- **#8**: provider extraction falls back to the model class module when `model.system` is missing.
- **#9**: `usage` handled as property or method per your suggestion.
- **#10**: structured outputs recorded via `model_dump_json()` plus a new `ai.completion.output_type` span attribute, with a guard so a failing dump can never break the user's call path.
- **#4/#5**: telemetry example rewritten around `auto_instrument()` zero-config tracing (honest that native tool/handoff spans arrive with the translator follow-up), and `penelope/examples/pydantic_ai_minimal.py` added. Also exported `PydanticAITarget` from the `rhesis.penelope` package root for parity with the other targets.
- **DCO**: all commits now carry `Signed-off-by` matching their author.

**Push 2 (84b028c3): peqy's two findings**
- **Bare-attachment leak**: confirmed real. A single multimodal part passed directly as `user_prompt` (not in a list) fell through to `str()`. It now gets the same `<attachment: kind>` placeholder as the list case, with a regression test.
- **Sync blocking**: went with documentation rather than loop-detection, because option (b) can't actually work here: `send_message()` delegates to `agent.run_sync()`, which itself refuses to run inside a running event loop regardless of how the file bytes are read. The docstring now states that `send_message()` blocks and async contexts must use `a_send_message()`.

Test state: 21/21 SDK pydantic-ai integration tests, 26/26 Penelope pydantic-ai target tests, 583/583 penelope suite, SDK suite green apart from a couple of pre-existing timing-flaky tests in `connector/test_executor*` and `test_exporter` (verified they fail identically without these changes).

rhiannalitchfield:
Confirmed real and fixed in ca1ffce9. I traced it to the backend to be sure: `tasks/file/extract_text.py` runs extraction for `("image/", "application/pdf", "text/")` with `extract_with_vision_fallback()`, so image FileReferences genuinely do arrive with OCR text in `extracted_text` — and all three targets would have sent the OCR text and silently dropped the real image.

Fix: gated the fast-path inside the shared `file_extracted_text()` helper to document-like content types (`text/*`, `application/pdf`); anything else falls through to the bytes path. Since all three targets (LangChain, LangGraph, PydanticAI) route through that one helper, this fixes them in one spot. That gate exactly complements the backend's extractable set: text and PDF keep the cheap text path, images keep their binary.

I went with gating rather than attaching both (text snippet + binary) to keep the change minimal; if you'd rather have both parts sent for OCR'd images, that's a small follow-up.

Regression tests added at all three layers: the helper (`test_file_compat.py`), the shared content blocks (`test_content_blocks.py`), and `PydanticAITarget` (image with OCR text still produces `BinaryContent`). Penelope suite: 587 passed.

rhiannalitchfield:
Good question. Yes, text-first for PDFs is the intended default, and it's deliberate rather than incidental:

- It predates the gating fix: PDFs with `extracted_text` were already sent as text in this PR (and reviewed as such); ca1ffce9 only stopped images from being swallowed by their OCR text.
- The backend treats PDF `extracted_text` as a faithful stand-in by design: `extract_with_vision_fallback` (documented as "the single authoritative extraction path") does text-layer extraction first, then a vision-model fallback for image-heavy documents. So even scanned/figure-heavy PDFs arrive with vision-derived text, not just a bare text layer. That's a different situation from images, where the OCR text is a lossy description of an inherently visual artifact.
- Practically, the text path works for text-only targets (many Penelope targets are), avoids re-fetching bytes from object storage on every conversation turn, and avoids base64-inflating the payload (a full PDF as base64 plus its extracted text roughly doubles the cost per turn if we attach both).

The trade-off is real though: a multimodal-capable target never sees the PDF's actual layout/figures. If you want that, the clean shape would be an opt-in flag on the targets (e.g. `prefer_extracted_text: bool = True`) rather than changing the default, since it's a per-target capability question that the shared helper can't decide. Happy to add that here or in the follow-up, whichever @harry-rhesis prefers — my lean is follow-up, since the translator work is already queued and this PR is otherwise settled.

harry-rhesis:
@rhiannalitchfield, thank you so much for addressing the issues! 

I agree, let's leave text-first for PDFs as the default for now and follow-up at a later stage. 

There are a few failing checks. Would you mind checking those? As a further (and last step) for this integration, one idea would be to add one PydanticAI agent to our `agents` folder, where we connect the agent to the Rhesis platform, and can see the real traces in Rhesis. But this is something for another PR. :)

rhiannalitchfield:
Fixed in aa1bc944. Both failures were the same root cause: `test-sdk` and `test-penelope` run `uv run pytest` without extras, and `pydantic-ai` only existed as an optional extra, so the new test modules failed collection with `ModuleNotFoundError`. This didn't bite before the merge because the dev-group convention for integration test deps (the one the MAF subpackages use, with the comment about avoiding `--all-extras`) landed on main after I branched.

Fix follows that exact pattern: `pydantic-ai` added to `[dependency-groups] dev` in both `sdk/pyproject.toml` and `penelope/pyproject.toml` (end users still install via the optional extras), locks regenerated. Verified locally with plain `uv run pytest` (no extras), mirroring CI: 21/21 SDK integration tests, 27/27 Penelope target tests.

And the `agents/` folder demo connecting a Pydantic AI agent to the platform sounds like a great capstone for the integration — happy to pick that up after the translator follow-up.

harry-rhesis:
Solid work, @rhiannalitchfield! Let's merge this! 🎉 Thanks for putting all the extra effort into getting this integration running! The translator follow-up is now the likely suspect right?

rhiannalitchfield:
That's the one! I'll start on the translator (coordinating with #2070 as discussed), and the agents/ folder demo sounds like a great capstone after that. Thanks for the thorough review and for merging! 🎉

PR #2058, opened 2026-06-27
Remove ENVIRONMENT var and fix auth cookie/logout handling

akwasigroch:
## Purpose
The legacy `ENVIRONMENT` variable overlapped with per-service env labels (`BACKEND_ENV`, `FRONTEND_ENV`, etc.) and was still used in auth/cookie logic. This caused confusion and incorrect local-vs-deployed behavior when `BACKEND_ENV=local` was set without a loopback `API_BASE_URL`.

## What Changed
- Remove `ENVIRONMENT` from config, CI, Docker, Helm, and docs; use `BACKEND_ENV` for deployment labels only
- Drop `test` from `BACKEND_ENV` allowed values and remove the unused `is_local` settings property
- Detect local-dev auth/cookie behavior from loopback `API_BASE_URL` (and Quick Start), not `BACKEND_ENV=local`
- Add `logout_cookie_domains()` to derive cookie-clear domains from `FRONTEND_URL` instead of hardcoding
- Simplify `migrate.sh` to gate DB wait/ownership on `BACKEND_ENV=local` only
- Pre-create `logs/` in `Dockerfile.dev` so the non-root user can write at runtime
- Update auth and url_utils tests to match the new behavior

## Additional Context
- `BACKEND_ENV` remains the deployment label for `start.sh` / `migrate.sh` but no longer drives session/cookie logic
- Loopback hostname in `API_BASE_URL` is the signal for local-dev session/cookie behavior

## Testing
- [x] `uv run pytest ../../tests/backend/auth/test_url_utils.py -v`
- [x] `uv run pytest ../../tests/backend/routes/test_auth.py::TestAuthLogout -v`
- [x] `uv run pytest ../../tests/backend/routes/test_auth.py::TestGetCallbackUrl -v`


tests/backend/routes/test_auth.py
```diff
@@ -563,6 +562,30 @@ def test_logout_post_logout_parameter(self, client: TestClient):
...
+            assert response.status_code == status.HTTP_200_OK
+            set_cookie_headers = [
+                value.lower()
+                for key, value in response.headers.items()
+                if key.lower() == "set-cookie"
+            ]
+            assert any("domain=app.example.com" in header for header in set_cookie_headers)
+            assert any("domain=.app.example.com" in header for header in set_cookie_headers)
+            assert any("domain=example.com" in header for header in set_cookie_headers)
+            assert any("domain=.example.com" in header for header in set_cookie_headers)
```
harry-rhesis: [Critical] This test asserts that logout clears cookies for `example.com` and `.example.com`, but `logout_cookie_domains()` now intentionally returns host-only domains (`["app.example.com"]`). The `logout()` function emits clears for `domain` and `.domain`, so only `app.example.com` and `.app.example.com` will appear in the response headers. These two assertions will fail.

Either remove the parent-domain assertions to match the host-only design, or re-introduce parent-domain clearing in `logout_cookie_domains()` if legacy deployments need it. The test name ("parent domain") should also be updated since the design explicitly avoids parent domains.

apps/backend/src/rhesis/backend/app/routers/auth.py
```diff
@@ -1351,7 +1345,8 @@ async def logout(
     # Create response with cookie clearing headers
     accept_header = request.headers.get("accept", "")
     frontend_url = get_frontend_settings().url
-    frontend_env = os.getenv("FRONTEND_ENV", "development")
+    cookie_domains = logout_cookie_domains(frontend_url)
+    cookie_secure = urlparse(frontend_url).scheme == "https"
```
harry-rhesis: [Security] `cookie_secure` is derived from the `FRONTEND_URL` scheme, but `SessionMiddleware` derives its `Secure` policy from `https_only=not is_running_locally()` (in `main.py`). These two signals can diverge:

- If `FRONTEND_URL` is misconfigured to `http://...` in a deployed env, logout emits non-`Secure` clears, but the cookies were set with `Secure=True` by the middleware, so they will not actually be cleared.
- If `FRONTEND_URL=https://...` but the backend is local, logout emits `Secure` clears for cookies that were set without `Secure`.

Safest fix: align with the same signal the middleware uses (`cookie_secure = not is_running_locally()`), or defensively clear both secure and non-secure variants (the mismatched one is a harmless no-op).

apps/backend/src/rhesis/backend/app/routers/auth.py
```diff
@@ -1385,8 +1380,7 @@ async def logout(
             samesite="lax",
```
harry-rhesis, line 1371: [Improvement] This domainless clear loop does not set `secure=True`, so `__Secure-next-auth.callback-url` and `__Secure-next-auth.session-token` (which per the cookie spec require `Secure`) will not actually be cleared in deployed environments. The domain-scoped clears below do set `secure=cookie_secure`, but those specify a domain which may not match how the cookie was originally set (host-only).

This pre-dates the PR, but since the logout logic is being reworked, it would be a good opportunity to add `secure=cookie_secure` to this loop as well (or at least for the `__Secure-` and `__Host-` prefixed names).

apps/backend/start.sh
```diff
@@ -23,15 +23,15 @@ handle_error() {
 
 # Function to check if we're in production mode
 is_production() {
-    [ "${ENVIRONMENT}" = "production" ] || [ "${BACKEND_ENV}" = "production" ]
+    [ "${BACKEND_ENV}" = "production" ]
```
harry-rhesis, line 26: [Nit] `is_production()`, `is_local()`, and `is_development()` use `${BACKEND_ENV}` without a default. While this works with `set -u` disabled, `migrate.sh` already uses `${BACKEND_ENV:-}` consistently. Worth matching that pattern here for safety if `set -u` is ever enabled.

apps/backend/src/rhesis/backend/app/auth/url_utils.py
```diff
@@ -89,3 +89,23 @@ def _is_loopback_dev_origin(parsed_origin) -> bool:
...
+    domain: it would be a no-op (no cookie is scoped there) and could nuke a
+    sibling environment's session if a shared cookie were ever introduced.
+
+    Derived from ``FRONTEND_URL`` so deployments never need a hardcoded domain.
+    """
+    host = (urlparse(frontend_url).hostname or "").lower()
+    if not host or host in _LOOPBACK_HOSTNAMES:
+        return []
+
+    return [host]
```
harry-rhesis: [Note] The host-only design here is the right approach for arbitrary-domain hosting. It avoids the public-suffix problem (e.g. `app.co.uk` trying to clear `co.uk`) and prevents cross-environment cookie interference (e.g. `dev.rhesis.ai` nuking `stg.rhesis.ai` sessions). It also matches how cookies are actually set, since both NextAuth and `SessionMiddleware` use host-only cookies. The docstring explains the rationale well.

harry-rhesis (commented):
Thanks for tackling this. The overall direction is solid: anchoring local-dev detection to an observable signal (loopback `API_BASE_URL`) instead of an ambiguous deployment label is the right call, especially since we can be hosted on any domain. The host-only approach in `logout_cookie_domains()` correctly avoids the public-suffix problem and cross-environment cookie interference.

A few things to address before merging: one critical test/implementation mismatch, one security-relevant alignment issue, and a couple of smaller items. Details in the inline comments.

harry-rhesis:
Hi @akwasigroch, thank you for this PR, was this concern addressed? 

----
Re-review after latest commits: ✅ the outdated parent-domain logout test was removed, and the BACKEND_ENV docs now match the settings validation.

Remaining items from earlier feedback still apply:

logout() derives cookie_secure from FRONTEND_URL scheme; this can drift from the backend’s actual cookie policy (SessionMiddleware https_only=not is_running_locally()), and may prevent clearing secure cookies if misconfigured.
In the host-only cookie-clear loop, consider setting secure= at least for __Secure- / __Host- cookie names so those clears aren’t rejected by browsers.

PR #2066, opened 2026-06-29
Fix endpoint API token field behavior

nicolai-rhesis:
## Purpose
The API token field in the endpoint "Authentication & headers" section behaved incorrectly:
- It showed masked dots (`••••`) for a saved token, but there was no way to delete the saved token.
- After adding a token and saving, the UI still claimed there was no token until a full page reload.

## What Changed
- `EndpointHeadersFields.tsx`: added a remove button (trash icon) for an existing token that toggles to a restore button, with an error-styled helper indicating the token will be removed on save.
- `EndpointConnectionTab.tsx`: added a `clear_token` draft flag; `onSave` now sends `auth_token: null` to remove the token, the typed value to replace it, or omits it when untouched. Typing a replacement cancels a pending removal.
- `endpointPatch.ts`: still strips an untouched empty string but lets an explicit `null` (clear) through to the backend.
- `useEndpointDetail.ts`: keeps the derived `has_auth_token` flag in sync after a save (true when set, false when cleared, unchanged when omitted) so the UI refreshes without a reload.
- `interfaces/endpoint.ts`: `EndpointEditData.auth_token` is now `string | null` (`null` clears, `undefined` keeps).

## Additional Context
- `auth_token` is a write-only field; the backend only returns `has_auth_token`. Sending `null` sets the column to `NULL` via the `EncryptedString` type, so `has_auth_token` becomes `false`.
- No backend changes were required; `EndpointUpdate` uses `exclude_unset=True`, so an explicit `null` is applied while an omitted field is preserved.

## Testing
- `npx tsc --noEmit`, `eslint`, and `prettier --check` pass on the changed files.
- Manual: edit an endpoint with a saved token → masked dots show; click trash → save → token removed and "No token configured" shown. Add a token → save → masked dots show immediately without reload. Leaving the field untouched preserves the existing token.

### DIFF
--- a/apps/frontend/src/app/(protected)/endpoints/[identifier]/components/EndpointConnectionTab.tsx
+++ b/apps/frontend/src/app/(protected)/endpoints/[identifier]/components/EndpointConnectionTab.tsx
@@ -184,8 +184,10 @@ export default function EndpointConnectionTab() {
               setDraft(prev => ({
                 ...prev,
                 auth_token: value,
-                // typing a replacement cancels a pending removal
-                clear_token: value !== '' ? false : prev.clear_token,
+                // typing a real replacement cancels a pending removal;
+                // whitespace-only input is treated as empty on save, so it
+                // must not silently cancel the removal
+                clear_token: value.trim() !== '' ? false : prev.clear_token,
               }))
             }
             onRequestHeadersChange={value =>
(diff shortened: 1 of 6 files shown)


nicolai-rhesis:
Thanks for the review.

1. Whitespace cancelling a pending removal: fixed in f04f24f5b — `clear_token` is now only cleared when `value.trim() !== ''`, so accidental whitespace no longer preserves a token the user meant to delete.

2. `headersInitial` deps: I left the deps as `[endpoint.request_headers]`. `saveFields` always writes a freshly-parsed `request_headers` object (and a new endpoint object replaces the whole record on refetch), so `headersInitial` already recomputes on every save/exit and re-entry starts from `clear_token: false`. Adding `endpoint.has_auth_token` to the deps is also flagged by `react-hooks/exhaustive-deps` as an unused dependency since it isn't referenced inside the memo, so I avoided introducing a lint warning.

PR #2068, opened 2026-06-30
Remove redundant frontend API calls

EmanueleDeRossi1:
This PR cuts unnecessary frontend API calls caused by two recurring patterns: 
1. pages fetching redundant count-only data the grid already had, and
2.  state updates that re-rendered (and re-fetched) even when nothing actually changed.

**Main fixes:**

- **Removed redundant count fetches** — 7 listing pages (Tests, Test-Runs, Test-Sets, Tasks, Endpoints, Knowledge, Explorer) each had a separate `limit:1` call just to populate a count for the empty state. Now grids report their total count directly via an `onTotalCountChange` callback, so the standalone fetch is gone.

- **Parallelized sequential fetches** — `ModelsPage` ran 5 awaits one after another; now they run via `Promise.all`, saving ~4 round trips. Same fix applied to `ExperimentDetailClient`'s two fetches.

- **Deferred drawer fetches** — `TestDrawer`'s create/update forms were always mounted (and fetching dropdown data) even when closed. Now they only mount when the drawer opens.

- **Fixed navigation re-fetch** — `ActiveProjectContext` was re-ferojects on every route change because `pathname` was an unnecessary dependency. Moved it into a ref so navigation no longer triggers refetches.

- **Short-circuited no-op state updates** — Filter and pagination state setters were creating new object references even when values hadn't changed, which cascaded into unnecessary fetches. Added equality checks so unchanged state doesn't trigger new fetches.

**Net effect:** 7 calls eliminated per cold load, 5 round trips saved on Models via parallelism, and 1 recurring per-navigation call removed.

**Testing:** Confirm pages still load data correctly, no extra project fetches on navigation, and drawer dropdowns populate on open.

### DIFF
--- a/apps/frontend/src/app/(protected)/endpoints/components/EndpointsGrid.tsx
+++ b/apps/frontend/src/app/(protected)/endpoints/components/EndpointsGrid.tsx
@@ -43,6 +43,7 @@ interface EndpointsGridProps {
   sessionToken?: string;
   refreshKey?: number;
   onRefresh?: () => void;
+  onTotalCountChange?: (count: number) => void;
   projectId?: string;
 }
 
@@ -100,6 +101,7 @@ export default function EndpointsGrid({
   sessionToken: sessionTokenProp,
   refreshKey,
   onRefresh,
+  onTotalCountChange,
   projectId,
 }: EndpointsGridProps) {
   const theme = useTheme();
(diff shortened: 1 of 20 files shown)


EmanueleDeRossi1:
Good point. All peqys comments until now refer to the same bug (`onTotalCountChange`)

The bug is broader than the 4 files flagged: `TestsGrid`, `TestSetsGrid`, and `TestRunsGrid` had the same issue. Fixed all 6. `ExplorerGrid` is actually fine since it filters client-side and `onTotalCountChange` always receives the unfiltered count.

PR #2072, opened 2026-06-30
Migrate grids and contexts to React Query

EmanueleDeRossi1:
## Purpose

Every data grid managed its own fetch lifecycle — manual useState for data/loading/error, a useCallback wrapping the API call, a useEffect to trigger it, and three more useEffects to sync search, type filter, and drawer filters into the MUI filter model. The same ~80 lines of boilerplate were copy-pasted across 6 grids. FeaturesContext and PermissionsContext had the same manual useEffect + cancelled flag pattern. This PR replaces all of it with TanStack Query, which was already installed but only used for file queries.

## What Changed

- `constants/query-keys.ts` — centralised key factories for all entities so cache invalidation always matches the query key
- `hooks/useGridState.ts` — shared hook owning filterModel/paginationModel/sortModel and the three sync effects that every grid duplicated
- `hooks/useGridQuery.ts` — thin wrapper around useQuery with keepPreviousData baked in
- `FeaturesContext` / `PermissionsContext` — manual effect loops replaced with useQuery, cancelled flags and intermediate state removed, public APIs unchanged
- All 6 grids — fetch boilerplate replaced with useGridState + useGridQuery, post-mutation fetchX() calls replaced with queryClient.invalidateQueries

## Performance Improvements

Navigating away from a grid and back no longer triggers a loading spinner — results stay cached for 5 minutes and render instantly, with a silent background refetch if stale. Changing page or applying a filter no longer flashes empty rows — keepPreviousData keeps the current rows visible while the next fetch completes. Features and permissions are now cached too, so re-renders that previously triggered duplicate fetches now hit the cache instead.

## Additional Context

- Net ~600 lines deleted across 12 files
- refreshKey props are kept on grids that have them — the effect now calls invalidateQueries instead of the old fetch callback, so callers don't need to change
- fileKeys moved from useFileQueries.ts into constants/query-keys.ts so all cache keys live in one place

## Testing

- TypeScript compiles clean
- Run each grid page and verify data loads, search/filter/sort/pagination all work, and creating/deleting records refreshes the grid

### DIFF
--- a/apps/frontend/src/app/(protected)/endpoints/components/EndpointsGrid.tsx
+++ b/apps/frontend/src/app/(protected)/endpoints/components/EndpointsGrid.tsx
@@ -12,9 +12,8 @@ import { Box, Typography, useTheme, Alert } from '@mui/material';
 import GridBadge from '@/components/common/GridBadge';
 import GridToolbar from '@/components/common/GridToolbar';
 import {
-  GridColDef,
-  GridPaginationModel,
   GridFilterModel,
+  GridColDef,
   GridToolbarColumnsButton,
   GridToolbarDensitySelector,
   GridToolbarExport,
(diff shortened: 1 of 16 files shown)


EmanueleDeRossi1:
About peqy's comments:
- `FeaturesContext`: this was already fixed.
- The other two are marginal concerns not worth to tackle (on the last one: NextAuth invalidates the session on logout, and in practice this app doesn't switch users without a full page reload (which resets the QueryClient).

PR #2082, opened 2026-07-01
Perf: eliminate refreshKey pattern and memoize grid columns

EmanueleDeRossi1:
## Purpose
Reduce unnecessary re-renders and network requests across the frontend.

## What Changed
- Replaced the `refreshKey` prop pattern across all grids with `queryClient.invalidateQueries` — mutations now directly invalidate the relevant React Query cache instead of triggering a cascade of prop/effect re-renders
- Migrated `useComments` and `TasksSection` to React Query (`useQuery`), removing manual `useState`/`useEffect` fetch loops
- Migrated `ExplorerGrid` from a manual fetch to `useQuery`
- Added optimistic updates in `useComments` mutations via `queryClient.setQueryData`
- Memoized column definitions in all major grids with `useMemo` to prevent column array recreation on every render

## Testing
- TypeScript passes (`tsc --noEmit`) with no errors
- Existing grid tests updated to remove now-obsolete refreshKey assertions

### DIFF
--- a/apps/frontend/src/app/(protected)/endpoints/components/EndpointsGrid.tsx
+++ b/apps/frontend/src/app/(protected)/endpoints/components/EndpointsGrid.tsx
@@ -46,8 +46,6 @@ import { useGridQuery } from '@/hooks/useGridQuery';
 
 interface EndpointsGridProps {
   sessionToken?: string;
-  refreshKey?: number;
-  onRefresh?: () => void;
   onTotalCountChange?: (count: number) => void;
   projectId?: string;
 }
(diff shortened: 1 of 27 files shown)


EmanueleDeRossi1:
About peqy's comments: they are all about the same two concerns, now surfacing across more files:

1. **Query key scoping (`taskKeys`, `explorerKeys`, `commentKeys`, etc.)**: Valid in principle, but not exploitable today: every logout path in the app (`client-auth.ts`, `actions/auth.ts`, `DangerZone.tsx`) does a full page navigation, which unmounts the `QueryClient` and wipes the cache for free. _This should be a separate follow-up_ because fixing this properly means scoping every key by user *or* introducing a shared `QueryClient` singleton with an explicit `.clear()` on logout.

2. **`TasksSection` generic error message**: Also valid, but it matches the convention this PR already sets in `TasksGrid`/`EndpointsGrid` (generic message, no `error.message` detail). _Follow-up because_: fixing just this one file would make it inconsistent with its own siblings. It would be better to revisit error-surfacing across all migrated grids at once.

PR #2090, opened 2026-07-02
Fix project-switch 404 UX, Architect errors, and minor frontend polish

nicolai-rhesis:
## Purpose
This PR bundles several frontend and backend fixes discovered while investigating project-scoped detail pages and Architect chat reliability.

1. **Project switch on detail pages** — Switching the active project while viewing an endpoint or task from another project correctly returns 404 from the backend (project scoping via `X-Project-Id`). The frontend was showing a raw API error string instead of the friendly not-found UI.
2. **Architect chat reliability** — Raw Python tracebacks (e.g. `litellm.APIConnectionError` with asyncio lock errors) could appear in the chat UI. Repeat turns in the same Celery worker could fail because `asyncio.run()` closed the event loop while litellm's Vertex AI singleton reused locks from the dead loop.
3. **Minor UI polish** — Knowledge empty state and trace span detail styling.

### Project switch not-found UX
- **Endpoint detail** (`/endpoints/[id]`): call `notFound()` on 404/410 API errors (React Query fetch)
- **Project-nested endpoint detail** (`/projects/[id]/endpoints/[endpointId]`): same `notFound()` handling
- **Task detail** (`/tasks/[id]`): fetch task directly so 404 propagates with `.status`, then render `TaskDetailNotFoundState` with copy mentioning project switching

### Architect error handling and event loop
- Add `format_user_facing_error()` in the SDK to strip tracebacks and map provider failures to actionable messages
- Route Architect and explore Celery tasks through `run_sync()` (persistent background event loop) instead of `asyncio.run()`
- Use `run_sync()` for in-process endpoint invocations from the Architect runner
- Add `reset_litellm_vertex_async_locks()` when the background loop is recreated
- Sanitize stream-end and error events in `WebSocketEventHandler`
- Frontend: show sanitized error text in the chat bubble without a redundant `Error:` prefix

### Other frontend fixes
- **Knowledge page**: use card empty state for consistency
- **Trace span detail**: fix code block corner radius (use `BORDER_RADIUS.xs` tokens instead of numeric `theme.shape.borderRadius` values that MUI sx multiplies)

### Tests
- Add SDK tests for `format_user_facing_error` and `run_sync`
- Update Architect/explore task tests for the new async path
- Fix `test_on_stream_end_with_error` to expect sanitized timeout message

## Additional Context
- Backend project isolation via ORM auto-filter is working as designed; the detail-page bug was frontend-only
- litellm keeps a module-level `vertex_chat_completion` singleton whose `asyncio.Lock` objects bind to the first event loop — `asyncio.run()` per Celery task was the root cause of Architect failures on subsequent messages
- Test runs already used the `notFound()` pattern; endpoints and tasks are now aligned

### Project switch
- [ ] Open an endpoint in Project A, switch to Project B → shared protected 404 page with "belongs to a different project" messaging
- [ ] Same flow for `/projects/{id}/endpoints/{endpointId}`
- [ ] Open a task in Project A, switch to Project B → `TaskDetailNotFoundState` with "Back to Tasks"
- [ ] Confirm genuine network/5xx errors still show the generic error UI

### Architect
- [ ] Restart Celery worker, send two Architect messages in the same session — second turn should not show event-loop errors
- [ ] On provider failure, chat shows a friendly retry message instead of a Python traceback

### Automated
- [x] Backend tests, frontend unit, SDK tests, and E2E all pass on CI

### DIFF
--- a/apps/backend/src/rhesis/backend/app/services/architect/event_handler.py
+++ b/apps/backend/src/rhesis/backend/app/services/architect/event_handler.py
@@ -9,6 +9,7 @@ from rhesis.backend.app.schemas.websocket import (
     WebSocketMessage,
 )
 from rhesis.backend.app.services.websocket.publisher import publish_event
+from rhesis.sdk.agents.errors import format_user_facing_error
 
 logger = logging.getLogger(__name__)
 
@@ -134,7 +135,11 @@ class WebSocketEventHandler:
         error: Optional[str] = None,
         **kw: Any,
     ) -> None:
-        self.publish(EventType.ARCHITECT_STREAM_END, {"content": content, "error": error})
+        friendly_error = format_user_facing_error(error) if error else None
+        self.publish(
+            EventType.ARCHITECT_STREAM_END,
+            {"content": content, "error": friendly_error},
+        )
 
     async def on_agent_end(self, *, result: Any, **kw: Any) -> None:
         pass  # handled after chat_async returns
@@ -142,5 +147,8 @@ class WebSocketEventHandler:
     async def on_error(self, *, error: Exception, **kw: Any) -> None:
         self.publish(
             EventType.ARCHITECT_ERROR,
-            {"error": str(error), "error_type": type(error).__name__},
+            {
+                "error": format_user_facing_error(error),
+                "error_type": type(error).__name__,
+            },
         )
--- a/apps/backend/src/rhesis/backend/app/services/architect/runner.py
+++ b/apps/backend/src/rhesis/backend/app/services/architect/runner.py
@@ -7,7 +7,6 @@ constructing an :class:`~rhesis.sdk.context.EndpointContext` to carry
 tenant identity.
 """
 
-import asyncio
 import logging
 from typing import Any, Dict, List, Optional
 from uuid import UUID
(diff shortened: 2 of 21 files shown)


harry-rhesis (approved):
Nice work, thank you! I reviewed the backend and SDK changes closely and they look solid.

The `asyncio.run()` to `run_sync()` migration is the right fix for the Vertex AI lock issue. litellm's module-level singleton binds its locks to the first event loop, so a persistent background loop avoids the lifecycle mismatch across Celery task turns.

The error sanitization via `format_user_facing_error()` is clean. Raw tracebacks stay in server logs (via `logger.error` with `exc_info=True`) and only friendly messages reach the frontend. The keyword matching covers the important provider failure modes with a sensible fallback.

A couple of minor observations (not blockers):

- In `base.py`, wrapping the error as `RuntimeError(format_user_facing_error(e))` before emitting means `on_error` in `event_handler.py` applies `format_user_facing_error` a second time. The second pass is a no-op on already-clean strings, so it's harmless, just slightly redundant.
- That same wrapping means `error_type` in the WebSocket payload will always be `"RuntimeError"` rather than the original exception class name. Doesn't affect users, though the debugging signal is still available in server logs.
- `reset_litellm_vertex_async_locks()` touches private litellm internals that could shift across versions, but the `try/except` with `logger.debug` handles that gracefully.

Tests look thorough. Ship it!

harry-rhesis (approved):

Issue #2070, opened 2026-06-30
Improve consistency and DRY across Penelope targets and SDK telemetry integrations


harry-rhesis:
## Summary

Address cross-cutting consistency gaps and DRY opportunities across the Penelope target implementations and SDK telemetry integrations, identified during the review of PR #2057.

## Background / Context

PR #2057 (Pydantic AI integration) revealed several inconsistencies and duplication patterns that exist across all targets and integrations, not just Pydantic AI. These are best addressed in a separate PR to keep #2057 focused.

## Deliverables

- [ ] Add `try/except ImportError` guards in `penelope/src/rhesis/penelope/targets/__init__.py` so that missing optional dependencies (langchain, langgraph, pydantic-ai) degrade gracefully instead of breaking the entire import
- [ ] Add native `a_send_message()` to `LangChainTarget` and `LangGraphTarget` (using `ainvoke` / `arun`) instead of relying on the base class thread-pool fallback, matching PydanticAITarget's pattern
- [ ] Extract `PatchState` base class or factory from the duplicated `AgentPatchState` (pydantic_ai.py) and `GraphPatchState` (langgraph.py) into `integrations/base.py`
- [ ] Add LLM-level child spans (`ai.llm.invoke`) inside Pydantic AI agent runs, matching the visibility LangChain/LangGraph provide for individual model calls within an agent run

## Acceptance Criteria

- [ ] `import rhesis.penelope.targets` succeeds on a minimal install without langchain, langgraph, or pydantic-ai
- [ ] `LangChainTarget` and `LangGraphTarget` have native async paths that do not use `asyncio.to_thread`
- [ ] `GraphPatchState` and `AgentPatchState` share a common base, reducing code duplication
- [ ] Pydantic AI agent runs produce `ai.llm.invoke` child spans for individual model calls

## Additional Context

Related to PR #2057: https://github.com/rhesis-ai/rhesis/pull/2057

rhiannalitchfield:
Opened https://github.com/rhesis-ai/rhesis/pull/2115 to close this out. Accounting for the original deliverables, since two were reshaped by the #2104 translator migration:

- **Import guards**: already satisfied by the deferred-import pattern from #2057; the PR extends the regression test to also cover agent_framework/MAFTarget.
- **Native async for LangChainTarget/LangGraphTarget**: delivered in the PR via native ainvoke() plus an async file-attachment path.
[...]

PR #2104, opened 2026-07-06
Migrate Pydantic AI integration to native instrumentation translator

rhiannalitchfield:
## Purpose

Follow-up to #2057, implementing the translator migration agreed there with @harry-rhesis: replace the `Agent.run` monkey-patch with Pydantic AI's built-in OpenTelemetry instrumentation plus a `gen_ai.*` -> `ai.*` translating exporter, mirroring the MAF integration. Covers the deferred review items #1 (`run_stream`), #2 (tool spans), #3 (agent handoffs), and #6 (multi-agent example).

## What Changed

- **New shared module `integrations/genai.py`** (the "one GenAI translator" coordination for #2070): the framework-neutral pieces of the `gen_ai.*` -> `ai.*` bridge extracted from the MAF integration — attribute/event translation tables, message-event synthesis from the attribute-based GenAI convention, the `TranslatedSpan` read-only span view, exporter-swapping helpers, and the `RHESIS_DISABLE_CONTENT_CAPTURE` env var. `agent_framework` re-imports everything under its historical names, so its public API is unchanged (all 90 MAF tests pass unmodified).
- **`integrations/pydantic_ai/` package** replacing the monkey-patch module:
  - `enable()` calls `Agent.instrument_all(InstrumentationSettings(...))` pinned to instrumentation version 5 (verified empirically against pydantic-ai 2.0.0), with `include_binary_content=False` so attachment bytes can never reach spans, and content capture honoring the shared opt-out env var. No `logfire` dependency — spans flow into the global `TracerProvider` that `RhesisClient` configures, and `enable()` verifies that provider is in place first (returns `False` otherwise, like MAF).
  - `PydanticAITranslatingExporter` wraps the existing exporter and rewrites spans on export: `ai.agent.invoke` (agent name, aggregated tokens, prompt/completion events from `pydantic_ai.all_messages`/`final_result`), per-model-call `ai.llm.invoke` with tokens and message events, `ai.tool.invoke` with input/output events. Unknown operations fall back to `function.pydantic_ai.*` so nothing fails backend span-name validation.
  - **Handoffs**: Pydantic AI has no handoff primitive — delegation is one agent called inside another's tool. The trace nests naturally (`invoke_agent` (child) under `execute_tool` under `invoke_agent` (parent)), and the exporter additionally synthesizes an `ai.agent.handoff` span with `ai.agent.handoff.from`/`.to` whenever a run's ancestor chain contains a different agent, resolved via a span-ancestry registry recorded at span start so it survives export batching (same approach as MAF).
  - `PydanticAILLMDedupSpanProcessor` records ancestry and toggles the LLM-observation flag during chat spans, so flag-checking instrumentations don't emit duplicate `ai.llm.invoke` spans.
  - `disable()` uses Pydantic AI's public off switch (`Agent.instrument_all(False)`) and unwraps the exporters.
- **Examples**: `pydantic_ai_example.py` updated (tools now traced automatically, no `@observe` wrappers); new `pydantic_ai_multi_agent_example.py` showing coordinator -> specialist delegation with handoff edges; README section refreshed.

## Additional Context

- Implements the plan agreed in the #2057 thread; sequenced before #2070 as discussed there.
- For #2070: `AgentPatchState` is gone, so the `PatchState` base-class extraction no longer applies to this integration, and the "`ai.llm.invoke` child spans for Pydantic AI" item is delivered here by the native chat spans. The shared `genai.py` module is the suggested home for any further GenAI-translation consolidation.
- The duplicate-span risk raised in #2057 (native instrumentation + patch both writing to the same provider) is resolved by construction: the patch no longer exists.
- Span shapes were verified empirically (scope `pydantic-ai`, operations `chat`/`invoke_agent`/`execute_tool`, message attributes `gen_ai.input.messages`/`gen_ai.output.messages`, agent-span `gen_ai.aggregated_usage.*`) rather than taken from docs.

## Testing

- `uv run pytest tests/sdk/telemetry/integrations/` — 157 passed (37 rewritten Pydantic AI tests + 90 MAF tests unchanged + others)
- Full regression: `uv run pytest tests/sdk --ignore=tests/sdk/integration` (2302 passed; the 5 failures are the known timing-flaky `connector/test_executor*` and `test_exporter` deadline tests, unrelated to this change) and `uv run pytest tests/penelope` (587 passed)
- New coverage: end-to-end runs against `TestModel` asserting agent/LLM/tool/handoff spans and events, `run_stream`, span-name validation for every exported span, binary-content exclusion, content-capture opt-out, error status, enable/disable lifecycle, exporter unwrap, and the dedup processor
- `ruff check` / `ruff format` clean on all new/modified files


rhiannalitchfield:
Both addressed in c48b75cd:

**1. Dedup flag stuck True** — confirmed real. `on_start` toggled based on the span *name* (`chat ` prefix, since attributes aren't assigned yet at start time) while `on_end` restored based on the *attribute* (`gen_ai.operation.name`), so any chat-named span that ended without that attribute would leak the flag as True for the rest of the context. Went with your first suggestion, which is the provably balanced one: `on_end` now restores based on `_prev_flags` membership — the dict only ever contains entries for spans whose start hook toggled the flag, so membership is the exact restore condition, with no attribute re-derivation at all. Regression test added (`test_flag_restored_even_without_operation_attribute`).

Worth noting: `MAFLLMDedupSpanProcessor` has the same start-by-name / end-by-attribute asymmetry. I left it untouched here to keep this PR scoped to Pydantic AI, but it should get the same fix — happy to fold it into the #2070 cleanup.

**2. enable() with no wrappable exporter** — agreed, and went further than logging: `_wrap_existing_exporters` now returns a boolean, wrapping happens *before* `instrument_all()`, and `enable()` fails closed (returns False) when translation cannot be installed — enabling instrumentation whose spans would all be dropped by backend span-name validation only produces broken traces. A failed `instrument_all` also reverts any wrapping already installed, so a failed enable leaves the provider exactly as it found it. The introspection-failure path now logs at warning level. Regression test added (`test_enable_fails_without_wrappable_exporter`).

This makes the pydantic_ai integration slightly stricter than MAF (which warns but still returns True in the same situation); aligning MAF is another #2070 candidate.

159 integration tests passing (both new regressions included).

harry-rhesis:
Thanks for the excellent work here, Rhianna. Reviewed the migration end to end: checked the genai.py extraction for backward compatibility with the MAF integration, verified both Peqy findings against the hardening commit, and ran the full telemetry test suite locally (372 passed). The architecture is clean and mirrors the existing MAF integration well.

One small thing we caught in review: the translating exporter could append a duplicate fallback span if handoff synthesis failed after the primary span was already translated, in both the MAF and Pydantic AI translators. Pushed a fix for that in commit 3d28e663d, which separates the two try blocks so a handoff synthesis failure can no longer duplicate the primary span. Verified with the existing test suite plus a manual check of both files.

One follow-up worth tracking separately, not blocking: the handoff ancestry tracking (`_AncestryRegistry` in pydantic_ai/translator.py and `_HandoffAncestryRegistry` in agent_framework/translator.py) still duplicates a fair amount of logic that could move into genai.py, following the same consolidation this PR already did for the attribute and event translation helpers. MAF's version has an extra sibling layout case Pydantic AI doesn't need, so it's not a pure copy paste, but worth a look in a future pass.

Great work overall. Merging this now.

rhiannalitchfield:
Thanks for the thorough review and for catching the duplicate-fallback bug, that try-block scoping was my miss and the fix is exactly right. Agreed on consolidating the two ancestry registries into genai.py; I'll fold that into the #2070 pass along with the MAF dedup-flag and fail-closed alignments flagged during review.

PR #2115, opened 2026-07-08
Improve consistency and DRY across targets and telemetry integrations

### LINKED ISSUE #2070: Improve consistency and DRY across Penelope targets and SDK telemetry integrations

rhiannalitchfield:
## Purpose

Closes #2070. The cross-cutting consistency and DRY pass across Penelope targets and SDK telemetry integrations, sequenced after the translator migration (#2104) as agreed on #2057 — two of the original deliverables were reshaped by that work (details below).

## What Changed

- **Native async paths for `LangChainTarget` and `LangGraphTarget`**: `a_send_message()` built on the runnable/graph's native `ainvoke()` instead of the base-class `asyncio.to_thread` fallback, matching `PydanticAITarget` and `MAFTarget`. `files_to_content_blocks()` gains an async sibling that materializes `FileReference` attachments via `aread_bytes()` so object-storage fetches don't block the event loop; sync and async share the block-building logic.
- **Shared `AncestryRegistry` in `genai.py`** (requested by @harry-rhesis in the #2104 review): the handoff from-agent resolution registry moves out of both translators. MAF's sibling-layout case (`HandoffBuilder` puts `invoke_agent` and `chat` as siblings under `executor.process`) is behind an opt-in `track_siblings` flag — it must stay off for properly nested delegation like Pydantic AI's, where a delegated agent span would otherwise resolve *itself* through its own parent entry and report its own name as the calling agent. `_HandoffAncestryRegistry` becomes a thin subclass with unchanged behavior.
- **MAF hardening parity** (ports of the #2104 fixes): `MAFLLMDedupSpanProcessor.on_end` now restores the LLM-observation flag from what `on_start` recorded rather than re-deriving chat-ness from attributes (no more stuck-True leak), and `enable()` wraps exporters *before* turning on instrumentation, failing closed when none can be wrapped. The ordering fix matters more for MAF than it did for pydantic_ai: MAF has no off switch, so previously a failed provider check still left instrumentation permanently on.

## Original #2070 deliverables accounting

- **Import guards**: already satisfied by the deferred-import pattern from #2057 (0517dc3) — `import rhesis.penelope.targets` succeeds without any optional dep. This PR extends the regression test to also block `agent_framework` and assert `MAFTarget` imports cleanly.
- **Native async for LangChain/LangGraph**: delivered here.
- **`PatchState` base extraction**: moot — `AgentPatchState` was deleted by the #2104 migration, leaving `GraphPatchState` (langgraph) as the only patch-state class, so there is no duplication left to extract. If langgraph later migrates to a translator too, it goes away entirely.
- **`ai.llm.invoke` child spans for Pydantic AI**: delivered by #2104's native chat spans.

## Testing

- `uv run pytest tests/sdk/telemetry/integrations/` — 161 passed (2 new MAF regressions)
- `uv run pytest tests/penelope` — 594 passed (7 new: async paths for both targets, aread_bytes usage, error capture, session state, MAF import guard)
- Full SDK regression: 2308 passed; the 3 failures are the known timing-flaky `connector/test_executor*` and `test_exporter` deadline tests, unrelated to this change
- `ruff check` / `ruff format` clean on all new/modified files

### DIFF
--- a/penelope/src/rhesis/penelope/targets/_content_blocks.py
+++ b/penelope/src/rhesis/penelope/targets/_content_blocks.py
@@ -27,25 +27,41 @@ def files_to_content_blocks(message: str, files: Optional[List[Any]]) -> Union[s
     return [create_text_block(message), *(_file_to_block(f) for f in files)]
 
 
-def _file_to_block(file: Any) -> Any:
+async def afiles_to_content_blocks(
+    message: str, files: Optional[List[Any]]
+) -> Union[str, List[Any]]:
+    """Async sibling of :func:`files_to_content_blocks`.
+
+    Uses ``FileReference.aread_bytes()`` to materialize object-storage
+    attachments so building the blocks doesn't block the event loop.
+    """
+    if not files:
+        return message
+
+    from langchain_core.messages.content import create_text_block
+
+    return [create_text_block(message), *[await _afile_to_block(f) for f in files]]
+
+
+def _extracted_text_block(file: Any, extracted_text: str) -> Any:
+    from langchain_core.messages.content import create_text_block
+
+    from rhesis.penelope._file_compat import file_attr
+
+    filename = file_attr(file, "filename", "attachment")
+    return create_text_block(f"[Attached file: {filename}]\n{extracted_text}")
+
+
+def _block_from_bytes(data: bytes, content_type: str) -> Any:
     import base64
 
     from langchain_core.messages.content import (
         create_audio_block,
         create_file_block,
         create_image_block,
-        create_text_block,
         create_video_block,
     )
 
-    from rhesis.penelope._file_compat import file_attr, file_bytes_and_type, file_extracted_text
-
-    extracted_text = file_extracted_text(file)
-    if extracted_text:
-        filename = file_attr(file, "filename", "attachment")
-        return create_text_block(f"[Attached file: {filename}]\n{extracted_text}")
-
-    data, content_type = file_bytes_and_type(file)
     kwargs = {"base64": base64.b64encode(data).decode(), "mime_type": content_type}
     if content_type.startswith("image/"):
         return create_image_block(**kwargs)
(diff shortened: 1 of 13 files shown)


rhiannalitchfield:
All three addressed in aa14d6fd, and the first one turned out to be deeper than it looked:

**1. `ainvoke` not validated** — confirmed, and digging into it surfaced a latent pre-existing bug. `target_interaction.a_execute()` calls `target.a_send_message()` unconditionally, but when `Target` moved from the Penelope base class to `rhesis.sdk.targets`, the `a_send_message()` thread-pool fallback did not move with it (the `rhesis.penelope.targets.base` shim still has it; the canonical SDK class did not). So async execution against *any* sync-only target has been an `AttributeError` on main all along, independent of this PR. Fix: restored the fallback on `rhesis.sdk.targets.Target`, and `LangChainTarget`/`LangGraphTarget` now delegate to it via `super()` when a duck-typed runnable/graph lacks `ainvoke()`. Regression tests for both targets (sync-only fakes go through the fallback and succeed).

**2. Sequential attachment reads** — agreed; `afiles_to_content_blocks()` now uses `asyncio.gather` (order preserved) so object-storage fetches run concurrently.

**3. Introspection-failure question** — yes, failing closed there is the intended tradeoff: if we can't see the processors we can't guarantee translation, and untranslated GenAI span names get dropped by backend validation anyway, so "refuse and say why" beats "enable and silently lose spans". Took your suggestion and clarified the warning in both integrations to state that instrumentation is being refused for that reason.

Penelope suite: 596 passed. Full SDK regression: 2308 passed (same 3 known timing flakes).

harry-rhesis:
Thanks for this, @rhiannalitchfield, really nice work!

The native async paths for LangChain/LangGraph, the shared `AncestryRegistry`, and the MAF hardening (flag-restore fix and fail-closed exporter wrapping) are all clean, well-documented, and well-tested. Also appreciated the base `Target.a_send_message` fix, that was a real latent `AttributeError` for any target without a native async override.

Merging this now.

PR #2117, opened 2026-07-08
perf(backend): split get_items_detail into id + join queries

EmanueleDeRossi1:
## Purpose
`/tests` (and every other list endpoint using `get_items_detail`) was joining ~13 tables, sorting, and applying LIMIT in a single query — meaning Postgres had to build every eager-loaded join for every matching row in the org before it could cut down to the requested page. For an org with ~1,470 tests, that's the difference between joining 1,470 rows and joining 25.

## What Changed
- `QueryBuilder.ids()` (`query_utils.py`): runs the built query with no eager-load joins, returning only matching IDs in sort order.
- `get_items_detail` (`crud_utils.py`): now runs as two queries — a joinless query resolves the page's IDs (filter + sort + LIMIT/OFFSET), then a second query eager-loads relationships scoped to `WHERE id IN (...)`. Re-orders results in Python since `IN (...)` doesn't preserve order. Phase 2 re-applies both `organization_filter` and `visibility_filter` for defense-in-depth, matching phase 1.
- `Makefile`: fixed `DIFF_FILES` to filter out paths deleted within the same branch (`wildcard`) — previously crashed `ruff format` on a file that was added and later removed in this branch's own history.

## Additional Context
Builds on #2110 (merged). Stacked discussion: this was deliberately kept separate from that N+1 fix PR since it's a more structural pagination change, to keep both reviewable independently.

A temporary request-vs-DB timing probe (`debug_request_timing.py`, env-gated behind `RHESIS_ENABLE_REQUEST_TIMING`) was added and used to measure the latency gap live on dev, then removed once diagnosis confirmed the split brought `/tests` in line with comparable list endpoints — same convention as the earlier (already-reverted) `debug_sql_timing.py` probe. It's not part of the final diff.

**Known limitation (documented, not fixed here):** `QueryBuilder.ids()` has no `DISTINCT`. An OData filter that joins a to-many relation (e.g. `test_sets/name eq 'x'`) can multiply matching rows before `LIMIT`, producing fewer than `limit` distinct IDs on some pages. This is a pre-existing property of `get_items_detail`'s pagination model, not something this PR introduces — the original single-query implementation has the identical multiply-then-`LIMIT` behavior for the same filters. A real fix needs a `DISTINCT`-safe rewrite (Postgres requires `ORDER BY` columns to appear in the `SELECT` list under `DISTINCT`, which conflicts with selecting only `id`), left for a follow-up.

## Testing
- New unit tests in `tests/backend/utils/test_get_items_detail_pagination_split.py`: sort order preserved across pages, empty-filter result, and a SQL-statement assertion that the join query is scoped to `id IN (...)` rather than the full table.
- Ran `tests/backend/utils/`, `tests/backend/crud/`, `tests/backend/security/` (1367 passed, 3 pre-existing skips) plus the `Test` model's crud/route tests — no regressions.
- Verified live on dev: confirmed via the (since-removed) timing middleware and browser network timing that `/tests` response time dropped from the originally reported ~2.2s total to ~500-700ms, now comparable to peer list endpoints (`test_sets` ~430ms, `behaviors` ~629ms).


EmanueleDeRossi1:
I agree with last peqy's comment, but would differ this to a separate PR. 

To notice, the original single-query get_items_detail has the identical multiply-then-LIMIT problem for the same filters. So this PR doesn't introduce it, but carries the same failure mode into the phase-1 query.

PR #2118, opened 2026-07-08
Fix onboarding lockout: seed org creator's RBAC role earlier

harry-rhesis:
## Purpose

Under RBAC, a freshly onboarding user hit multiple "Unauthorized"/403 errors during onboarding — reported live on dev-app.rhesis.ai. Two distinct bugs, both boiling down to "code assumed the org exists before it actually does":

1. **Backend RBAC role ordering**: the org creator's Owner `organization_member` role was only assigned inside `load-initial-data`, the *last* onboarding step, but every earlier onboarding call (attaching the org via `PUT /users/{user_id}`, inviting teammates via `POST /users/`, the frontend's org-context fetch) is capability-gated on that same role already existing. The creator could never reach the step that would grant them the role.
2. **Frontend org-scoped fetches with no guard**: two globally-mounted contexts (`OnboardingContext`'s DB-backed progress sync, `ActiveProjectContext`'s project list fetch) called org-scoped endpoints (`/users/settings`, `/projects/mine`) without checking whether `session.user.organization_id` was actually populated yet, so they 403'd with "User is not associated with an organization" during the legitimate pre-org window of onboarding, and in the stale-session window right after onboarding completes.

## What Changed

**Backend:**
- Seed the Owner/Member role in `PUT /users/{user_id}` at the moment `organization_id` is first set on the user, instead of deferring it to `load-initial-data`. This closes the gap for all downstream onboarding calls in one place (invite teammates, `load-initial-data`, and the latent `PUT /organizations/{id}` case) rather than patching each capability-gated endpoint individually.
- The existing call in `load-initial-data` is left in place as a harmless idempotent safety net for any flow that doesn't go through `PUT /users/{user_id}`.

**Frontend:**
- `OnboardingContext`'s load/sync/force-sync paths now gate on `session.user.organization_id`, matching the pattern `OrganizationContext` already uses, instead of firing unconditionally and logging a 403 as a real error.
- `ActiveProjectContext.fetchProjects` now also gates on `session.user.organization_id`, not just the `/onboarding` pathname — closing the window right after onboarding completes where a stale client-side session can still read `organization_id: null`.

**Tests:**
- Two router-level tests proving the RBAC role is seeded exactly when `organization_id` transitions from `None`, and not on ordinary profile updates.
- One full end-to-end HTTP test (`TestClient`, RBAC forced on) driving the real onboarding sequence — attach org, invite teammate, load initial data — proving none of them 403.

## Additional Context

Root-caused from a live "Unauthorized" report during onboarding on dev-app.rhesis.ai. Community-tier auth has an `organization.owner_id` bypass that masked bug #1; the EE `PermissionAuthorizationProvider` (used once RBAC is licensed for an org) has no such bypass and strictly requires the `organization_member` row, which is why this was invisible to existing tests — none of them combined "RBAC actually enabled" with the real onboarding call sequence. Bug #2 was found via live console logs during manual onboarding testing after the backend fix was already deployed.

## Testing

- `RHESIS_SKIP_MIGRATIONS=1 uv run pytest ../../tests/backend/ee/rbac/test_default_org_role.py -v` — 12 passed
- Verified the new end-to-end test fails on the pre-fix code (403 on the invite step) and passes with the fix
- `uvx ruff check` / `uvx ruff format --check` clean on all touched backend files
- `tsc --noEmit` and `eslint` clean on both touched frontend files

### DIFF
--- a/apps/backend/src/rhesis/backend/app/routers/user.py
+++ b/apps/backend/src/rhesis/backend/app/routers/user.py
@@ -402,8 +402,25 @@ def update_user(
         if not authorize(principal, Permission.Member.MANAGE, project_id=None, db=db):
             raise HTTPException(status_code=403, detail="Not authorized to update this user")
 
+    had_no_organization = db_user.organization_id is None
+
     # Update the user
     updated_user = crud.update_user(db, user_id=user_id, user=user)
+
+    # Joining an org for the first time (e.g. the creator attaching to their own
+    # org during onboarding) — seed the default RBAC role now, not later. Every
+    # subsequent onboarding call (invite teammates, load-initial-data, the
+    # frontend's org-context fetch) is capability-gated on this role existing;
+    # deferring assignment locks the creator out of their own onboarding flow
+    # once RBAC is enabled. This route runs on a plain (non-tenant) session, so
+    # session variables must be set explicitly — organization_member RLS
+    # requires app.current_organization for the INSERT to succeed.
+    if had_no_organization and updated_user.organization_id is not None:
+        from rhesis.backend.app.auth.org_membership_hook import on_user_org_assigned
+        from rhesis.backend.app.database import set_session_variables
+
+        set_session_variables(db, str(updated_user.organization_id), str(updated_user.id))
+        on_user_org_assigned(db, updated_user.id, updated_user.organization_id)
 
     # If this is the current user being updated, refresh their session token
     if str(updated_user.id) == str(current_user.id):
--- a/apps/backend/src/rhesis/backend/app/routers/user.py
+++ b/apps/backend/src/rhesis/backend/app/routers/user.py
@@ -404,6 +404,30 @@ def update_user(
 
     had_no_organization = db_user.organization_id is None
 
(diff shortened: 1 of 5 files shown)


harry-rhesis:
Addressed the peqy review findings:

- **Critical (`routers/user.py:418`)**: valid, and traced further than the comment scope — this was a pre-existing gap (predates this PR) where `PUT /users/{user_id}` accepted a client-supplied `organization_id` with no ownership validation, and a self-update returns a freshly minted session token, so any orgless user could self-assign an arbitrary org and get immediate ambient tenant-scope access to it. Fixed in aa45553c1: only the org-creator-attaching-to-their-own-org case is now allowed; everything else (unowned org, reassignment while already in an org, on another user's behalf) is rejected with 403. Verified this is the only case any current frontend caller relies on.
- **Two test-assertion comments**: valid, tightened `!= 403` to exact `== 200` in aa45553c1 so the onboarding-sequence test proves success, not just "not forbidden."

Added two new regression tests (`TestUpdateUserRejectsIllegitimateOrgAssignment`) covering the rejected cases.

PR #2120, opened 2026-07-09
Unify Rhesis agent skill with PRD workflow, metric scope, and tag tools

nicolai-rhesis:
## What

One agent skill (`/rhesis`) that turns product requirements into a ready-to-run test foundation in Rhesis — behaviors, metrics, test sets, and tags — instead of setting that up by hand.

It also fixes a silent failure: if you pair the wrong metric type with the wrong test type (single-turn vs multi-turn), Rhesis drops those metrics at run time with no error. This PR makes the agent catch that before anything is created.

## Why

- **Faster setup** — Paste a PRD or describe a feature; the agent proposes a plan you approve.
- **Less guesswork** — Clear paths for quick check, full exploration, PRD foundation, or run/analyze.
- **Fewer broken runs** — Metric scope is validated up front.
- **One source of truth** — Skill, Telemachus, and docs share the same reference files instead of duplicating platform knowledge.

---

## How the reference system works

Platform knowledge is split into **three layers**. Each layer has a job; agents should not mix them up.

### Layer 1 — Human docs + machine index (`docs.rhesis.ai`)

For humans and coding agents with network access.

| Surface | Role |
|---|---|
| **`llms.txt`** | Machine index — fetch `.md` URLs, not HTML |
| **Agent reference** (`/docs/agent-skill/for-agents`) | Routing table, common confusions, plan table templates |
| **Glossary** (`/glossary`, 75 terms) | Canonical vocabulary — `glossary/behavior.md`, `glossary/metric-scope.md`, etc. |
| **Metric scope** (`/docs/metrics/metric-scope`) | Product doc for Single-Turn vs Multi-Turn alignment |
| **PRD workflow** (`/docs/agent-skill/prd`) | Requirements → test foundation pipeline |

Glossary is canonical for *what terms mean*. Agent reference holds *how to route and plan*. Repo references hold *operational detail* (tool chains, field constraints, golden examples).

### Layer 2 — Repo skill references (`skills/rhesis/references/`)

Operational knowledge for MCP agents and Telemachus. **Single source of truth for runtime behavior** — docs link here; Telemachus includes these files directly.

```
skills/rhesis/
├── SKILL.md                 # Thin router (~90 lines) — intake, menu, pointer to refs + docs
└── references/
    ├── workflow-index.md    # Read first — route by intent, reference catalog
    ├── definitions.md       # Confusions + PRD traceability only (glossary first)
    ├── entity-model.md      # Entity graph, relations, tool chains
    ├── metric-scope.md      # Scope rules + link to docs/glossary
    ├── prd-workflow.md      # PRD → behaviors/metrics/test sets pipeline
    ├── use-case-bracketfeld.md  # Golden PRD plan shape (fictional; repo-only)
    ├── tool-catalog.md      # MCP tool listing + chains
    ├── odata-patterns.md    # $filter, $select, name resolution
    ├── exploration-strategies.md
    ├── result-analysis.md
    ├── telemachus-parity.md # Skill ↔ Architect parity checklist
    └── phases/
        ├── discovery.md
        ├── planning.md
        ├── creation.md
        ├── execution.md
        ├── analysis.md
        └── direct-requests.md
```

**Routing flow (coding agents):**

1. `SKILL.md` → detect intent or show four-path menu
2. `workflow-index.md` → pick reference files for this request
3. Read only what's needed (e.g. PRD path: `prd-workflow.md` → `use-case-bracketfeld.md` → `metric-scope.md`)
4. Terminology: fetch [glossary](https://docs.rhesis.ai/glossary.md) first; `definitions.md` only when concepts are mixed up
5. **Write gate:** no `create_*` / `generate_*` until user approves the plan

**Shared skeleton** (every full workflow):

```
resolve by name → list_behaviors + list_metrics once → plan → user approval
  → create in order → optional execute → analyze
```

Creation order: behaviors → metrics (with `metric_scope`) → mappings → tags → `generate_test_set` → verify.

### Layer 3 — Telemachus runtime (Architect in the app)

Telemachus uses the **same `skills/rhesis/references/` files** as the coding skill, loaded via Jinja — not a separate prompt corpus.

**`prompt_loader.py`** builds a Jinja `ChoiceLoader` over:
- `sdk/.../prompt_templates/` (Architect-specific `.j2` partials)
- `skills/rhesis/references/` (shared markdown refs)

Override path: `RHESIS_SKILLS_REFERENCES=/custom/path`.

#### Fixed system prompt (~289 lines)

Loaded once per session. Always includes:

| Include | Purpose |
|---|---|
| `workflow-routing.j2` | Four-path menu, write gate, context detection |
| `entity-model.md` | Entity graph — every agent needs this |
| `telemachus-resolution.j2` | Resolve entities by name, never ask for IDs |
| `telemachus-security.j2` | Boundaries, no schema leakage |

Does **not** include phase files, PRD workflow, or metric-scope detail — those load per turn.

#### Lazy phase loading (per iteration)

`iteration_prompt.j2` injects a **Phase Guidance** block based on `AgentMode` + `WorkflowPath`:

| Mode | Workflow path | References loaded |
|---|---|---|
| Discovery | Explore | `phases/discovery.md`, `exploration-strategies.md`, `odata-patterns.md` |
| Discovery | PRD | `prd-workflow.md` |
| Discovery | Run/analyze | `phases/execution.md`, `phases/analysis.md`, `result-analysis.md` |
| Discovery | Direct | `phases/direct-requests.md`, `odata-patterns.md` |
| Discovery | Unset | *(routing only — system prompt)* |
| Planning | * | `phases/planning.md`, `metric-scope.md`, `telemachus-save-plan.j2`, `telemachus-reuse.j2`, `telemachus-guidelines.j2` + `prd-workflow.md` if PRD |
| Creating | * | `phases/creation.md`, `telemachus-creation-order.j2`, `metric-scope.md` + `prd-workflow.md` if PRD |
| Executing | * | `phases/execution.md`, `phases/analysis.md`, `result-analysis.md` |

`workflow.py` infers `WorkflowPath` from the user message (PRD signals, menu digits 1–4, run/analyze keywords, direct commands). Path is persisted in `ArchitectAgentStateSnapshot`.

**Token impact:**

| Prompt | Lines (rendered) |
|---|---|
| Old monolithic system prompt | ~641 |
| Old eager load (all refs in system) | ~1,402 |
| **New fixed system prompt** | **~289** |
| Per-turn phase block (discovery → planning → creating) | +370–530 |

---

## Parity: skill vs Telemachus

| | Coding agent (Cursor, Claude Code) | Telemachus (Architect UI) |
|---|---|---|
| Entry | `SKILL.md` + `workflow-index.md` | `system_prompt.j2` + per-turn phase block |
| References | Agent reads `references/` from repo | Jinja `{% include %}` same files |
| Docs | Fetches `llms.txt` + glossary `.md` URLs | Repo refs embed docs URLs at top of key files |
| Tools | Rhesis MCP over HTTP | `LocalToolProvider` + `mcp_tools.yaml` |
| Plan approval | User says yes in chat | `save_plan` + Accept/Change UI |
| Scope guard | Prompt + plan tables | `MetricSpec.metric_scope` + `save_plan` validation |

Checklist: `skills/rhesis/references/telemachus-parity.md`

---

### Agent skill
- Merged `rhesis-prd` into single `rhesis` skill; four-path menu
- Slim `SKILL.md`; `workflow-index.md` as routing hub
- `definitions.md` trimmed — glossary canonical; file keeps confusions + FR/AC traceability
- Tag tools documented in `tool-catalog.md`
- Golden example `use-case-bracketfeld.md` stays repo-only (not duplicated in docs)

### Architect / SDK
- `prompt_loader.py` — shared ref loader + `render_phase_knowledge()`
- Slim `system_prompt.j2`; phase refs in `iteration_prompt.j2`
- `workflow.py` — `WorkflowPath` inference + persistence
- `plan.py` — `MetricSpec.metric_scope`; coverage validation on `save_plan`
- Tests: `test_architect_prompt_loader.py`, `test_plan_metric_scope.py`

### Backend / MCP
- `list_tags`, `assign_tag` in `mcp_tools.yaml`
- `GET /tags/` OData `$select` support

### Docs + glossary
- `agent-skill/` pages, `metric-scope.mdx`, agent reference
- `llms.txt` **Agent reference** section
- Glossary fixes: `test-result` entity corrected, silent `metric_scope` drop documented, MCP disambiguation, new `tag` term (75 total)

---

## Test plan

- [ ] `cd sdk && uv run pytest ../tests/sdk/agents/test_plan_metric_scope.py ../tests/sdk/agents/test_architect_prompt_loader.py -v`
- [ ] `cd apps/backend && uv run pytest ../../tests/backend/routes/test_tag.py::TestTagRoutes::test_list_tags_with_select -v`
- [ ] `cd apps/backend && uv run pytest ../../tests/backend/services/test_mcp_tools_yaml.py -k TagMcp -v`
- [ ] `cd docs && node --test src/lib/__tests__/llms-txt.test.js`
- [ ] `npx skills add rhesis-ai/rhesis -g` → `/rhesis` → four-path menu → PRD path → plan with `metric_scope`
- [ ] Architect: phase refs appear per turn; `save_plan` rejects scope mismatches

---

## Bottom line

**Before:** Platform knowledge lived in three places (Telemachus prompts, skill files, ad-hoc docs) with duplicated definitions and no shared routing. Wrong `metric_scope` pairings failed silently at run time.

**After:** One reference tree (`skills/rhesis/references/`) shared by skill and Telemachus; docs + glossary are the canonical human/machine vocabulary layer on top. Telemachus loads phase knowledge lazily (~70% smaller fixed prompt). Agents route through `workflow-index.md`, fetch glossary for terms, and catch scope mismatches in planning — not after an empty test run.

**CI:** Lint + backend + SDK green ✅


sdk/src/rhesis/sdk/agents/architect/prompt_loader.py
```diff
@@ -0,0 +1,153 @@
...
+
+    Override with ``RHESIS_SKILLS_REFERENCES`` for custom layouts.
+    """
+    override = os.environ.get("RHESIS_SKILLS_REFERENCES", "").strip()
+    if override:
+        path = Path(override)
+        return path if path.is_dir() else None
+
+    # architect → agents → sdk package → rhesis → src → sdk project → repo root
+    repo_root = _ARCHITECT_DIR.parents[5]
```
harry-rhesis: Blocker (see main comment). `parents[5]` assumes a monorepo checkout on disk. In the backend image, `/app/skills` is never copied (the Dockerfile copies `sdk/`, `penelope/`, `packages/rhesis/`, `ee/backend/` only), and the SDK wheel does not package `skills/rhesis/references/`. So this resolves to a path that does not exist in production. The hard-coded depth is also fragile: if this module ever moves, the lookup silently breaks. Suggest packaging the references into the wheel via hatch `force-include` so the `skill_refs` fallback below is real.

sdk/src/rhesis/sdk/agents/architect/prompt_loader.py
```diff
@@ -0,0 +1,153 @@
...
+    references = repo_root / "skills" / "rhesis" / "references"
+    if references.is_dir():
+        return references
+
+    bundled = _ARCHITECT_DIR / "prompt_templates" / "skill_refs"
+    if bundled.is_dir():
+        return bundled
+
+    logger.warning(
+        "skills/rhesis/references not found — Architect prompt includes may be incomplete"
```
harry-rhesis, line 49: This warning-and-return-None path is the one that actually fires in production, because neither `skills/rhesis/references/` nor the `skill_refs` bundle exists in the image. Nothing populates `skill_refs` at build time, so the fallback is currently dead code. Once the loader returns None, the fixed system prompt include of `entity-model.md` raises instead of degrading (see the system_prompt.j2 comment).

sdk/src/rhesis/sdk/agents/architect/prompt_templates/system_prompt.j2
```diff
@@ -2,536 +2,12 @@ You are the Rhesis Architect, an expert conversational agent that helps users de
...
+Operational knowledge loads **per turn** based on your current mode and workflow path. This system prompt stays fixed; phase guidance appears in each iteration message.
 
-**Resolution:** list_* + $filter by name → id → get_* (full detail) or mutate/execute.
-**After generate_test_set:** get_test_set + list_test_set_tests to verify before offering execute_test_set.
-**Metric edits:** get_metric (read prompt) → update_metric (precise fields) or improve_metric (natural language).
+{% include "workflow-routing.j2" %}
 
-## Resolving Entities by Name
-When the user refers to any entity by name (endpoint, metric, behavior, test set, project, etc.), look it up yourself using the appropriate list_* tool — never ask the user for an ID. Always pass the search value lowercased and wrap the column with `tolower()` for case-insensitive matching. This applies to ALL entity types: endpoints, metrics, behaviors, categories, topics, test sets, projects, etc.
+{% include "entity-model.md" %}
```
harry-rhesis, line 13: This is the exact crash point in production. `entity-model.md` exists only under `skills/rhesis/references/`, and this is a plain include (no `ignore missing`). When `resolve_skills_references_dir()` returns None (which it does in the image), Jinja raises `TemplateNotFound` here at agent construction. Either package the references (preferred), or use `{% include "entity-model.md" ignore missing %}` as a defensive measure so the agent degrades rather than crashing.

sdk/src/rhesis/sdk/agents/architect/prompt_loader.py
```diff
@@ -0,0 +1,153 @@
...
+                    "phases/analysis.md",
+                    "result-analysis.md",
+                    "odata-patterns.md",
+                ]
+            )
+        elif wp == WorkflowPath.DIRECT:
+            includes.extend(["phases/direct-requests.md", "odata-patterns.md"])
+        # UNSET: routing only (system prompt) — no phase block
+
+    elif m == AgentMode.PLANNING:
```
harry-rhesis, line 98: Two things here. First, this branch loads `phases/planning.md` alongside `telemachus-reuse.j2`, `telemachus-guidelines.j2`, and `telemachus-save-plan.j2`, which restate the same reuse/approval/naming guidance. That is the largest per-turn payload and it is largely redundant (see DRY item in main comment). Second, `telemachus-guidelines.j2` carries global rules (no tool names in prose, no raw IDs, confirm-before-acting) but is only loaded in PLANNING, so those rules are absent during CREATING/DISCOVERY/EXECUTING. Recommend hoisting the global rules into the fixed system prompt.

harry-rhesis (commented):
Thank you, @nicolai-rhesis, for this contribution. This is a substantial and genuinely valuable piece of work, and the direction is right. Collapsing the two skills into one `/rhesis` flow, moving the Architect to lazy per-turn prompt loading, and validating metric scope at plan time (instead of letting mismatched single/multi-turn metrics get silently dropped at run time) all improve the agent experience meaningfully. The layered reference system is thoughtful and the `metric_scope` validation is well tested.

I have one blocking issue, two behavioral concerns, and some DRY cleanup that would make the stated "one source of truth" goal actually hold. Summary below, specifics inline.

## 1. Blocker: the Architect crashes in production (skill references are not packaged)

`prompt_loader.resolve_skills_references_dir()` locates the shared `.md` references by walking `_ARCHITECT_DIR.parents[5] / "skills" / "rhesis" / "references"`, which assumes a monorepo checkout on disk. That holds in local dev but not in the deployed image:

- The SDK wheel `include` list in `sdk/pyproject.toml` packages only `py.typed`, `detectors.yaml`, and synthesizer `*.jinja`. It does not package `skills/rhesis/references/`.
- `apps/backend/Dockerfile` copies `sdk/`, `penelope/`, `packages/rhesis/`, and `ee/backend/`, but never copies repo-root `skills/`. So `/app/skills` does not exist in the image.
- The `bundled` fallback (`prompt_templates/skill_refs`) is never populated by any build step, so it is effectively dead code.
- No `RHESIS_SKILLS_REFERENCES` override is set in any Dockerfile, compose file, or deploy config.

So in production `resolve_skills_references_dir()` returns `None`, and `system_prompt.j2`'s plain `{% include "entity-model.md" %}` (no `ignore missing`, and that file exists only under `skills/rhesis/references/`) raises `TemplateNotFound` at agent construction. This is a regression: on `main` the system prompt was a self-contained packaged `.j2`. Suggested fix: package the references into the wheel via hatch `force-include` into the existing `skill_refs` fallback path, and add a test that renders the fixed system prompt with no monorepo present.

## 2. Global behavioral rules now load only in PLANNING mode

`telemachus-guidelines.j2` carries global rules (never mention tool names, never paste raw IDs in prose, and the full confirm-before-acting protocol), but `phase_include_names()` injects it only in PLANNING. During CREATING, DISCOVERY, and EXECUTING those rules are absent; the only always-on guard is the one-line write gate in `workflow-routing.j2`. On `main` these rules were always in context. Recommend hoisting the truly global rules into the fixed system prompt.

## 3. DRY: "one source of truth" is only half true

The `.md` references are genuinely single-sourced through the ChoiceLoader, which is great. But the `telemachus-*.j2` partials are a second, hand-maintained copy of the same domain guidance, and `telemachus-parity.md` is a manual checklist to keep the two aligned. A parity checklist is the tell: it exists only because two copies can drift. Concretely duplicated:

- The four-path menu block is byte-identical across `SKILL.md`, `workflow-index.md`, and `workflow-routing.j2`.
- `telemachus-creation-order.j2` mirrors `phases/creation.md`.
- Reuse/naming/approval-gate prose is duplicated between the `.j2` partials and `phases/*.md`.

In PLANNING, `phase_include_names()` loads both copies into the same prompt (`phases/planning.md` plus `telemachus-reuse.j2`, `telemachus-guidelines.j2`, `telemachus-save-plan.j2`). Recommendation: let the `.j2` partials carry only agent-runtime mechanics (the JSON response schema, `save_plan` validation, `await_task`, security/identity), keep all domain guidance in the shared `.md` files consumed by both, and then delete `telemachus-parity.md` because the parity problem is gone.

## 4. Context loading: right call

The lazy per-turn design genuinely reduces context. The fixed system prompt drops from roughly 8,900 words loaded every turn to roughly 2,500, and even the heaviest turn (PLANNING) stays under the old always-on constant. The remaining win is realizing that fully by removing the PLANNING double-load in item 3 and hoisting the always-on rules in item 2.

## 5. Smaller items

- `workflow_path` latches once and never re-classifies, so pasting a PRD after an exploration start will not engage the PRD path (only `reset()` clears it).
- `odata-patterns.md` is injected in nearly every branch of `phase_include_names()`, so it is a candidate for the fixed system prompt instead.
- `phases/planning.md` points the model at `use-case-bracketfeld.md`, which is never injected, so it references content the agent never receives.
- Test gaps: the `$select` "always include id" contract is untested, and `list_tags` `default_query` / `assign_tag` `requires_confirmation` are not asserted in the yaml test.

## Backend: looks good

`tag.py` `$select` stays org-scoped (the projection runs after the already-scoped query) and the count-header path is safe. The MCP `list_tags` / `assign_tag` tools map correctly to existing endpoints. No multi-tenancy concerns.

## Implementation Plan

Please checkout the implementation plan proposed by Claude: https://claude.ai/code/artifact/63d73faf-f522-41e2-aa11-c7c3c21c21e2?via=auto_preview

sdk/src/rhesis/sdk/agents/architect/prompt_templates/workflow-routing.j2
```diff
@@ -0,0 +1,33 @@
...
+| Endpoint + test/explore | Exploration |
+| Compare runs, analyze last run | Run / analyze |
+| Otherwise | Four-path menu |
+
+### Four-path menu (ambiguous start)
+
+```text
+What would you like to do?
+
+1. Quick exploration — fast scan of an endpoint's domain and boundaries
```
harry-rhesis, line 25: This four-path menu is duplicated byte-for-byte across three files: here, `skills/rhesis/references/workflow-index.md`, and `skills/rhesis/SKILL.md`. Good candidate to single-source into one `.md` and include it in the others via the loader, so the menu cannot drift between the coding skill and Telemachus.

skills/rhesis/references/telemachus-parity.md
```diff
@@ -0,0 +1,47 @@
+# Telemachus (Architect) parity plan
```
harry-rhesis: This parity checklist is the clearest signal of the remaining duplication: it exists to keep two hand-maintained copies (the `telemachus-*.j2` partials and the `phases/*.md` references) in sync. If the `.j2` partials are trimmed to agent-only mechanics and all domain guidance lives once in the shared `.md` files, this file can be deleted because there is nothing left to keep in parity.

sdk/src/rhesis/sdk/agents/architect/workflow.py
```diff
@@ -0,0 +1,105 @@
...
+    "list metric",
+    "list behavior",
+    "update metric",
+    "improve metric",
+    "link metric",
+    "unlink metric",
+)
+
+
+def infer_workflow_path(message: str, *, has_attachments: bool = False) -> WorkflowPath | None:
```
harry-rhesis, line 53: Minor: the caller latches `workflow_path` once and never re-classifies (only `reset()` clears it). A user who starts with exploration and later pastes a PRD will not have the PRD path engage, so `prd-workflow.md` never loads for that conversation. Consider allowing a forward upgrade to PRD on a strong PRD signal even after the path is set.

tests/backend/services/test_mcp_tools_yaml.py
```diff
@@ -100,6 +100,22 @@ def test_new_tools_in_yaml(self):
         assert not missing, f"Missing tools: {sorted(missing)}"
 
 
+@pytest.mark.unit
+class TestTagMcpToolsPresent:
+    TAG_TOOLS = frozenset({"list_tags", "assign_tag"})
+
+    def test_tag_tools_in_yaml(self):
```
harry-rhesis, line 107: Nice to have: this asserts the tools exist and checks `assign_tag`'s path/method, but not the fields most likely to regress silently: `list_tags` `default_query` (`$select: id,name`) and `assign_tag` `requires_confirmation: true`. Worth an assertion each.

akwasigroch:
Few comments from my side:
- expand the PRD abbreviation, for developers it can be a bit confusing because they would think about production
- the code is not in the code blocks that we decided to use in the documentation
- do we need the tool-catalog.md in the skill? the agent can discover it from the MCP that shares the tool catalog

nicolai-rhesis:
Thanks for the thorough review, @harry-rhesis — the blocker and the main behavioral concerns are addressed in `e8c17fd2f` and follow-ups on this branch.

**Production packaging (blocker)**
- Skill references are bundled into the SDK wheel via hatch `force-include` (`sdk/pyproject.toml` → `prompt_templates/skill_refs/`).
- `resolve_skills_references_dir()` walks up from the module and falls back to the bundled path.
- Added tests: `TestBundledSkillReferences::test_render_system_prompt_with_bundled_refs_only` and `test_wheel_includes_skill_refs`.

**Global rules always on**
- `telemachus-guidelines.j2` and `odata-patterns.md` are now in the fixed `system_prompt.j2`, not only in PLANNING.

**DRY / duplication**
- Removed `telemachus-parity.md` and `telemachus-creation-order.j2`; consolidated creation/reuse guidance into shared `phases/*.md`.
- PLANNING no longer double-loads the telemachus reuse/guidelines partials alongside the phase markdown.

**Workflow re-classification**
- Added `resolve_workflow_path_update()` so pasting a PRD after an exploration start can switch to the requirements path.

**Latest commit (`9b7dca202`)**
- Renamed docs/skill references from "PRD workflow" → "Requirements workflow" (`/docs/agent-skill/requirements`, `requirements-workflow.md`), with Product Requirements Document (PRD) spelled out on first mention.

CI is green on the latest push. Would appreciate a re-review when you have a moment — happy to tackle the remaining non-blocking DRY items (single-sourcing the four-path menu) in a follow-up if you prefer keeping this PR focused.

harry-rhesis (approved):
Thank you, @nicolai-rhesis! You addressed every review point thoroughly. The blocker is fixed properly (force-include into the wheel + test that renders without a monorepo), global rules are hoisted into the fixed system prompt, the parity checklist and duplicate partials are gone, workflow path upgrades work, odata-patterns is in the fixed prompt, use-case-bracketfeld loads for PRD planning, and both tag tool test assertions are in place.

The one remaining item (the four-path menu living in three files) is low-risk and tracked in #2196.

Great work on this one.

PR #2123, opened 2026-07-09
chore(backend): temporary per-query request timing probe

EmanueleDeRossi1:
Temporary diagnostic to find out why `/tests` is still slow on dev after #2117. Env-gated (`RHESIS_ENABLE_REQUEST_TIMING=1`), logs a total/db/other breakdown plus every query's own normalized SQL + duration, slowest first, per request.

Plain ASGI middleware, not `BaseHTTPMiddleware` — the latter can silently skip the post-request logging code if the connection gets cancelled behind a reverse proxy (bit us on the last version of this probe).

Not meant to merge long-term — revert once diagnosis is done, same as `debug_sql_timing.py` (57ed58154 → 639a09d3c) and the previous version of this file (6ec310897 → beb79ed3a → d5811d40f in #2117).

### DIFF
--- a/apps/backend/src/rhesis/backend/app/debug_request_timing.py
+++ b/apps/backend/src/rhesis/backend/app/debug_request_timing.py
@@ -1,0 +1,113 @@
+"""Temporary request-vs-DB timing probe for diagnosing the /tests latency gap.
+
+Per-query timing already includes network round-trip time, since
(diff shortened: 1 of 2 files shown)


EmanueleDeRossi1:
Closing this: this was not meant to be merged, and I do not need it anymore.

PR #2136, opened 2026-07-09
Fix v0.10.0 UI polish and onboarding video build

nicolai-rhesis:
## Summary
- Fix Playground endpoint select label overlay (`InputLabel shrink` + `notched`)
- Resume recent Architect conversations within 15 minutes via `sessionStorage`
- Reorder knowledge source detail sections (tags below extracted content) and remove double-elevated comments card
- Fix token creation role template select label/placeholder styling
- Point onboarding "Back home" to `/architect` instead of marketing site
- Bake `NEXT_PUBLIC_ONBOARDING_VIDEO_URL` into K8s frontend Docker builds so the welcome video works on staging

## Test plan
- [ ] Playground: endpoint select shows floated label without overlap
- [ ] Architect: start chat, navigate away, return within 15 min — conversation restores
- [ ] Knowledge source detail: tags below extracted content; comments single card elevation
- [ ] Create token drawer: role template shows "Select a role" placeholder correctly
- [ ] Onboarding: "Back home" navigates to app home (`/architect`)
- [ ] Rebuild/deploy frontend on K8s and verify onboarding welcome video plays

### DIFF
--- a/.github/workflows/frontend-k8s.yml
+++ b/.github/workflows/frontend-k8s.yml
@@ -78,6 +78,8 @@ jobs:
       context: .
       environment: ${{ inputs.environment || github.event.inputs.environment || 'dev' }}
       tag: ${{ inputs.tag || github.event.inputs.tag }}
+      build_args: |
+        NEXT_PUBLIC_ONBOARDING_VIDEO_URL=${{ secrets.NEXT_PUBLIC_ONBOARDING_VIDEO_URL }}
     secrets: inherit
 
   deploy:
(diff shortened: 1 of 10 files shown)


nicolai-rhesis:
Addressed peqy review:

1. **ArchitectClient (blocking):** Clear `activeSessionId` and `sessions` immediately when starting a project-scoped refetch, then restore the resumable session after `getSessions()` resolves.

2. **TokenScopeField:** Restored `size="small"` on the FormControl while keeping `shrink` + `displayEmpty` + placeholder `renderValue`.

**On `NEXT_PUBLIC_ONBOARDING_VIDEO_URL`:** Yes, intentional — it's a public YouTube watch URL (same value already in Helm `values-*.yaml`). `NEXT_PUBLIC_*` must be baked at build time for client components; the ConfigMap alone cannot reach `WelcomeVideoStep`.

PR #2137, opened 2026-07-09
Fix K8s frontend build secret for onboarding video URL

nicolai-rhesis:
## Summary
- Fix invalid `release.yml` workflow parse error caused by referencing `secrets` inside a reusable workflow `with:` input
- Declare `NEXT_PUBLIC_ONBOARDING_VIDEO_URL` on `k8s-build-push` and inject it during the Docker build step instead

## Test plan
- [ ] `release.yml` workflow file validates on GitHub
- [ ] `deploy-all-k8s` → `frontend-k8s` → `k8s-build-push` chain runs successfully
- [ ] Frontend image build receives `NEXT_PUBLIC_ONBOARDING_VIDEO_URL` build arg when secret is set

### DIFF
--- a/.github/workflows/frontend-k8s.yml
+++ b/.github/workflows/frontend-k8s.yml
@@ -78,8 +78,6 @@ jobs:
       context: .
       environment: ${{ inputs.environment || github.event.inputs.environment || 'dev' }}
       tag: ${{ inputs.tag || github.event.inputs.tag }}
-      build_args: |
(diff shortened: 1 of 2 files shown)


nicolai-rhesis:
Addressed peqy review: removed the `workflow_call.secrets` block from `k8s-build-push.yml`.

Declaring only `NEXT_PUBLIC_ONBOARDING_VIDEO_URL` there would restrict the called workflow to that single secret. With no `workflow_call.secrets` declaration, callers using `secrets: inherit` pass the full secret set (including `GCP_SA_KEY_*`, `TF_PROJECT_ID_*`), and the build step still reads `secrets.NEXT_PUBLIC_ONBOARDING_VIDEO_URL` when present.

PR #2138, opened 2026-07-09
Fix onboarding welcome video via server-side env

nicolai-rhesis:
## Summary
- Read `NEXT_PUBLIC_ONBOARDING_VIDEO_URL` in the server `onboarding/page.tsx` and pass it as a prop to `WelcomeVideoStep`
- No workflow/YAML changes — uses the existing pod env from Helm ConfigMap (`values-*.yaml` → ConfigMap)

## Why this works
`WelcomeVideoStep` is a client component; `process.env.NEXT_PUBLIC_*` there is only populated at Next.js **build time**. On K8s, the value is already provided at **runtime** via ConfigMap. Server components read pod env on each request.

The GitHub **environment** secret `NEXT_PUBLIC_ONBOARDING_VIDEO_URL` (stg/dev) is used by Cloud Run deploys (`frontend.yml`). K8s runtime uses Helm values — no hardcoding in workflows.

## Test plan
- [ ] Deploy frontend to staging
- [ ] Onboarding welcome step shows video thumbnail (not "coming soon")

### DIFF
--- a/.github/workflows/frontend-k8s.yml
+++ b/.github/workflows/frontend-k8s.yml
@@ -78,8 +78,6 @@ jobs:
       context: .
       environment: ${{ inputs.environment || github.event.inputs.environment || 'dev' }}
       tag: ${{ inputs.tag || github.event.inputs.tag }}
-      build_args: |
-        NEXT_PUBLIC_ONBOARDING_VIDEO_URL=https://youtu.be/fNtxqGtBu2Y
     secrets: inherit
 
   deploy:
--- a/.github/workflows/frontend-k8s.yml
+++ b/.github/workflows/frontend-k8s.yml
@@ -78,6 +78,8 @@ jobs:
       context: .
       environment: ${{ inputs.environment || github.event.inputs.environment || 'dev' }}
       tag: ${{ inputs.tag || github.event.inputs.tag }}
+      build_args: |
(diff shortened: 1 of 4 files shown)


nicolai-rhesis:
Updated approach — reverted `build_args` (same class of fix as #2137).

**Correct fix:** Read `NEXT_PUBLIC_ONBOARDING_VIDEO_URL` in the **server** `onboarding/page.tsx` and pass it as a prop to `WelcomeVideoStep`. The Helm ConfigMap already sets this env var on the pod; server components read it at **request time**, so no Docker build-arg or GitHub secret is required.

This avoids both failure modes:
1. `secrets` in reusable workflow `with:` inputs (invalid YAML)
2. Missing `NEXT_PUBLIC_ONBOARDING_VIDEO_URL` GitHub secret (build-arg never set)

nicolai-rhesis:
Addressed review comments:

- **peqy (hard-coded YouTube URL in workflow):** Reverted — latest diff has **no** `frontend-k8s.yml` changes and no hardcoded URL.
- **User:** Confirmed `NEXT_PUBLIC_ONBOARDING_VIDEO_URL` exists as a GitHub **environment** secret on `stg`/`dev` (not repo-level). K8s pods already get the same value from Helm ConfigMap. This PR reads that runtime env in the server page — no workflow or build-arg changes needed.

PR #2144, opened 2026-07-10
Store T&C acceptance in user_settings.terms

nicolai-rhesis:
## Updated (2026-07-14)
This PR originally added dedicated `user.terms_accepted_*` columns and a login-page checkbox.

**That is no longer the flow.** Terms acceptance is now stored in `user_settings.terms` and enforced post-login.

## Summary
Persist T&C acceptance server-side with versioning and enforce acceptance when terms change.

## Final flow (current behavior)
- **Persistence**
  - Acceptance is stored under `user_settings.terms`:
    - `version`: current T&C version (e.g. `1.0`)
    - `accepted_at`: ISO timestamp

- **Onboarding step 0 (Organization Details)**
  - If the user has *not* accepted the current version, step 0 shows a T&C checkbox.
  - Clicking **Next** calls `POST /auth/accept-terms` (authenticated; org context not required) to record acceptance.
  - If already accepted, the checkbox is hidden.

- **Global post-login gate (version bumps / existing users)**
  - After login, the protected layout checks `GET /auth/terms-status` (authenticated; user-scoped).
  - If the user has *not* accepted the current version, a **non-dismissable modal** blocks the app until they accept.
  - Modal copy distinguishes first acceptance vs. version bump (`has_prior_acceptance`).
  - Skipped in Quick Start mode (local dev auto-login).
  - Accepting in the modal calls `POST /auth/accept-terms`.

- **Endpoints**
  - `GET /auth/terms-status` (authenticated, user-scoped): returns `terms_accepted` and `has_prior_acceptance`.
  - `POST /auth/accept-terms` (authenticated, user-scoped): records acceptance in `user_settings.terms`.

## Notes
- No unauthenticated email-based “terms status” lookup (avoids boolean-oracle concerns).
- Acceptance is server-managed (not writable via `PATCH /users/settings` or `PUT /users/{id}`).
- **Fail-open policy:** if `/auth/terms-status` is unreachable (network/API failure), the gate does not block the app. A definitive `terms_accepted: false` from a 200 response still blocks until acceptance. This avoids hard-locking all users during transient outages; compliance relies on the status endpoint being reachable in normal operation.
- `client-wrapper.test.tsx` changes are unrelated to terms — they came from a merge/rebase with `main` (project detail test coverage).

## Test plan
- New user: sign up → onboarding step 0 requires checkbox → Next succeeds and acceptance is recorded.
- Existing user after bumping `CURRENT_TERMS_VERSION`: login → modal blocks app → accept → modal disappears.
- Verify `GET /auth/terms-status` returns `terms_accepted=true` after acceptance.
- Verify `PUT /users/{id}` with `user_settings.terms` returns 422.


nicolai-rhesis:
Addressed in 5d06252:

- **Migration**: revision ID clashed with an existing migration (`b8c9d0e1f2a3`); renamed to `c9e8f7a6b5d4`.
- **`accept_terms` contract**: frontend only sends it when the user checks the box this session (not when `/auth/terms-status` already returned true). Returning logins omit the field.
- **Backend**: `record_terms_acceptance` is now a no-op when the current version is already on file; `_ensure_terms_accepted` rejects login/magic-link when acceptance is still required and not provided.

OAuth `terms_accepted` remains a session flag set only after the checkbox gate on the login page — same pattern as before, just persisted server-side on callback.

nicolai-rhesis:
Addressed peqy critical in 2b97cc: `POST /auth/magic-link` is enumeration-safe again — when terms are required but `accept_terms` is missing, it returns the same generic 200 without creating a user or sending email. Regression test added.

nicolai-rhesis:
Moved T&C acceptance off the login page into onboarding step 0 (Organization Details). Users authenticate first, then accept terms when completing step 0 via `POST /auth/accept-terms`. Login/register/OAuth/magic-link no longer require or display the checkbox.

nicolai-rhesis:
Addressed peqy re-review in 2978821:

- **`/auth/terms-status`**: now authenticated (no `email` param); onboarding passes the session token. Removed from public routes.
- **Doc drift**: updated `auth/terms.py` and endpoint docstrings for onboarding-based flow.

Re **existing-user gate**: not in this PR — new sign-ups hit step 0; returning users with an org skip onboarding. A follow-up guard (redirect when `!user_has_accepted_current_terms`) would cover version bumps for existing users; happy to add in a separate change if you want it here.

nicolai-rhesis:
Addressed peqy re-review in 2978821:

- **`/auth/terms-status`**: now authenticated (no `email` param); onboarding passes the session token. Removed from public routes.
- **Doc drift**: updated `auth/terms.py` and endpoint docstrings for onboarding-based flow.

Re **existing-user gate**: not in this PR — new sign-ups hit step 0; returning users with an org skip onboarding. A follow-up guard (redirect when `!user_has_accepted_current_terms`) would cover version bumps for existing users; happy to add in a separate change if you want it here.

harry-rhesis (commented):
Thanks for the thorough work on this, Nicolai — the direction is clearly right. Moving from `localStorage` to server-side persistence with versioning is a solid improvement, and shifting T&C acceptance into onboarding step 0 is a better UX than gating login. The idempotent `record_terms_acceptance` helper, tests, and the magic-link enumeration-safety fix all look good.

## Main feedback: consider `user_settings` instead of dedicated columns

Rather than adding `terms_accepted_at` / `terms_accepted_version` as columns on `user`, I'd suggest storing this in the existing `user_settings` JSONB — e.g. a `terms: { accepted_at, version }` slice on `UserSettings`, alongside `onboarding`, `polyphemus_access`, etc.

### Why I'd lean toward `user_settings`

1. **Less invasive** — no migration, no new `User` model columns, no extra fields on `UserBase`.
2. **Infrastructure already exists** — `UserSettingsManager`, `GET/PATCH /users/settings`, and `useUserSettings` on the frontend. The write path can stay server-controlled (`POST /auth/accept-terms` updates `user_settings.terms` internally; don't expose it via the user-writable `PATCH /users/settings`).
3. **Co-located with onboarding** — T&C now lives on onboarding step 0, and onboarding progress is already in `user_settings`. Keeping terms there avoids splitting user lifecycle metadata across two storage mechanisms.
4. **Smaller surface area** — this approach could drop the migration, model columns, public `GET /auth/terms-status?email=…`, its rate limit, and the separate `auth-client` fetch from onboarding.

### Trade-offs (dedicated columns vs `user_settings`)

| | Dedicated columns (this PR) | `user_settings` |
|---|---|---|
| **Querying / compliance** | Easy indexed SQL — e.g. "all users not on v2.0" | Requires JSONB queries; fine at current scale, weaker for large-scale audit reports |
| **Schema rigidity** | DB-enforced types, clear audit trail in migrations | Flexible, consistent with how we store similar data (`onboarding`, `polyphemus_access`) |
| **Invasiveness** | Migration + model + schema changes | Schema-only change to `UserSettings` |
| **Read path** | Public `/auth/terms-status?email=` works pre-org | `GET /users/settings` currently 403s without an org — needs a fix for onboarding step 0 |
| **Write safety** | Only writable via dedicated endpoint | Must ensure users can't self-assert acceptance via `PATCH /users/settings` |

**My take:** for this codebase, `user_settings` is the better default — it's less invasive and matches existing patterns. Dedicated columns are defensible if we expect frequent compliance queries across the full user base (e.g. "find everyone who hasn't accepted v2.0"). If that's not a near-term need, the flexibility and consistency of `user_settings` wins.

### Pre-org read path

One reason the PR added `GET /auth/terms-status?email=` is that `GET /users/settings` 403s for users without an organization (`get_tenant_context`), which is exactly onboarding step 0. If we move storage to `user_settings`, we still need a read path for pre-org users, but it doesn't have to be an unauthenticated email oracle. Options:

- Allow authenticated self-read of settings (or just `terms`) without org context — e.g. `require_current_user_or_token_without_context` + `get_db_session`, same pattern as `POST /auth/accept-terms`.
- Or return `terms_accepted` on an existing authenticated profile/session endpoint.

That would let us drop `/auth/terms-status` entirely — no enumeration concern, no extra rate limit, no direct `getClientApiBaseUrl()` fetch from onboarding.

### Smaller follow-ups

- **Existing users:** if someone never re-enters onboarding, is there a guard when `CURRENT_TERMS_VERSION` bumps? Worth a redirect or middleware check, not only onboarding step 0.
- **Stale docs:** `auth/terms.py` and the `/auth/terms-status` docstring still describe login-page behavior — worth updating to match the onboarding flow.

Overall this is a well-executed PR; the main architectural question is where to persist the data. Happy to discuss which trade-off we want to optimize for.

nicolai-rhesis:
Addressed Harry’s main suggestion in 3441bd4:

- Persist T&C acceptance in `user_settings.terms` (server-managed), not dedicated `user` columns.
- Removed the terms columns + Alembic migration.
- `/auth/terms-status` + `/auth/accept-terms` continue to be authenticated and user-scoped.
- Added a global post-login non-dismissable modal gate to enforce re-acceptance after version bumps.
- Added a clear onboarding UX guard when `sessionToken` is missing.

nicolai-rhesis:
Addressed peqy’s latest questions/improvements in recent commits:

- **Global enforcement gate**: added a non-dismissable post-login modal in the protected layout; if `GET /auth/terms-status` reports not accepted, the app is blocked until the user accepts via `POST /auth/accept-terms`.
- **Onboarding missing token**: step 0 now shows a clear session-expired error instead of attempting `acceptTerms` with an empty token.
- **Terms status** remains authenticated + user-scoped (no email oracle).

(Also incorporated Harry’s suggestion: acceptance now lives under `user_settings.terms`.)

nicolai-rhesis:
Addressed peqy **critical**: `user_settings.terms` is now **read-only**.

- Removed `terms` from the writable `UserSettings` schema.
- Exposed it only on `UserSettingsRead`.

So `PUT /users/{user_id}` (and any other writable path using `UserSettings`) can no longer accept `user_settings.terms` — requests with that key now fail validation (extra=forbid), preventing self-asserted acceptance.

nicolai-rhesis:
@harry-rhesis please re-review.

harry-rhesis (approved):
Thanks Nicolai for incorporating the feedback and iterating quickly on this. The final design looks good.

## Summary

The PR now matches the architectural direction we discussed:

- Acceptance is stored in `user_settings.terms` (`version` + `accepted_at`), not dedicated user columns. The migration was removed.
- Read/write is properly scoped: `GET /auth/terms-status` and `POST /auth/accept-terms` are authenticated and user-scoped (no email oracle).
- Pre-org onboarding works via `require_current_user_or_token_without_context` and authz exemptions.
- `terms` is read-only for clients (`UserSettingsRead` only), so users cannot self-assert acceptance via `PUT /users/{id}` or `PATCH /users/settings`.
- A global `TermsAcceptanceGate` handles version bumps and existing users who skip onboarding.
- Onboarding step 0 remains as a secondary path (checkbox + `acceptTerms` on Next).

CI is green across backend, frontend, lint, and E2E.

## Minor nits (non-blocking)

1. **PR title** still says "user record"; worth updating to reflect `user_settings`.
2. **`terms.py` docstring** still says "stored on the user record"; should reference `user_settings.terms`.
3. **Modal copy**: "Updated Terms" may confuse first-time users who have never accepted. Consider different copy for first acceptance vs version bump.
4. **Fail-open on API errors** in `TermsAcceptanceGate`: if `/auth/terms-status` fails, the gate silently lets users through. Fine for dev/E2E, but worth confirming this is acceptable for production compliance.
5. **`client-wrapper.test.tsx`** looks unrelated to terms; fine if from a rebase, but ideally split or noted in the PR description.
6. **Regression test**: a small test asserting `PUT /users/{id}` rejects `user_settings.terms` would lock down the peqy fix.

Approving. Nice work.

PR #2146, opened 2026-07-10
chore(config): clean up env and harden docker setup

akwasigroch:
## Purpose
Clean up environment/config handling so local development works with sensible defaults, remove duplicated values across `.env.example`, `docker-compose.yml`, and the `rh` dev script, and harden the Docker setup. Along the way, make `FRONTEND_URL` the single source of truth for the frontend's public origin (replacing `NEXTAUTH_URL`), derive cookie security from the configured URL scheme, and fix UUID generation for self-hosted deployments served over plain HTTP.

### Config defaults & hardening (backend)
- **settings.py**: dev-friendly defaults for `DB_HOST`/`DB_NAME`, `FRONTEND_URL`, and `API_BASE_URL` so the app loads out of the box locally; DB credentials remain required (validated at `app_url` build time). Bump `JWT_ACCESS_TOKEN_EXPIRE_MINUTES` default to 10080 (7 days). Add a `secure_cookies` property that reports `https` based on the `API_BASE_URL` scheme.
- **main.py / auth.py**: the backend `SessionMiddleware` now derives `https_only` from `ApplicationSettings.secure_cookies` (URL scheme) instead of a "running locally" check; logout comment updated to match.

### FRONTEND_URL as the canonical frontend origin
- Replace `NEXTAUTH_URL` with `FRONTEND_URL` everywhere: `.env.example`, `docker-compose.yml`, `apps/frontend/Makefile`, `playwright.config.ts`, e2e seed helper, the CI `frontend.yml` build args, Helm charts (`values*.yaml`, `configmap.yaml`), infra secret scripts, and docs.
- **auth.ts**: at startup, seed `NEXTAUTH_URL` from `FRONTEND_URL` when neither `AUTH_URL`/`NEXTAUTH_URL` is set, so Auth.js gets a correct base URL instead of inferring it from the container hostname (which broke post-login redirects). Frontend session cookie `secure` flag now comes from `shouldUseSecureCookies()` (derived from `FRONTEND_URL`).
- **url-resolver.ts**: add `shouldUseSecureCookies()`.
- **proxy.ts**: session-clearing cookies use the same `shouldUseSecureCookies()` source as when they're set, so the clear reliably overwrites the cookie (replaces the old `FRONTEND_ENV`-based branch).

### Docker & host ports
- **docker-compose.yml**: collapse duplicated anchors, hardcode internal container ports/credentials, add an `x-host` anchor as the single source of truth for host-facing ports and derived URLs, require `DB_ENCRYPTION_KEY` (no unsafe default), and drop unnecessary host port exposure for postgres/redis/worker. Uses `BACKEND_PORT`/`FRONTEND_PORT` for host binds independent of the public `*_URL` values.
- **.env.example**: clarify that `BACKEND_PORT`/`FRONTEND_PORT` are host bind ports (mapped to fixed container ports 8080/3000) and independent of the `*_URL` public URLs; drop `NEXTAUTH_URL`.
- **rh**: prompt for backend/frontend host ports (Enter accepts defaults) and write matching `BACKEND_PORT`/`FRONTEND_PORT` + public URLs to `.env.docker.local`; trim now-defaulted env vars from the generated dev `.env`; pass `--env-file .env.docker.local` to `down`/`logs`/`delete` compose commands; show the chosen ports in the access URLs and drop the worker health-check line.

### UUID over HTTP
- **utils/uuid.ts** (new): `safeRandomUUID()` prefers `crypto.randomUUID`, falls back to a v4 UUID built from `crypto.getRandomValues` (a CSPRNG, not secure-context gated), and only uses `Math.random` as a last resort. Fixes `crypto.randomUUID is not a function` on self-hosted deployments reached over plain HTTP on a LAN IP / internal hostname.
- Route client call sites through it: onboarding, invite forms (`InviteTeamStep`, `TeamInviteForm`), tool import, and run-drawer correlation ids.

### Tests & docs
- **tests**: update settings tests for the new defaults and deferred credential validation.
- **docs**: update self-hosting and contribute guides to reflect `FRONTEND_URL` and the removal of `NEXTAUTH_URL`.

## Testing
- `uv run pytest tests/backend/app/test_settings.py`
- Local `./rh start` (port prompts, generated `.env.docker.local`, access URLs)
- Frontend login/logout redirects and cookie clearing over both http and https


harry-rhesis (commented):
Thanks for the context on the hardcoded DB/Redis credentials, that makes sense given Postgres and Redis are no longer published to the host in this setup.

One follow-up: `.env.example` and `docs/deployment/self-hosting.mdx` still instruct users to change `APP_DB_PASS`/`REDIS_PASSWORD` for production ("Must change the Database and Redis Configuration for the Production deployment" and "Change all default passwords and secrets"). Since docker-compose now hardcodes these values, that guidance is stale and could give self-hosters false confidence that they've secured their deployment. Would be good to update those docs to reflect the new design, for example noting these vars only apply outside docker-compose, or dropping the instruction entirely.

Nice work on the NEXTAUTH_URL to FRONTEND_URL migration too. I traced it across CI, Helm charts, and infra scripts and didn't find any stragglers.

PR #2183, opened 2026-07-16
Fix session refresh with BFF pattern and short-lived access tokens

harry-rhesis:
## Purpose
Access tokens lived for 7 days as a workaround for a broken refresh flow, and the browser held the backend access token directly. This PR activates the full refresh-token flow (15-minute access tokens, 7-day stable refresh tokens) and moves all client API traffic behind a BFF proxy so no backend token ever reaches browser JS.

## What Changed
- **BFF proxy**: client calls go through same-origin `/api/backend/*`; the route injects `Authorization` server-side from the httpOnly JWE session cookie. Server components/actions use `getFreshAccessToken()` directly.
- **Refresh flow**: proactive refresh 60s before expiry, coalesced per refresh token (in-flight map + 5s result cache). Refreshed cookies are persisted from both the proxy route **and** the middleware — the middleware previously used zero-arg `auth()`, which structurally discards Set-Cookie, causing repeated refreshes on every navigation.
- **Auth codes**: login redirects now carry an opaque single-use code; the session/refresh tokens it references are stored server-side in Redis (sha256-keyed, GETDEL, 60s TTL) instead of being embedded in the URL.
- **Security hardening**: 401 origin disambiguation (`x-auth-origin` header) so an expired proxy session signs out cleanly instead of revoking the refresh-token family; Sec-Fetch-Site cross-site rejection on the proxy; open-redirect fix in the NextAuth redirect callback; server-side following of same-origin backend redirects (trailing-slash 307s no longer bypass auth); cookie-chunking size guard.
- **Session-status predicates**: `isAuthenticated`/`isSessionLoading`/`isSessionUnauthenticated` replace 187 literal `status === '...'` comparisons across 105 files.
- **Tests**: new middleware routing-contract suite (16 tests, node jest environment), backend opaque-auth-code tests, E2E no-docker seed fixed to produce a decryptable JWE.

## Additional Context
- UI logins mint stable (non-rotating) refresh tokens (`client_id IS NULL`); rotating tokens with family reuse detection remain for token exchange.
- Requires `JWT_ACCESS_TOKEN_EXPIRE_MINUTES=15` (code default) — remove any 10080 override from deployed envs.
- Redis unavailability falls back to legacy JWT-embedded auth codes (loudly logged), so logins keep working.

## Testing
- Frontend: 181 jest suites / 1809 tests, `tsc` and eslint clean
- Backend: 419 auth tests, ruff clean
- Manual: login, expiry-boundary navigation (single refresh, then none), logout, revoked-refresh-token path
- E2E no-docker: seed round-trip verified (encode → `getToken` decode → local verification); full Playwright run needs port 8080 free


harry-rhesis:
@nicolai-rhesis, would you kindly take a look at this implementation? This substantially upgrades our authentication layer to become more secure. And lays the foundation to avoid having to expose the backend in local deployments.

PR #2191, opened 2026-07-17
Deploy K8s Production

asad-miah:
This PR introduces changes from the `feature/deploy-k8s-production` branch.

## 📁 Files Changed (      23 files)

```
.gitignore
charts/rhesis/values-prd.yaml
kubernetes/bootstrap/argocd/kustomization.yaml
kubernetes/bootstrap/argocd/patches/argocd-accounts-cm.yaml
kubernetes/clusters/prd/cluster-dns/kube-dns.yaml
kubernetes/clusters/prd/cluster-dns/patch-job.yaml
kubernetes/clusters/prd/cnpg-cluster/cnpg-cluster.yaml
kubernetes/clusters/prd/cnpg-operator.yaml
kubernetes/clusters/prd/grafana-resources/grafana-replicas.yaml
kubernetes/clusters/prd/grafana-resources/kustomization.yaml
kubernetes/clusters/prd/internal-dns/kustomization.yaml
kubernetes/clusters/prd/kustomization.yaml
terraform/infrastructure/envs/prd/main.tf
terraform/infrastructure/modules/kubernetes/gcp/firewall.tf
terraform/infrastructure/modules/kubernetes/gcp/main.tf
terraform/infrastructure/modules/kubernetes/gcp/variables.tf
terraform/infrastructure/modules/wireguard/gcp/bind9-update.tf
tests/k6/common.js
tests/k6/load.js
tests/k6/README.md
tests/k6/soak.js
tests/k6/spike.js
tests/k6/stress.js
```

## 📋 Commit Details

```
5d80ca8f9 - feat(dev): configure prod infra and load testing (Md Asaduzzaman Miah, 2026-07-17 13:04)
```

## ✅ Checklist

- [ ] All tests pass locally


terraform/infrastructure/modules/kubernetes/gcp/firewall.tf
```diff
@@ -110,3 +110,39 @@ resource "google_compute_firewall" "gke_lb_health_checks" {
...
+  network  = var.vpc_name
+  project  = var.project_id
+  priority = 900
+
+  allow {
+    protocol = "tcp"
+    ports    = ["80", "443"]
+  }
+
+  source_ranges = local.public_ingress_source_ranges
```
harry-rhesis, line 146: Worth noting in the comment above that Cloudflare's edge ranges are shared across all their customers, so this rule blocks arbitrary direct-to-origin traffic but does not authenticate that requests came through our Cloudflare zone. Anyone who learns the origin IP could still reach it via their own Cloudflare-proxied traffic. If we want that guarantee later, Authenticated Origin Pulls (mTLS) or a secret header checked at ingress-nginx are the way. Happy to ship this as-is for now as defense-in-depth, just let's frame it that way in the variable description so it is not mistaken for a lockdown.

terraform/infrastructure/modules/wireguard/gcp/bind9-update.tf
```diff
@@ -105,11 +105,10 @@ resource "terraform_data" "bind9_config_update" {
...
           set -e && \
-          apt-get install -y -qq bind9 > /dev/null 2>&1 || true && \
+          apt-get install -y -qq bind9 > /dev/null 2>&1 && \
           systemctl stop dnsmasq 2>/dev/null || true && \
           systemctl disable dnsmasq 2>/dev/null || true && \
           base64 -d /tmp/named.conf.tf.b64 > /tmp/named.conf.new && \
-          rm /tmp/named.conf.tf.b64 && \
           [ -s /tmp/named.conf.new ] || { echo \"ERROR: decoded named.conf is empty\"; rm -f /tmp/named.conf.new; exit 1; } && \
           named-checkconf /tmp/named.conf.new && \
           mv /tmp/named.conf.new /etc/bind/named.conf && \
```
harry-rhesis, line 114: Removing the old `rm /tmp/named.conf.tf.b64` here was the right fix, since the whole block is retried by gcloud_retry and the b64 file has to survive for a retry after a failed named-checkconf or mv. Only downside is the decoded config (which can carry TSIG secrets) now lingers in /tmp until the next run's stale-file cleanup. Low risk on an IAP-only host, but if you want to close it, add `rm -f /tmp/named.conf.tf.b64 && \` right after the successful `mv` so it only clears on success and retry safety is preserved.

harry-rhesis (commented):
Thanks for putting this together, Asad. This is a clean, well-scoped change and the comments on the non-obvious parts are really helpful. The priority-900-vs-1000 note explaining the tie-break against GKE's auto-created per-Service rule is exactly the kind of context that saves the next person a lot of time, and having both new flags opt-in and off by default keeps existing environments safe.

Two small things worth a look before merge, neither a blocker:

1. The one that matters most is not really a code bug but a security expectation: restricting the firewall to Cloudflare's edge ranges keeps out arbitrary direct-to-origin scanning, but those ranges are shared across all Cloudflare customers, so it does not prove traffic actually came through our zone. If we want true origin protection the real controls are Authenticated Origin Pulls (mTLS) or a secret-header check at ingress-nginx. Fine to ship as defense-in-depth, just worth being explicit that it is not origin authentication.

2. A minor cleanup in the BIND9 provisioner (inline).

I also confirmed the dropped `rm` in bind9-update.tf was the right call for retry safety, and removing `|| true` from the apt-get line is a genuine improvement. Nice work overall.

terraform/infrastructure/modules/kubernetes/gcp/firewall.tf
asad-miah: added description for reference the reference for introducing secret header.

terraform/infrastructure/modules/wireguard/gcp/bind9-update.tf
asad-miah: added rm -f to cleanup named.conf.tf.b64 file.

PR #2194, opened 2026-07-17
Add Annotations review hub

nicolai-rhesis:
## Purpose
Make human reviews findable under Improve → Annotations, with resolve/search/filter and deep-links back to test results and traces.

## What Changed
- Backend `GET /annotations/` flattens `test_reviews` / `trace_reviews` with search, rating, resolved, and target filters
- Dual-gate auth: `test_result:read` and/or `telemetry:read` (nav uses `requiredAnyOf`)
- Frontend Annotations page/grid with Open/Resolved pills, filter drawer, and new-tab row opens
- Mark reviews resolved/reopened on test results and traces; invalidate annotations cache on create/resolve/delete
- Deep-link to test-run Reviews tab; mention rendering polish

## Additional Context
- Scoped slice: hub + list + resolve (no custom forms, queues, or playground annotate yet)
- Trace Reviews UI remains gated on `trace_metrics` (pre-existing)

## Testing
- [ ] Open Improve → Annotations; confirm list loads with rating/status
- [ ] Filter Open/Resolved pills and drawer filters (type, rating, target)
- [ ] Search by comment/annotator; row click opens source in a new tab on Reviews
- [ ] Resolve/Reopen a review; Annotations grid updates without a hard refresh
- [ ] `cd apps/backend && uv run pytest ../../tests/backend/routes/test_annotations.py -v`
- [ ] Frontend: `npm test -- --testPathPatterns=NavSection.test --no-coverage`


nicolai-rhesis:
Pushed behavior column for test-result annotations (SQL join, no N+1), target label **Output**, and column order Annotator → Behavior.

harry-rhesis:
First off, thanks @nicolai-rhesis, this is a genuinely nice piece of work. The dual-gate auth, the `requiredAnyOf` nav plumbing, the deep-links back into the Reviews tab, the resolve/reopen flow, and the docs are all well thought through, and the backend tenant-scoping in the raw SQL correctly follows our ambient-scope rules. None of the feedback below is about execution quality.

## Recommendation: hold, let's not merge this as-is

Reading through the PR, almost every awkward corner traces back to the same root cause: reviews are JSONB blobs nested inside `test_result.test_reviews` / `trace.trace_reviews` rather than first-class rows.

- `list_annotations` has to hand-write `organization_id` / `project_id` predicates because `db.execute(text(...))` bypasses the ORM auto-filter.
- `populate_review_permitted_actions` exists only because JSONB sub-documents never flow through a `WithPermittedActions` schema.
- The `AUTHZ_EXEMPT_ROUTES` exemption plus the in-handler dual-gate (with a bare `"telemetry:read"` string, since there is no `Permission.Telemetry` enum) is a workaround for a route that cannot use the standard router-derived RBAC backstop.
- Search is `ILIKE` over `jsonb_array_elements` LATERAL scans with `UNION ALL` and `COUNT(*) OVER()`, no index path.
- Mutations go through `flag_modified` plus dict-key poking instead of a normal UPDATE.

This also lines up with our broader goal of **making annotations a first-class entity in Rhesis**: something you can attach to any entity (tests, test results, traces, and beyond), not just the two places reviews live today. So rather than building more surface area on top of the JSONB model, I want to use this as the moment to introduce that entity. If we merge the hub against the current model, we would be migrating the storage and unwinding this router/service/exemption shortly after.

## Proposed course of action

1. **Introduce a first-class `annotation` entity** using our established polymorphic pattern, the same `(entity_type, entity_id)` pair validated by the shared `EntityType` enum that `Comment`, `Task`, `Tag`, `File`, `Embedding`, and `TestContext` already use (with the standard `GET /entity/{entity_type}/{entity_id}` route). No reason to invent a new shape here; consistency with the six existing polymorphic models wins, and we accept the same app-layer integrity tradeoff they do rather than DB-level FKs. For the sub-entity target (a specific metric or turn inside a result) we already have the `ReviewTarget` enum (`test_result` / `trace` / `turn` / `metric`), so the annotation carries a target descriptor (`target_type` + `target_reference`) reusing that, layered on top of the `entity_type` / `entity_id` parent link.
2. **Preserve the two subtleties the current reviews carry.** These are the crux, not the table itself:
   - *Override semantics*: a test review is not just a comment, it overrides the automated pass/fail (`apply_review_override`, `original_status_id` snapshotting, conflict detection). The entity needs an explicit home for that, not a silent drop.
   - *Sub-entity targets*: today's `target: {type, reference}` points inside a result at a specific metric or turn, so the parent `entity_type` / `entity_id` link is not enough on its own; it needs the `ReviewTarget`-based target descriptor above alongside it.
3. **Migrate** existing `test_reviews` / `trace_reviews` JSONB into `annotation` rows, keeping the JSONB readable during transition as a safety net.
4. **Repoint the review drawers and this hub at the new table.** At that point the hub collapses into a standard CRUD list: no raw SQL, no dual-gate exemption, no affordance projection. It inherits tenant filtering, affordances, and RBAC for free.

## The heavy lifting (eyes open before we commit)

The entity table is the easy part. The real work is unwinding two years of "reviews live inside their parent" assumptions. Rough blast radius: ~11 backend files / ~74 refs and ~10 frontend files / ~25 refs touch `test_reviews` / `trace_reviews`.

**Backend**

- Data migration is the highest-risk item: read every `test_result.test_reviews.reviews[]` and `trace.trace_reviews.reviews[]`, fan each element out into an `annotation` row, map `status` / `user` / `target` / `resolved` / timestamps, and preserve `metadata.original_status_id` so conflict detection still works. Needs to be idempotent and reversible.
- `services/review.py`, `routers/test_result.py`, and `routers/telemetry.py` currently mutate the JSONB in place (`apply_review_override`, `update_review_metadata`, `apply_review_resolved`, `flag_modified`). All of that has to move to row operations, and the override path stays entangled with `test_result.status_id`, so it needs care rather than a mechanical port.
- `crud.py`, `models/test_result.py`, `models/trace.py`, `schemas/*`, `services/telemetry/tree_builder.py`, and `tasks/execution/result_processor.py` all read or write the embedded shape and each needs revisiting.

**Frontend (this is the larger and trickier half)**

Today the UI never fetches reviews separately: it reads them straight off the parent object it already has (`result.test_reviews?.reviews`, `trace.trace_reviews`). Making annotations first-class means that embedded array goes away, and every consumer has to move to a fetched-and-cached resource. That is a structural change, not a rename.

- `test-run-summary-utils.ts` (and its test) is the core derivation layer: `getResultReviews`, `getLatestMetricReviewForResult`, and the merge with `review_summary` all iterate `result.test_reviews.reviews`. This becomes "join the fetched annotations for this result," which changes the data-flow shape, not just the field name.
- Review state is consumed well beyond the Reviews tabs: `TestDetailReviewsTab`, `TestDetailMetricsTab`, `TestDetailConversationTab`, `TraceReviewsTab`, `TraceMetricsTab`, and `ConversationTraceView` all read the embedded reviews to drive conflict badges, per-metric and per-turn overrides, and status chips. Each needs rewiring to the new source and its own cache-invalidation story (the PR already hints at this pain with the manual `getTestResult` refetch to backfill missing `permitted_actions`).
- Loading and empty states change everywhere reviews were previously "just there" synchronously on the parent; components that assumed the array was present now have to handle an async, independently-cached fetch.
- `interfaces/test-results.ts` and `interfaces/telemetry.ts` embed `Review` / `TraceReview` inside the parent types; those move to a standalone annotation interface, and every call site is type-checked against the change.
- The good news: none of the UX in this PR is wasted. The grid, filter drawer, deep-links, resolve/reopen flow, and mention rendering all carry straight over. They just point at a cleaner backend and read from one source of truth instead of two embedded shapes.

## Suggestion

Let's lift the product thinking and UI from this PR onto the new entity rather than merging it against the JSONB model. Happy to pair on the schema and the migration outline, and to sequence the frontend cutover so the tabs move over incrementally rather than in one big bang.

Really solid effort. I mainly want to make sure we build it on the foundation we actually want. 🙏

apps/backend/src/rhesis/backend/app/services/annotations.py
```diff
@@ -0,0 +1,219 @@
...
+) -> tuple[list[AnnotationListItem], int]:
+    """Return flattened reviews with correct row-level pagination.
+
+    Each source branch is included only when the corresponding flag is True.
+    Explicit org/project predicates are required because this uses ``text()``
+    SQL (ORM auto-filter does not apply).
+    """
+    if source == "test_result":
+        include_traces = False
+        include_test_results = True
```
harry-rhesis: This block doesn't just narrow to one source, it *forces* the matching flag to `True` regardless of what the caller passed in. It's safe today only because the router 403s before ever calling this with a `source` the caller can't read, but the service itself has no defense-in-depth: a future caller (export job, another endpoint) passing `source="trace"` with `include_traces=False` would silently get trace data anyway.

Suggest only forcing the *excluded* branch off, and letting the caller's permission-derived flag stand:

```python
if source == "test_result":
    include_traces = False
elif source == "trace":
    include_test_results = False
```

apps/backend/src/rhesis/backend/app/routers/annotations.py
```diff
@@ -0,0 +1,108 @@
...
+from rhesis.backend.app.services.annotations import list_annotations
+
+router = RhesisRouter(
+    prefix="/annotations",
+    tags=["annotations"],
+    responses={404: {"description": "Not found"}},
+)
+
+TEST_RESULT_READ = Permission.TestResult.READ
+TELEMETRY_READ = "telemetry:read"
```
harry-rhesis: This is the only in-handler `authorize()` call in the codebase using a raw permission string; every other explicit `authorize()` / `authorize_object()` site uses a `Permission.X.Y` enum member. `Permission` has no `Telemetry` class today. `authorize()` fails closed, so a typo (`"telemtry:read"`) would produce an unexpected 403 rather than a bypass, but it's unenumerated and not caught by any static/drift check.

Suggest adding `Permission.Telemetry.READ` to `auth/capabilities.py` and referencing it here, matching the sibling telemetry routes that derive `telemetry:read` from their router `resource=` stamp.

harry-rhesis (commented):
Following up on the discussion with @nicolai-rhesis: we're aligned to land this against the current model and tackle the unified first-class `annotation` entity as a separate future effort (tracked in my earlier comment). This review is scoped to what should be addressed before we merge, not the larger refactor.

**Please address before merge:**
1. The `list_annotations` flag override (defense-in-depth in the service layer).
2. The bare `"telemetry:read"` string (add a `Permission.Telemetry.READ` enum member).
3. A negative test for the dual-gate 403 branches.

**Nice-to-have / follow-up:** the `encodeURIComponent` on deep-links, the performance note on the LATERAL scans, and the incidental scope-creep changes (call your judgement on whether to split those out).

Inline comments below. Really nice work overall, the UX and product thinking carry straight over to the entity approach later. 🙏

apps/backend/src/rhesis/backend/app/routers/annotations.py
```diff
@@ -0,0 +1,108 @@
...
+    """
+    organization_id, _user_id = tenant_context
+    principal = resolve_principal_from_request(current_user, request)
+    project_id = project_id_from_scope(db)
+    project_id_str = str(project_id) if project_id else None
+
+    can_read_test_results = authorize(principal, TEST_RESULT_READ, project_id=project_id, db=db)
+    can_read_traces = authorize(principal, TELEMETRY_READ, project_id=project_id, db=db)
+
+    if not can_read_test_results and not can_read_traces:
```
harry-rhesis, line 78: These three 403 branches are the unusual part of this endpoint (manual in-handler authorization instead of the router-derived backstop), but nothing in `test_annotations.py` exercises them. Worth a couple of negative tests: (a) caller with neither `test_result:read` nor `telemetry:read` gets 403, and (b) `source=trace` without `telemetry:read` gets 403. Exactly the branches most worth locking down given the bespoke auth.

apps/backend/src/rhesis/backend/app/services/annotations.py
```diff
@@ -0,0 +1,219 @@
...
+                elem AS review,
+                COALESCE(elem->>'updated_at', elem->>'created_at') AS sort_at
+            FROM test_result tr
+            LEFT JOIN test tst
+              ON tst.id = tr.test_id
+             AND tst.deleted_at IS NULL
+            LEFT JOIN behavior b
+              ON b.id = tst.behavior_id
+             AND b.deleted_at IS NULL
+            CROSS JOIN LATERAL jsonb_array_elements(tr.test_reviews->'reviews') AS elem
```
harry-rhesis, line 114: Note for later (not blocking): this flattens via `LATERAL jsonb_array_elements` over every test result / trace row with a non-empty reviews array on each list request, with no index path and `ILIKE` search on JSONB text. Fine at current review volumes; worth revisiting if review counts grow. The future `annotation` entity naturally resolves this with real indexes.

apps/frontend/src/app/(protected)/annotations/components/AnnotationsGrid.tsx
```diff
@@ -0,0 +1,410 @@
...
+      },
+    ],
+    [theme]
+  );
+
+  const handleRowClick = useCallback((params: GridRowParams) => {
+    const row = params.row as AnnotationListItem;
+    let url: string | null = null;
+    if (row.source === 'test_result' && row.test_run_id && row.test_result_id) {
+      url = `/test-runs/${row.test_run_id}?selectedresult=${row.test_result_id}&detailTab=reviews`;
```
harry-rhesis: Minor: these interpolate `test_run_id` / `test_result_id` / `trace_id` / `project_id` straight into the query string without `encodeURIComponent`. Low risk while these are opaque UUID/hash-like IDs, but not defensive if a value ever contains a reserved character. Cheap to wrap now.

apps/frontend/src/components/common/MentionTextInput.tsx
```diff
@@ -389,6 +383,32 @@ export default function MentionTextInput({
   );
 }
 
+/**
+ * Readable mention colors — metrics use primary (not secondary orange).
+ */
+export function getMentionTypeStyles(theme: {
```
harry-rhesis, line 389: Nit / scope: this mention-color refactor (and the `data_factories.py` long-name fix) are unrelated to the annotations hub. No objection to the changes themselves, but per our one-logical-change-per-PR guideline they'd ideally be split out, especially since this PR is already ~44 files. Your call given we're landing this soon.

apps/backend/src/rhesis/backend/app/services/annotations.py
nicolai-rhesis: Addressed — `source` now only turns the excluded branch off and leaves the caller's permission-derived include flag intact.

apps/backend/src/rhesis/backend/app/routers/annotations.py
nicolai-rhesis: Added `Permission.Telemetry` (READ/CREATE/UPDATE/DELETE, matching the seeded catalog + frontend) and wired the router to `Permission.Telemetry.READ`.

apps/backend/src/rhesis/backend/app/routers/annotations.py
nicolai-rhesis: Added `TestAnnotationsDualGateAuth` with both negative cases: neither read permission → 403, and `source=trace` without `telemetry:read` → 403.

apps/backend/src/rhesis/backend/app/services/annotations.py
nicolai-rhesis: Agreed — leaving as-is for this PR; the future first-class `annotation` entity is the right place to get real indexes.

apps/frontend/src/app/(protected)/annotations/components/AnnotationsGrid.tsx
nicolai-rhesis: Done — deep-link IDs are now wrapped with `encodeURIComponent`.

apps/frontend/src/components/common/MentionTextInput.tsx
nicolai-rhesis: Leaving in this PR given we're landing soon — happy to split in a follow-up if you'd rather keep the hub diff tighter.

nicolai-rhesis:
Addressed the pre-merge items from [@harry-rhesis](https://github.com/harry-rhesis)'s review in \`da84e0e24\`:

1. **Service flag override** — \`source\` only forces the *excluded* branch off
2. **\`Permission.Telemetry.READ\`** — enum added; router no longer uses a bare string
3. **Negative dual-gate tests** — neither-permission and \`source=trace\` without telemetry:read
4. **Nice-to-have** — \`encodeURIComponent\` on annotation deep-links

Left the LATERAL/perf note and the incidental mention-color/data_factories changes as follow-ups (as suggested).

apps/backend/src/rhesis/backend/app/routers/annotations.py
```diff
@@ -0,0 +1,108 @@
...
+    current_user: User = Depends(require_current_user_or_token),
+) -> List[AnnotationListItem]:
+    """List human annotations (reviews) across test results and traces.
+
+    Dual-gated: callers need ``test_result:read`` and/or ``telemetry:read``.
+    Each source branch is included only when the matching permission is present.
+    """
+    organization_id, _user_id = tenant_context
+    principal = resolve_principal_from_request(current_user, request)
+    project_id = project_id_from_scope(db)
```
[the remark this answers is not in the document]
nicolai-rhesis: Agreed — `/annotations/` now fails closed with **400** when `project_id_from_scope()` is `None`, so we never widen to an org-wide list. Covered by `test_requires_project_scope`.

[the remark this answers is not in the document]
nicolai-rhesis: Addressed in the follow-up: missing ambient project → 400 (fail closed).

apps/backend/src/rhesis/backend/app/schemas/annotation.py
```diff
@@ -0,0 +1,40 @@
...
+
+    # Deep-link fields
+    test_result_id: Optional[UUID] = None
+    test_run_id: Optional[UUID] = None
+    trace_db_id: Optional[UUID] = None
+    trace_id: Optional[str] = None
+    project_id: Optional[UUID] = None
+    span_name: Optional[str] = None
+    # From the linked test (test results only; null for traces)
+    behavior_id: Optional[UUID] = None
```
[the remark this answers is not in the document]
nicolai-rhesis: Added defense-in-depth `_as_resolved()`: only real bools / true-like strings count as resolved (so `"false"` stays open).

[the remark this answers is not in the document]
nicolai-rhesis: Fixed via `_as_resolved()` — `"false"` no longer coerces to `True`.

apps/frontend/src/utils/api-client/annotations-client.ts
```diff
@@ -0,0 +1,58 @@
...
+    }
+    if (params.target_type) {
+      queryParams.append('target_type', params.target_type);
+    }
+
+    const path = `${API_ENDPOINTS.annotations}/?${queryParams.toString()}`;
+    const url = this.baseUrl + path;
+    const headers = this.getHeaders();
+
+    const response = await fetch(url, {
```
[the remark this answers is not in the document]
nicolai-rhesis: Switched `AnnotationsClient.getAnnotations` to `this.fetchPaginated()`, matching other list clients.

PR #2197, opened 2026-07-19
Unify coding agent instructions across Claude Code and Cursor

harry-rhesis:
## Purpose

We maintained parallel, largely duplicated instruction sets for Claude Code (`CLAUDE.md`) and Cursor (`.cursor/rules/*.mdc`). The same rules were written twice, drifted apart, and loaded on every turn regardless of relevance — costing tokens and inviting inconsistency. This PR establishes a single source of truth for both tools and scopes rules so they load only when relevant.

## What Changed

**Rules — one source, read by both tools:**
- Added a nested `AGENTS.md` hierarchy (read natively by Cursor, imported by Claude Code): root `AGENTS.md` for repo-wide rules, plus `apps/backend/`, `apps/frontend/`, `sdk/`, and `docs/` for area-specific rules.
- Each area also has a one-line `CLAUDE.md` (`@AGENTS.md`) so Claude Code picks up the same content.
- Root `CLAUDE.md` trimmed from 564 lines to a thin `@AGENTS.md` import plus pointers.
- Removed all 11 `.cursor/rules/*.mdc` files — their content now lives once in the `AGENTS.md` files. Also fixes the `documentation.mdc` bug where its entire body was duplicated.
- Scoped rules now load on demand (backend rules only when touching `apps/backend/`, etc.) instead of everything on every turn.

**Skills — tool-agnostic canonical home:**
- Canonical skills live in a neutral `.agents/skills/` directory (`python-linting`, `sprint-review`).
- `.cursor/skills/` and `.claude/skills/` hold per-skill symlinks back to `.agents/skills/`, the arrangement both tools actually support (Cursor follows per-skill symlinks inside a real project-level dir, but not a symlinked skills dir itself).
- `.agents/README.md` documents the convention.

Net: **+567 / −2,612 lines** with no loss of coverage.

## Additional Context

- The distributable `skills/rhesis/` plugin and `.claude/commands/` slash commands are unchanged — the latter has no Cursor equivalent to share.
- **Windows note:** the skill symlinks require `git config core.symlinks=true` (default with Git for Windows) to materialize as real links; otherwise they appear as text files. This matches the repo's existing `rh worktree` `.env` symlink convention.

## Testing

- Verified both `.cursor/skills/<name>/SKILL.md` and `.claude/skills/<name>/SKILL.md` resolve through the symlinks to `.agents/skills/`.
- Confirmed Claude Code discovered the `python-linting` skill live through the symlink during development.
- Confirmed git records the tool-side skill entries as symlinks (mode `120000`) sharing identical blob hashes.


harry-rhesis:
@nicolai-rhesis @akwasigroch please take a look at this if you have time :)

akwasigroch:
It's a great direction. 

- sprint-review skill can be removed, because we are not using it (there is a separate repo for the sprint review for the whole team)

We should add more skills in the future to speed up our development - https://github.com/comet-ml/opik/tree/main/.agents/skills here is a nice list that we can take inspiration from

nicolai-rhesis (approved):
Approved.

PR #2202, opened 2026-07-20
feat: OWASP Top 10 test set generation (multi-turn, tagging)

alexeyhramov123:
## Purpose

Turns the OWASP synthesizer (merged earlier in #2054) into an end-to-end
feature: users can generate an OWASP Top 10 test set (LLM or Agentic
framework, single- or multi-turn) directly from the Test Sets page, and
the resulting behaviors/metrics are tagged and filterable as "OWASP".

## Scope

- **Backend API**: new `GET /owasp/categories` and `POST /owasp/generate`
  router (`apps/backend/.../routers/owasp.py`), backed by an `owasp`
  service that downloads/parses the official OWASP PDF and caches the
  parsed report (Redis).
- **Generation task**: `generate_and_save_owasp_test_set` Celery task
  (`tasks/test_set.py`) drives the SDK's `OWASPSynthesizer` to produce a
  Rhesis test set from the selected categories, including a new
  multi-turn generation path.
- **Metrics/behaviors seeding**: `initial_data.json` gains the "OWASP LLM
  Top 10" / "OWASP Agentic Top 10" behaviors and their 20 associated
  metrics. Three Alembic migrations backfill these into **existing**
  organizations and tag them (new orgs get them via onboarding as usual):
  - `38ed899b9f41` — sync OWASP behaviors + metrics to existing orgs
  - `30877432d102` — add `owasp:create` / `owasp:read` permission rows
  - `b857edcac3c0` — tag existing OWASP metrics/behaviors with an `OWASP` tag
- **Frontend**: `OwaspGenerateDrawer` on the Test Sets page (framework,
  purpose, categories, single-/multi-turn toggle), an `OwaspIcon`, an
  API client, and an "OWASP" filter on the Metrics directory tab.
- **SDK**: `OWASPSynthesizer` gains multi-turn generation
  (`owasp_synthesizer_multi_turn.jinja`) and caches parsed OWASP reports.

## How to run this locally

1. From the repo root, bring up Postgres/Redis once and start them:
   ```bash
   ./rh dev init   # first time only
   ./rh dev up
   ```
2. Install backend deps and apply the migrations (from `apps/backend`):
   ```bash
   uv sync --dev --extra all --extra cpu
   source .venv/bin/activate
   cd src/rhesis/backend
   uv run alembic upgrade head
   ```
   This runs the three OWASP migrations above against your existing local
   organizations (idempotent — safe to re-run). New organizations get the
   same behaviors/metrics automatically via onboarding, no migration needed.
3. Start the backend and a worker (repo root, separate terminals):
   ```bash
   ./rh dev backend
   ./rh dev worker
   ```
4. Start the frontend as usual and open **Test Sets** → **Generate** →
   **OWASP**. Pick a framework (LLM / Agentic Top 10), describe the
   system under test, optionally narrow to specific categories, and
   toggle single- vs multi-turn.
5. To check the tag filter: **Metrics** directory tab → filter by
   **OWASP** should show the seeded metrics.

## Testing

- `cd apps/backend && make test` (backend unit tests, spins up
  Postgres/Redis via docker compose automatically)
- `cd sdk && make test` (SDK unit tests — note: `tests/sdk/` has no
  dedicated `OWASPSynthesizer` coverage yet, corrected from an earlier
  version of this description that claimed otherwise)
- Manual: run the flow in step 4 above end-to-end for both frameworks
  and both single-/multi-turn modes.

## Notes for reviewers

- This branch previously shipped #2054 (already merged into `main`);
  this PR only contains the 14 commits added since then.
- `2a2b2d0f0` is a small fixup for a two-head Alembic chain introduced
  while rebasing onto `main` — no behavioral change.


apps/backend/src/rhesis/backend/app/main.py
```diff
@@ -463,6 +463,34 @@ def _log_task_exception(t: asyncio.Task) -> None:
 
         setup_mcp_server(app)
 
+    # Pre-warm OWASP report section cache in background (non-blocking)
```
harry-rhesis, line 490: **Correctness: this will slow down, and possibly hang, the backend test suite.**

This pre-warm fires two PDF downloads plus two pdfminer parses through `asyncio.to_thread` on every
app lifespan start. The Garak pre-warm 30 lines above is skipped under tests via
`RHESIS_SKIP_GARAK_WARM_CACHE`, set in `tests/backend/conftest.py:94`, and its comment spells out
the exact failure mode: every test builds a fresh `TestClient`, so a fresh lifespan, the worker
threads are not cancellable, and the pile up snowballs across pytest-xdist workers.

The OWASP warm has the same shape and no guard, so it inherits the same problem plus real network
I/O per test.

Suggestion: put both behind one flag, for example `RHESIS_SKIP_WARM_CACHES`, keeping
`RHESIS_SKIP_GARAK_WARM_CACHE` as an alias so nothing that sets it today breaks. Adding
`RHESIS_SKIP_OWASP_WARM_CACHE` to the conftest env block works too, it is just a second knob to
remember.

apps/backend/src/rhesis/backend/app/main.py
```diff
@@ -463,6 +463,34 @@ def _log_task_exception(t: asyncio.Task) -> None:
...
+            )
+
+            for framework in OWASP_FRAMEWORKS:
+                await asyncio.to_thread(list_category_summaries, framework)
+        except Exception as e:
+            logger.warning(f"OWASP cache pre-warming failed (non-fatal): {e}")
+
+    owasp_cache_task = asyncio.create_task(warm_owasp_cache())
+
+    def _log_owasp_task_exception(t: asyncio.Task) -> None:
```
harry-rhesis: Small one: this is character for character `_log_task_exception` at line 451, apart
from the log string. Worth one helper that takes the label:

```python
def _log_warm_task_exception(label: str):
    def _cb(t: asyncio.Task) -> None:
        try:
            t.result()
        except asyncio.CancelledError:
            pass  # Expected during shutdown
        except Exception as e:
            logger.error(f"{label} cache pre-warming task failed: {e}", exc_info=True)
    return _cb
```

apps/backend/src/rhesis/backend/tasks/test_set.py
```diff
@@ -634,3 +634,155 @@ def generate_and_save_test_set(
...
+            "OWASP test set generated",
+            actual_tests_generated=len(test_set.tests),
+            requested_tests=num_tests,
+            generation_time_seconds=round(gen_elapsed, 1),
+        )
+
+        self.update_state(state="PROGRESS", meta={"status": "Saving to database"})
+
+        test_set_name = name or f"{behavior}: {purpose[:60]}"
+        db_test_set = _save_test_set_to_database(
```
harry-rhesis, line 745: **Correctness: the duplicate test set bug is still reachable.**

`f7fe037` fixed the trigger, the missing `tests_generated` kwarg, but not the class of bug. This
save commits a new row, and everything after it (`_build_task_result`, the logging calls) still runs
inside the same `try`, under `autoretry_for = (Exception,)` and `max_retries = 3`
(`tasks/base.py:69,52`). So any *other* post save failure still retries the whole task body and
still leaves up to 4 test sets behind. The next `_build_task_result` signature change, or a
`DetachedInstanceError` of the kind its own docstring describes, reproduces the original incident.

The sibling task is not exposed to this because the router pre-creates the row and passes
`test_set_id`, so retries reattach to the same row via `_attach_tests_to_existing_test_set` instead
of creating a new one.

Two ways out, in the order I would pick them:

1. Adopt the pre-created row path. Router creates the `TestSet`, task attaches. It matches the
   sibling, and it gives the frontend an id to redirect to immediately, which the OWASP drawer would
   probably want anyway.
2. Keep the current shape but move the post save work out of the retryable region, or raise
   something non-retryable for it. `dont_autoretry_for` already exists for exactly this
   (`tasks/base.py:75`).

apps/backend/src/rhesis/backend/tasks/test_set.py
```diff
@@ -634,3 +634,155 @@ def generate_and_save_test_set(
...
+        )
+
+        self.log_with_context(
+            "info",
+            "Task completed successfully",
+            test_set_id=str(db_test_set.id),
+            tests_generated=len(test_set.tests),
+        )
+
+        return result
```
harry-rhesis: **Correctness: OWASP generation is not metered.**

Nothing in this task dispatches a quota accrual, so OWASP generated tests do not count against the
org's test generation quota. Every other generation path does:

- `tasks/test_set.py:588`, the sibling task:
  `dispatch_accrual(org_id, QuotaResource.TEST_GENERATION, len(test_set.tests))`
- `tasks/garak.py:87` and `tasks/garak.py:133`, for probe import and sync

`dispatch_accrual` is already imported at the top of this module, so it is a one line addition after
the save, matching line 588.

apps/backend/src/rhesis/backend/app/services/owasp.py
```diff
@@ -0,0 +1,197 @@
...
+    (e.g. Docker's ``file:///app/storage`` on a Mac host) does not discard a
+    successful parse — Redis still gets the id/name metadata for the picker.
+    """
+    storage = _get_storage()
+    payload = json.dumps(sections).encode("utf-8")
+    try:
+        storage.put_object_bytes(
+            payload, storage.get_owasp_content_path(framework), "application/json"
+        )
+    except OSError as e:
```
harry-rhesis: **Correctness: this except is too narrow to keep the promise in the docstring.**

The docstring says a storage write failure will not discard a successful parse, but `OSError` only
covers the local filesystem case it mentions. `StorageService.put_object_bytes` goes through fsspec,
and the cloud backends raise types that are not `OSError` subclasses (botocore `ClientError`,
gcsfs `HttpError`). Those propagate out of `fetch_owasp_sections`, out of
`list_category_summaries`, and land as a 502 for the user *after* the expensive parse already
succeeded, which is the opposite of the intent here.

```suggestion
    except Exception as e:
```

apps/backend/src/rhesis/backend/app/services/owasp.py
```diff
@@ -0,0 +1,197 @@
...
+    DEFAULT_OWASP_AGENTIC_PDF_URL,
+    DEFAULT_OWASP_LLM_PDF_URL,
+    ReportSection,
+    fetch_owasp_sections,
+)
+
+logger = logging.getLogger(__name__)
+
+# Framework id -> report URL / behavior label stamped on generated tests.
+OWASP_FRAMEWORKS: Dict[str, Dict[str, str]] = {
```
harry-rhesis, line 34: DRY: `behavior` and `label` hold identical strings for both frameworks. The router
reads `label` (`routers/owasp.py:117`) and the task reads `behavior` (`tasks/test_set.py:693`),
which as far as I can tell is accidental rather than meaningful. Worth collapsing to one key so the
two cannot drift apart later.

apps/backend/src/rhesis/backend/app/services/owasp.py
```diff
@@ -0,0 +1,197 @@
...
+            if len(" ".join(collected)) >= max_len:
+                break
+
+    text = " ".join(collected).strip()
+    if len(text) > max_len:
+        text = text[: max_len - 1].rsplit(" ", 1)[0] + "…"
+    return text
+
+
+def list_categories(framework: str) -> List[ReportSection]:
```
harry-rhesis: Dead code: `list_categories` has no callers anywhere in the repo. I grepped
backend, SDK and frontend.

It is also a bit of a trap for whoever finds it next, since it always returns `content=""` and the
docstring has to explain that. Worth deleting rather than documenting.

apps/backend/src/rhesis/backend/app/services/owasp.py
```diff
@@ -0,0 +1,197 @@
...
+def list_category_summaries(framework: str) -> List[dict]:
+    """Return ``[{id, name, description}, ...]`` for the category picker."""
+    if framework not in OWASP_FRAMEWORKS:
+        valid = sorted(OWASP_FRAMEWORKS)
+        raise ValueError(f"Unknown OWASP framework {framework!r}. Valid: {valid}")
+
+    _cache.initialize()
+    cached = _cache.get_sections(framework)
+    # v2 cache entries include description; ignore legacy id/name-only rows.
+    if cached is not None and cached and "description" in cached[0]:
```
harry-rhesis: Small one: sniffing the cached payload's shape to detect legacy entries works, but
versioning the key does the same job without the conditional. Something like
`owasp:sections:v2:{framework}` in `get_sections`/`set_sections`, then this line goes back to a
plain `if cached:` and legacy rows just age out on their own TTL.

harry-rhesis (commented):
Thanks Alexey, this is a lot of careful work and it shows. A few things I want to call out
specifically before the findings:

- The three migrations are idempotent, and ordered correctly in both directions (behaviors before
  metrics on the way up, metrics before behaviors on the way down). The docstrings explain *why*
  rather than restating what the code does, which made them quick to verify.
- Sourcing the OWASP names from `initial_data.json` by prefix instead of duplicating 22 string
  literals into the migration. That is the right call and it kept the migration short.
- The two layer cache design is sound: permanent parsed content in the object store, id/name/
  description in Redis in front of it, and the SDK stays cache backend agnostic via the
  loader/writer callbacks rather than importing anything backend specific.
- You caught the duplicate test set bug and the metrics filter pill bug yourself, and both commit
  messages explain the mechanism precisely. That is what made the rest of this reviewable.

I looked at the backend and SDK Python only, not the frontend. Everything below is inline on the
specific lines.

**Correctness, worth fixing before merge**

- `app/main.py`: the OWASP pre-warm has no test suite guard, unlike the Garak one right above it.
- `tasks/test_set.py`: OWASP generation is not metered. Every other generation path accrues.
- `tasks/test_set.py`: the retry after commit duplication is still reachable. The fix in `f7fe037`
  addressed the trigger, not the class of bug.
- `app/services/owasp.py`: `except OSError` is too narrow for the fsspec cloud backends.
- `sdk/.../owasp_synthesizer.py`: the multi-turn path loses retry and batch size reduction.
- `sdk/.../owasp_extractor.py`: a cached empty list skips the no-sections guard.

**DRY, worth doing while the code is fresh**

- `sdk/.../owasp_synthesizer.py`: the multi-turn schema and repacker duplicate
  `synthesizers/multi_turn/base.py` field for field. Fixing this also fixes the retry loss above.
- `app/services/organization.py`: third hand rolled tag get-or-create in the codebase.
  `crud.assign_tag` already does exactly this.
- `alembic/utils/metric_sync.py`: the behavior sync mirrors the metric sync closely enough to share
  the org listing.
- `app/services/owasp.py`: `OWASP_FRAMEWORKS` holds the same string under two keys.

**Small stuff, all optional**

Dead `list_categories`, dead `behavior_definitions` parameter, a duplicated done callback, redundant
pdfminer child loggers, cache shape sniffing that a versioned key would remove,
`get_owasp_content_path` not using `self`, and `initial_data.json` getting read four times per
migration run. Take or leave any of these.

**One correction for the PR description**

It says `cd sdk && make test` covers `OWASPSynthesizer`, but `tests/sdk/` has no OWASP tests at all.
The new cache hooks in `fetch_owasp_sections`, `_distribute`, and the flat to nested repack are pure
functions, so they are cheap to cover if you want them before this lands.

Nice piece of work overall. Happy to pair on any of these, especially the SDK consolidation.

sdk/src/rhesis/sdk/synthesizers/owasp_synthesizer.py
```diff
@@ -23,9 +28,29 @@
...
 )
 from rhesis.sdk.synthesizers.base import TestSetSynthesizer
+from rhesis.sdk.synthesizers.utils import load_prompt_template
 
 logger = logging.getLogger(__name__)
 
+
+# Flat schema for multi-turn generation, repacked to the nested test_configuration
+# shape after generation — mirrors multi_turn.base.FlatTest.
+class FlatMultiTurnTest(BaseModel):
```
harry-rhesis: **DRY: this is the largest reuse opportunity in the PR.** Four separate pieces here
already exist in `synthesizers/multi_turn/base.py`:

- `FlatMultiTurnTest` / `FlatMultiTurnTests` are field for field identical to `FlatTest` /
  `FlatTests` (`multi_turn/base.py:38-51`), same `ge=1, le=50` bounds included.
- `_flat_multiturn_test_to_nested` (line 232) duplicates `_flat_test_to_nested`
  (`multi_turn/base.py:80-99`). The only real difference is that the base tolerates missing turn
  counts.
- `generate()` (line 173) duplicates the tail of `MultiTurnSynthesizer.generate`
  (`multi_turn/base.py:167-170`).
- `_generate_multiturn_tests` (line 248) reimplements the batch loop that `_generate_with_retry`
  already owns (`synthesizers/base.py:329`).

A shape that removes all four: import `FlatTests`, lift `_flat_test_to_nested` into
`synthesizers/utils.py` where both can reach it, then override `_generate_batch(num_tests, **kwargs)`
here. That is the hook `_generate_with_retry` already calls, so switching template, schema and
repacker off `self.test_type` inside it is enough:

```python
def _generate_batch(self, num_tests: int, **kwargs: Any) -> List[Dict[str, Any]]:
    if self.test_type != TestType.MULTI_TURN:
        return super()._generate_batch(num_tests, **kwargs)
    prompt = self._get_multi_turn_template().render(num_tests=num_tests, **kwargs)
    response = self.model.generate(prompt=prompt, schema=FlatTests)
    return self._process_batch_response(response, num_tests, batch_index=0)
```

`_generate_without_sources` then always calls `self._generate_with_retry(n, **context)`, and
`_generate_multiturn_tests`, `_get_multi_turn_template`, `_flat_multiturn_test_to_nested` and both
local `Flat*` classes all go away. Roughly 90 lines down to about 20, and it fixes the retry problem
I flagged separately at line 248 for free.

The one thing to check while doing it: `_process_batch_response` hardcodes
`TestType.SINGLE_TURN.value`, so it needs the test type threaded through or read off `self`.

sdk/src/rhesis/sdk/synthesizers/owasp_synthesizer.py
```diff
@@ -167,6 +220,73 @@ def _generate_without_sources(self, num_tests: int = 10, **kwargs: Any) -> List[
...
+                "scenario": flat["test_configuration_scenario"],
+                "min_turns": int(flat["test_configuration_min_turns"]),
+                "max_turns": int(flat["test_configuration_max_turns"]),
+            },
+            "behavior": flat["behavior"],
+            "category": flat["category"],
+            "topic": flat["topic"],
+        }
+
+    def _generate_multiturn_tests(
```
harry-rhesis: **Correctness: the multi-turn path silently drops a whole section's tests on one
flaky LLM call.**

Single-turn generation goes through `_generate_with_retry` (`synthesizers/base.py:329`), which
reduces the batch size on failure, allows 3 consecutive failures before giving up, and logs per
batch. This loop `break`s on the first exception instead, so one transient error means that section
contributes nothing and the user just gets a smaller test set than they asked for, with the reason
buried in worker logs.

Second, smaller thing: these tests never get `metadata.generated_by`, which
`_process_batch_response` (`base.py:260-269`) sets for every single-turn test. So multi-turn OWASP
tests are missing provenance that every other synthesizer path stamps.

Routing multi-turn through `_generate_batch` as suggested at line 38 fixes both, since
`_generate_with_retry` and `_process_batch_response` then apply unchanged.

sdk/src/rhesis/sdk/synthesizers/owasp_synthesizer.py
```diff
@@ -130,6 +170,16 @@ def report_url(self) -> str:
     def _get_synthesizer_name(self) -> str:
         return "OWASPSynthesizer"
 
+    def generate(self, num_tests: int = 10, **kwargs: Any) -> TestSet:
```
harry-rhesis, line 158: Part of the same consolidation as line 38: these four lines are
`MultiTurnSynthesizer.generate`'s tail (`multi_turn/base.py:167-170`). Worth a shared helper, for
example `stamp_multi_turn(test_set)` in `synthesizers/utils.py`, so the naming convention
`"... (Multi-Turn)"` lives in one place. If the suffix ever changes, right now two files have to
change together.

sdk/src/rhesis/sdk/services/owasp_extractor.py
```diff
@@ -378,16 +392,39 @@ def fetch_owasp_sections(
...
-        raise ValueError(
-            f"No top-10 sections found in the PDF at {url}. "
-            "The document may not follow the expected OWASP section-per-page layout."
-        )
+    key = cache_key or hashlib.sha256(url.encode("utf-8")).hexdigest()
+
+    raw_sections: Optional[list[ReportSection]] = None
+    if cache_loader is not None:
+        cached = cache_loader(key)
+        if cached is not None:
```
harry-rhesis: **Correctness: a cached empty list bypasses the no-sections guard.**

If `cache_loader` returns `[]`, this sets `raw_sections = []`, which makes the
`if raw_sections is None` check at line 410 false. So the download, the parse, and the
`ValueError("No top-10 sections found...")` are all skipped, and the caller silently gets zero
sections instead of either real data or a clear error.

`save_owasp_content_cache` cannot write `[]` today because the guard runs first, so this needs a
truncated or corrupt cache entry to trigger. Still worth closing, since it is a one word change and
the failure is silent:

```suggestion
        if cached:
```

apps/backend/src/rhesis/backend/app/services/organization.py
```diff
@@ -275,6 +285,78 @@ def unenroll_all_project_members(
     return len(user_ids)
 
 
+def _get_or_create_owasp_tag(db: Session, organization_id: str, user_id: str) -> models.Tag:
```
harry-rhesis: **DRY: this is the third hand rolled "get or create Tag, then get or create
TaggedItem" block in the backend.** The other two:

- `app/services/garak/importer.py:443-475`, `_tag_garak_behaviors`, which your docstring correctly
  points at as the idiom being mirrored.
- Migration `b857edcac3c0:545-573`, in this same PR.

`crud.assign_tag` (`crud/__init__.py:1261`) already does all of it: get or create the tag by
`(name, organization_id)`, check the `TaggedItem` unique constraint keys, insert, flush. It also
verifies the entity exists with org filtering, which these copies do not.

Either call it directly, or extract one helper and route `organization.py` and Garak's importer
through it. Per `apps/backend/AGENTS.md` a new helper belongs in a per entity crud module rather
than growing `crud/__init__.py`.

Two smaller things while you are in here: the `"Behavior"` and `"Metric"` literals at lines 432 and
922 should be `EntityType` members, which the codebase already has and `assign_tag` already takes.
And `_get_or_create_owasp_tag` is called unconditionally at line 417, so an org whose
`initial_data.json` had no OWASP entries would still get an empty `OWASP` tag. Harmless today since
the file always has them, just slightly surprising.

apps/backend/src/rhesis/backend/alembic/utils/metric_sync.py
```diff
@@ -342,3 +356,207 @@ def remove_metrics_from_organizations(
         if verbose:
             print(f"\n❌ Metric removal failed: {e}\n")
         raise
+
+
+def sync_behaviors_to_organizations(
+    session: Session,
+    behavior_names: List[str] | None = None,
+    behavior_definitions: List[Dict[str, Any]] | None = None,
```
harry-rhesis: Dead parameter: no caller passes `behavior_definitions`. The only caller is
migration `38ed899b9f41`, which passes `behavior_names`.

It costs a branch at line 403, two docstring paragraphs, and an "only applies when
behavior_definitions is not provided" caveat on the parameter above it. Worth dropping until
something actually needs it.

apps/backend/src/rhesis/backend/alembic/utils/metric_sync.py
```diff
@@ -342,3 +356,207 @@ def remove_metrics_from_organizations(
...
+                all_behaviors = [b for b in all_behaviors if b["name"] in behavior_names_set]
+                if verbose:
+                    print(f"   Filtered to {len(all_behaviors)} behaviors by name")
+
+        if verbose:
+            print(f"   Found {len(all_behaviors)} behavior definitions")
+
+        # List orgs with raw SQL so migrations run before newer Organization columns
+        # (e.g. sso_config, slug) exist do not fail with UndefinedColumn on ORM loads.
+        org_rows = session.execute(
```
harry-rhesis: DRY: this org listing is now written three times. Here, in
`sync_metrics_to_organizations`, and again in migration `b857edcac3c0` in this PR. All three repeat
the same details: the raw SQL, `deleted_at IS NULL`, the `owner_id or user_id` fallback, the string
conversion, and the skip when neither is set.

A small generator would carry all of it, including the comment about migrations running before newer
`organization` columns exist, which is a genuinely useful thing to explain once instead of three
times:

```python
def _iter_organizations(session: Session, verbose: bool = True):
    """Yield (org_id, user_id) for every live org, via raw SQL so migrations that
    predate newer Organization columns do not fail on ORM loads."""
```

More broadly, `sync_behaviors_to_organizations` and `sync_metrics_to_organizations` also share the
four key stats dict, the verbose print scaffolding, and the try/except print reraise wrapper. I
would not push for a full merge of the two, the entity specific parts differ enough, but the org
listing is worth pulling out.

apps/backend/src/rhesis/backend/alembic/utils/metric_sync.py
```diff
@@ -26,27 +26,41 @@ def upgrade() -> None:
...
 from rhesis.backend.app.constants import EntityType
 from rhesis.backend.app.utils.crud_utils import (
+    get_or_create_behavior,
     get_or_create_entity,
     get_or_create_status,
     get_or_create_type_lookup,
 )
 
 
+def _load_initial_data() -> Dict[str, Any]:
```
harry-rhesis, line 38: Small one: worth an `@functools.lru_cache` here. Migration `38ed899b9f41` calls
`_owasp_behavior_names()` and `_owasp_metric_names()`, each of which calls this, so
`initial_data.json` gets read and parsed four times per upgrade, and four more per downgrade. The
file is not small.

apps/backend/src/rhesis/backend/logging/logging_config.py
```diff
@@ -290,3 +290,17 @@ def set_logger(worker_role: str | None = None):
         "kombu.pidbox",
     ):
         logging.getLogger(name).setLevel(logging.WARNING)
+
+    # pdfminer emits per-token DEBUG lines while parsing OWASP report PDFs; under
+    # uvicorn --log-level debug that can fill hundreds of MB and stall the
+    # categories endpoint / cache warm-up for minutes.
+    for name in (
+        "pdfminer",
```
harry-rhesis: Small one: the six child entries are redundant. A logger with no level set
inherits its parent's effective level, and none of these set one, so
`logging.getLogger("pdfminer").setLevel(logging.WARNING)` on its own silences all of them.

Good catch on the underlying problem though. Hundreds of MB of DEBUG output from a PDF parse is a
nasty thing to debug from the symptom.

apps/backend/src/rhesis/backend/app/services/storage_service.py
```diff
@@ -240,6 +240,10 @@ def get_attachment_thumbnail_path(
         prefix = self.get_attachment_prefix(organization_id, entity_type, entity_id, file_id)
         return f"{prefix}/thumb-{size}.webp"
 
+    def get_owasp_content_path(self, framework: str) -> str:
```
harry-rhesis: Small one: this does not touch `self`, so it can be a `@staticmethod`, or the
f-string can just live at the two call sites in `services/owasp.py`.

Worth noting the docstring is doing real work here, since every other path helper on this class is
org scoped and this one deliberately is not. That is correct, the source PDFs are public, but it is
the kind of thing that gets "fixed" by someone who does not read the docstring. A brief note that
the content is not tenant specific would make the intent harder to miss.

harry-rhesis (approved):
Thanks Alexey, every item from the review has been addressed. The three post-review commits are clean and well-scoped:

- The correctness fixes (pre-warm guard, metering, post-save isolation, broadened cache errors, multi-turn retry, cached empty list) all land with matching test coverage.
- The DRY consolidation (shared FlatTests import, tag_crud.assign_tag, _list_organizations_with_owner, collapsed OWASP_FRAMEWORKS keys) tidied up exactly the spots that needed it.
- The small stuff (dead code removal, lru_cache, versioned cache key, staticmethod, pdfminer parent logger) is all taken care of too.

Nice work on the SDK test suite as well. The flaky-LLM-call retry test and the post-save isolation test are both good regression guards.

This is good to merge.

PR #2212, opened 2026-07-21
Fix backend crash/freeze on large Garak probe import/sync

harry-rhesis:
## Purpose

Importing a large Garak probe (e.g. `continuation.ContinueSlursReclaimedSlursFull`,
6,463 prompts) or several probes at once could crash/freeze the backend.
`POST /garak/import` and `POST /garak/sync/{test_set_id}` ran fully
synchronously inside `async def` FastAPI handlers, blocking the worker's
entire event loop for the whole operation — and re-instantiated Garak probe
classes from scratch on every call instead of reusing the Redis-backed probe
cache. Past gunicorn's 60s worker timeout this SIGKILLs the worker, dropping
every other request routed to it.

## What Changed

- Move `/garak/import` and `/garak/sync/{test_set_id}` to Celery
  (`task_launcher`, `202` + `task_id`), mirroring the existing
  `/garak/generate` pattern.
- Reuse the probe cache (`GarakProbeService.get_probe`/`get_probes_for_module`
  + `preload_probes()` on `GarakImporter`/`GarakSyncService`) instead of
  re-extracting Garak probe classes from scratch on every import/sync/preview.
- `get_cached_probes()` is a GET-only cache read that never triggers
  enumeration on a miss (that path is CPU-heavy); the still-synchronous
  preview endpoints' enumeration-on-miss now runs in a thread
  (`anyio.to_thread.run_sync`) so a cold cache can't block the loop either.
- Filter the probe cache down to the exact probe class(es) a request needs
  before serializing through the Celery dispatch — not whole modules or the
  full ~168-probe corpus.
- `resolve_sync_target()` mirrors `sync_test_set`'s exact branching (both
  `garak_module` and `garak_probe_class` required for the modern
  single-probe path) so a background sync is never starved of the probes it
  actually needs.
- Fix an N+1 status lookup and a wasted duplicate-content SELECT in
  `create_prompt()` for bulk Garak imports (`cache=`, `skip_duplicate_check=`).
- Frontend: `importProbes()`/`syncTestSet()` return the new task-response
  shape; `GarakImportDrawer` treats import/sync as fire-and-forget (removes
  the fake progress-bar simulation); `onSuccess(testSetIds)` replaced with
  `onImportStarted()` since no test set IDs are available synchronously
  anymore.
- Remove the now-dead `GarakImportResponse`/`GarakImportedTestSet`/
  `GarakSyncResponse` schemas (superseded by the `202` task-response schemas).

## Additional Context

- Investigated and planned in collaboration with @harry-cruz; scope was
  agreed as backend-only frontend adaptation (no new polling UI).
- Known merge conflict: `GarakImportDrawer.tsx` was also touched by #2183
  (BFF session-refresh — removed `sessionToken` threading). Will need a
  rebase/resolve before merge; flagging here rather than resolving
  unprompted since it touches unrelated auth plumbing.

## Testing

- `cd apps/backend && uv run pytest ../../tests/backend/services/garak/ ../../tests/backend/routes/ ../../tests/backend/tasks/ ../../tests/backend/schemas/test_garak.py -v`
  — 237/237 pass.
- Full `services/routes/tasks/schemas` suite: 3784 passed (9 unrelated
  `test_endpoint.py` failures confirmed as resource-contention noise from the
  large combined run — pass 49/49 in isolation).
- `ruff check`/`ruff format --check` clean on all changed backend files.
- `npx tsc --noEmit`, `npm run lint`, and the `garak-client` jest suite clean
  on the frontend.
- Manual verification recommended before merge: import a known-large probe
  against a production-like build (gunicorn timeout active) and confirm the
  response returns `202` immediately while `GET /health` stays responsive on
  the same worker.


harry-rhesis:
Fixed in cab963e — thanks for catching this.

**Critical (legacy sync wipe):** `import_probes`/`sync_test_set` now abandon the whole preload (fall back to `probes_by_module=None`, i.e. live extraction in the task) if *any* requested probe/module is missing from a warm cache, rather than building a partial dict. Also added a defense-in-depth guard in `_sync_legacy_test_set`: it now raises instead of proceeding when every listed module resolves to zero probes.

**Improvement (import silent skip):** same fix covers this — the import path now falls back to `None` (not a partial dict) the moment any requested `(module, class)` pair is missing from the cache, so the importer's per-probe silent-skip path in `_get_probe_info` is no longer reachable via a partial preload.

Added regression tests for both (router-level "missing module/probe → None" for import and sync, plus a direct unit test on `_sync_legacy_test_set`'s new guard).

PR #2228, opened 2026-07-23
Fix T&C gate prompting onboarded users for baseline version

nicolai-rhesis:
## Purpose

Existing onboarded users without a `user_settings.terms` record (accounts created before server-side T&C tracking) were being blocked by the post-login Terms modal even though T&Cs are still at baseline version `1.0` and they already accepted at signup.

## What Changed

- Grandfather onboarded users (`organization_id` set) with no terms record as having accepted `BASELINE_TERMS_VERSION` (`1.0`)
- Keep requiring an explicit accept for pre-onboarding users and after `CURRENT_TERMS_VERSION` is bumped past baseline
- Allow `record_terms_acceptance` to persist a real record for grandfathered users when accept is called
- Add unit and route tests for grandfathering, version bump, and pre-onboarding behavior

## Additional Context

- Hotfix for release PR https://github.com/rhesis-ai/rhesis/pull/2227
- Related to T&C tracking introduced in #2144

## Testing

- `uv run pytest ../../tests/backend/auth/test_terms.py -v`
- `uv run pytest ../../tests/backend/routes/test_auth.py::TestTermsAcceptance -v`
- On staging: log in with an existing onboarded account that has no `user_settings.terms` — modal should not appear while version is still `1.0`


apps/backend/src/rhesis/backend/app/auth/terms.py
```diff
@@ -17,27 +23,49 @@
...
     return (user.user_settings or {}).get("terms") or {}
 
 
+def _is_grandfathered_baseline_acceptance(user: User) -> bool:
+    """Onboarded users with no terms row accepted baseline T&Cs at signup."""
+    if CURRENT_TERMS_VERSION != BASELINE_TERMS_VERSION:
+        return False
+    if _user_terms(user).get("accepted_at"):
+        return False
+    return bool(user.organization_id)
```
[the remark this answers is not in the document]
nicolai-rhesis: Good catch — addressed in 9c0db0409.

Grandfathering is now scoped to **pre-tracking** onboarded accounts only: `organization_id` set **and** `created_at < TERMS_TRACKING_STARTED_AT` (2026-07-14, when #2144 shipped). Post-tracking users with an org but no `user_settings.terms` (admin-created, imports, invitees, etc.) still get `terms_accepted=false` until they accept explicitly.

tests/backend/routes/test_auth.py
```diff
@@ -1519,14 +1519,37 @@ def test_terms_status_returns_false_for_outdated_version_with_prior_acceptance(
...
+    def test_terms_status_returns_false_for_pre_onboarding_user(
+        self, client: TestClient, test_db, test_org_id
+    ):
+        """Users still in onboarding (no organization) must accept explicitly."""
         from rhesis.backend.app.auth.token_utils import create_session_token
 
         email = _unique_email("terms-new")
-        org = create_test_organization(test_db, "Terms New Org")
-        user = create_test_user(test_db, org.id, email, "Terms New User")
+        user = create_test_user(test_db, None, email, "Terms New User")
```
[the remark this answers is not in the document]
nicolai-rhesis: Fixed in 9c0db0409 — `create_test_user` now takes `organization_id: Optional[UUID] = None` and documents that `None` means a pre-onboarding user.

apps/backend/src/rhesis/backend/app/auth/terms.py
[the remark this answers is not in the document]
nicolai-rhesis: Agreed — merge time was the wrong anchor (prod is still on v0.10.0; staging only got the gate with this release).

Switched approach in 87c9a3833: dropped the `created_at` grandfathering entirely and added a one-shot Alembic backfill (`a1b2c3d4e5f9`) that stamps baseline `1.0` acceptance for onboarded users missing a terms record. That runs at deploy per environment, so the effective “tracking started” moment is when the migration applies — no hardcoded rollout date to keep in sync.

apps/backend/src/rhesis/backend/app/auth/terms.py
```diff
@@ -17,27 +23,75 @@
...
+# Version that shipped with server-side tracking. Pre-tracking onboarded users
+# without a persisted record are treated as having accepted this version.
+BASELINE_TERMS_VERSION = "1.0"
+
+# Server-side terms tracking shipped in #2144 (merged 2026-07-14). Accounts
+# created before this instant may lack ``user_settings.terms`` even though they
+# accepted at signup. Accounts created on/after this instant must have an
+# explicit record — organization membership alone is not enough (admin-created
+# / imported users, invitees, etc.).
+TERMS_TRACKING_STARTED_AT = datetime(2026, 7, 14, tzinfo=timezone.utc)
```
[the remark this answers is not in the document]
nicolai-rhesis: Agreed — merge time was the wrong anchor (prod is still on v0.10.0; staging only got the gate with this release).

Switched approach in 87c9a3833: dropped the `created_at` grandfathering entirely and added a one-shot Alembic backfill (`a1b2c3d4e5f9`) that stamps baseline `1.0` acceptance for onboarded users missing a terms record. That runs at deploy per environment, so the effective “tracking started” moment is when the migration applies — no hardcoded rollout date to keep in sync.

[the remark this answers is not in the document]
nicolai-rhesis: Follow-up: the backfill revision id is `b5c6d7e8f9a0` (the first push reused `a1b2c3d4e5f9`, which already existed for the cancelled-status migration — fixed in dfc9c2342).

apps/backend/src/rhesis/backend/alembic/versions/b5c6d7e8f9a0_backfill_baseline_terms_acceptance.py
```diff
@@ -0,0 +1,69 @@
...
+Revises: d4e7f2a9c1b8
+Create Date: 2026-07-23
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "b5c6d7e8f9a0"
```
harry-rhesis, line 25: Nit: this file is still named with the old revision id (`a1b2c3d4e5f9_...`) from before the collision fix in dfc9c2342, but the revision here is now `b5c6d7e8f9a0`. Every other migration in `versions/` has its filename prefix match its revision id. Alembic resolves migrations by file content, not filename, so this isn't functionally broken, but renaming the file would avoid confusion for anyone searching by revision id later.

apps/backend/src/rhesis/backend/alembic/versions/a1b2c3d4e5f9_backfill_baseline_terms_acceptance.py
```diff
@@ -0,0 +1,69 @@
...
+                    COALESCE(user_settings, '{}'::jsonb),
+                    '{terms}',
+                    jsonb_build_object(
+                        'version', :version,
+                        'accepted_at', COALESCE(joined_at, created_at, now())
+                    ),
+                    true
+                )
+            WHERE organization_id IS NOT NULL
+              AND (deleted_at IS NULL OR deleted_at > now())
```
harry-rhesis: Nit: `deleted_at` is always stamped with `datetime.now(timezone.utc)` at soft-delete time (see `models/base.py`), never a future value, so `deleted_at > now()` can never actually be true here. This mirrors a pattern already present in other backfill migrations in the repo, so not something introduced here, but it is effectively dead logic. Could simplify to just `deleted_at IS NULL`.

apps/backend/src/rhesis/backend/alembic/versions/a1b2c3d4e5f9_backfill_baseline_terms_acceptance.py
```diff
@@ -0,0 +1,69 @@
...
+              AND (
+                    user_settings -> 'terms' IS NULL
+                 OR NOT (user_settings ? 'terms')
```
harry-rhesis: Nit: `NOT (user_settings ? 'terms')` is redundant. Postgres's `->` operator returns NULL for a missing key, so the `user_settings -> 'terms' IS NULL` check above already covers this case.

apps/backend/src/rhesis/backend/alembic/versions/b5c6d7e8f9a0_backfill_baseline_terms_acceptance.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+def downgrade() -> None:
+    # Best-effort: cannot distinguish backfill from an explicit 1.0 accept.
+    op.execute(
+        sa.text(
+            """
+            UPDATE "user"
+            SET user_settings = user_settings #- '{terms}'
+            WHERE user_settings -> 'terms' ->> 'version' = :version
```
harry-rhesis, line 65: FYI, not a blocker since the comment above already discloses it: this downgrade removes `terms` from any user whose stored version is baseline `1.0`, including someone who explicitly accepted while `1.0` was current, not only the ones this migration backfilled. Worth remembering if this migration is ever rolled back against a live environment.

apps/backend/src/rhesis/backend/app/auth/terms.py
```diff
@@ -37,7 +41,7 @@ def record_terms_acceptance(user: User) -> None:
     """Persist acceptance of the current T&C version (no-op if already current)."""
     if user_has_accepted_current_terms(user):
         return
-    settings = user.user_settings or {}
+    settings = dict(user.user_settings or {})
```
harry-rhesis, line 44: Nice catch making this a copy instead of mutating `user.user_settings` in place. Reassigning the same dict object back onto a JSONB attribute can silently fail to mark it dirty unless `flag_modified()` is called downstream, a known SQLAlchemy gotcha. Every current caller happens to be safe already (the router calls `flag_modified` explicitly, and `local_init.py` starts from `None`), but this protects any future caller that forgets it.

harry-rhesis (commented):
Thanks for getting this hotfix out so quickly, Nicolai. Reviewed for correctness, conciseness, and architecture.

The pivot from runtime-computed grandfathering (hardcoded cutoff timestamp plus per-request `created_at` checks) to a one-time backfill migration is the right call. It ties the "tracking started" moment to the actual per-environment deploy time instead of a guessed constant, keeps `terms.py` back down to its simple original shape, and leaves a real acceptance record instead of a virtual one. It also matches the pattern already used by the other backfill migrations in this repo, so it fits established convention rather than introducing a new one.

Verification done:
- Ran `pytest ../../tests/backend/auth/test_terms.py ../../tests/backend/routes/test_auth.py::TestTermsAcceptance -v`, 10/10 pass.
- `ruff check` on all four changed files, clean.
- Confirmed the new migration resolves to a single Alembic head, no branching.
- Confirmed `create_test_user`'s new optional `organization_id` default stays backward compatible with all existing call sites.

A few small, non-blocking items called out inline:
- The migration file's name still carries the old revision id from before the collision fix, worth a rename for consistency.
- A couple of minor redundancies in the backfill WHERE clause.
- The `dict()` copy in `record_terms_acceptance` is a nice defensive catch, flagged inline for context.

Nice work stacking this cleanly on top of #2227.

apps/backend/src/rhesis/backend/alembic/versions/b5c6d7e8f9a0_backfill_baseline_terms_acceptance.py
nicolai-rhesis: Done in 3f0e0be64 / d3c7c18cf — file is now `b5c6d7e8f9a0_backfill_baseline_terms_acceptance.py`.

apps/backend/src/rhesis/backend/alembic/versions/a1b2c3d4e5f9_backfill_baseline_terms_acceptance.py
nicolai-rhesis: Simplified to `deleted_at IS NULL` in 3f0e0be64.

apps/backend/src/rhesis/backend/alembic/versions/a1b2c3d4e5f9_backfill_baseline_terms_acceptance.py
nicolai-rhesis: Dropped the redundant `? 'terms'` check in 3f0e0be64.

PR #2242, opened 2026-07-23
Fix: Sso Token Not Fount Annotation

asad-miah:
This PR introduces changes from the `fix/sso-token-not-fount-annotation` branch.

```
apps/frontend/src/app/(protected)/annotations/components/AnnotationsGrid.tsx
apps/frontend/src/app/(protected)/annotations/page.tsx
```

## 📋 Commit Details

```
f5a418755 - fix(frontend): stop gating annotations page on session_token (Md Asaduzzaman Miah, 2026-07-23 15:18)
```

## ✅ Checklist

- [ ] All tests pass locally

### DIFF
--- a/apps/frontend/src/app/(protected)/annotations/components/AnnotationsGrid.tsx
+++ b/apps/frontend/src/app/(protected)/annotations/components/AnnotationsGrid.tsx
@@ -39,7 +39,6 @@ import AnnotationFilterDrawer, {
 } from './AnnotationFilterDrawer';
 
 interface AnnotationsGridProps {
-  sessionToken: string;
   onTotalCountChange?: (count: number) => void;
 }
 
@@ -116,7 +115,6 @@ function formatTarget(item: AnnotationListItem): string {
 }
 
 export default function AnnotationsGrid({
-  sessionToken,
   onTotalCountChange,
 }: AnnotationsGridProps) {
   const theme = useTheme();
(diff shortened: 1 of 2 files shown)


nicolai-rhesis (commented):
## Review

Solid, correct fix. This does exactly what the frontend BFF Auth Pattern in `apps/frontend/AGENTS.md` prescribes, and all CI checks (lint, test, 3 E2E shards, security) pass.

### What's right
- Removes the dead `sessionToken` prop-drilling: `new ApiClientFactory(sessionToken)` was silently dropping the token client-side anyway (`buildAuthHeaders()` only attaches `Authorization` when `typeof window === 'undefined'`), so the no-arg `new ApiClientFactory()` + same-origin `/api/backend` proxy is the correct path.
- Gating on `isSessionLoading(status)` / `isAuthenticated(status)` instead of `session_token` presence matches the documented pattern.
- Dropping `enabled: !!sessionToken` is safe since the grid only mounts after `page.tsx` confirms `isAuthenticated(status)`.
- Cleanup is complete — no residual `sessionToken` references remain in `annotations/`.

### One nit (worth fixing before merge)
The unauthenticated fallback copy still reads `"No session token available"`, but the gate is no longer about a token — it's `!isAuthenticated(status)`. Please update it to reflect the real state, e.g. "You are not signed in" / "Session expired".

### Non-blocking
- The `!isAuthenticated(status)` branch is likely unreachable in practice (an unauthenticated user fails the `canRead` check first and hits `AccessDenied`), so that fallback is close to vestigial — pre-existing, not introduced here.
- No tests needed (no existing annotations tests). A short "Testing" note in the PR description would help reviewers.

PR #2245, opened 2026-07-23
Fix mapping tab staleness, missing created by/on, and silent test errors

EmanueleDeRossi1:
## Purpose

Fixes three bugs on the endpoint detail page reported by a user testing staging.

## What Changed

- Sync the saved endpoint into the React Query cache after a mapping edit, so switching tabs and back no longer reverts to stale cached data (previously only local state was updated, not the cache).
- Expose `created_at`/`updated_at` on the backend `Endpoint` schema and wire up `endpoint.user`/`endpoint.created_at` on the frontend so "Created by"/"Created on" render instead of a hardcoded placeholder / wrong metadata field.
- Surface real error details on the Test tab's "Test connection" action: the server action can resolve with `{ success: false, error }` without throwing, and the backend can also return HTTP 200 with an in-band `error: true` payload (e.g. on a connection failure) — neither case was previously checked, so failures silently showed a `200` with no message.

## Additional Context

- Backend intentionally returns 200 for some invocation failures (`rest_invoker.py`) to preserve full request/response context for tracing; the frontend now checks the payload's `error` field explicitly instead of inferring success from HTTP status.

## Testing

- `npx tsc --noEmit` and `npx eslint` clean on all changed frontend files.
- `uvx ruff check` / `uvx ruff format` clean on the changed backend file.
- Not manually verified against a running local stack (requires full backend + DB + auth); please exercise the mapping tab, endpoint overview metadata, and Test tab error path on staging before merging.

### DIFF
--- a/apps/backend/src/rhesis/backend/app/schemas/endpoint.py
+++ b/apps/backend/src/rhesis/backend/app/schemas/endpoint.py
@@ -1,3 +1,4 @@
+from datetime import datetime
 from typing import Any, Dict, List, Literal, Optional
 
 from pydantic import UUID4, BaseModel, ConfigDict, field_validator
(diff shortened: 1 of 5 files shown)


EmanueleDeRossi1:
Thanks to @harry-rhesis for spotting these bugs!

PR #2259, opened 2026-07-24
Add Summarize FAB on test run detail page

nicolai-rhesis:
## Purpose

Harry Cruz asked to mirror the Insights “Summarize insights” Architect FAB on the test run detail view so users can open a pre-seeded Architect session from a single run.

## What Changed

- Added `TestRunSummarizeFab` that creates an Architect session and opens `/architect?session=<id>` with a handoff prompt for the current run
- Added prompt/title builders for the test-run handoff
- Wired the FAB into the test run detail header `FabGroup` (first in the group)
- Added unit tests for the prompt builders

## Additional Context

- Reuses the shared `createAndOpenArchitectSession` handoff and `Capability.Architect.CREATE` gate
- Scope is the whole test run (header is tab-agnostic); Test Cases-tab client filters are not included in the prompt

## Testing

- [ ] Open a test run detail page with `architect:create`
- [ ] Confirm the Summarize FAB appears in the header
- [ ] Click it and verify a new Architect tab opens with a session summarizing that run
- [ ] Confirm users without `architect:create` do not see the FAB


nicolai-rhesis:
Addressed the double-click race: `creatingRef` now gates `handleClick` before `setCreating`, so a second click before re-render is ignored.

PR #2269, opened 2026-07-26
fix(backend): support relationship field sorting

### LINKED ISSUE #2254: Fix: sorting by Behavior column on Tests page returns API error
## Summary
Sorting the Tests grid by the "Behavior" column fails with a 400 API error and the grid shows no rows. Previously flagged by Emanuele De Rossi.

## Steps to Reproduce
1. Go to the Tests page.
2. Click the "Behavior" column header to sort by it.
3. Observe: an error banner appears and the grid shows "No rows".

## Expected Behavior
Sorting by Behavior should order the test list by the linked behavior's name (or a sensible proxy) without erroring.

## Actual Behavior
API error is shown:
```
API error: 400 - Invalid sort field: behavior.name. Must be one of: prompt_id, test_type_id, priority, user_id, assignee_id, owner_id, test_configuration, parent_id, topic_id, behavior_id, category_id, status_id, source_id, test_metadata, id, nano_id, created_at, updated_at, deleted_at, project_id, organization_id
```
The grid ends up empty (0 of 0 rows).

## Possible Cause
The Behavior column's DataGrid field is `behavior.name` (`apps/frontend/src/app/(protected)/tests/components/TestsGrid.tsx`), which is passed straight through to the backend `sort_by` param via `gridSortToApiParams` (`apps/frontend/src/utils/grid-sort.ts`). That mapping table only translates a few fields (`counts.comments`, `counts.tasks`, `tags`) and has no entry for `behavior.name`, so the raw client-side field name reaches the backend, which only understands `behavior_id`.

- [ ] Clicking the Behavior column header to sort no longer produces an API error.
- [ ] Sorting behaves reasonably (either sorts by behavior via a backend-supported approach, or the column is marked non-sortable if server-side sort by behavior name isn't supported).
- [ ] Regression test covering `gridSortToApiParams` / TestsGrid sort behavior for the Behavior column.

## Additional Context
Reported by @edenrossi (Emanuele De Rossi) as already flagged previously.

Missing-Identity:
## Summary

Adds backend support for sorting Tests by relationship-backed fields using correlated scalar subqueries, following the existing virtual count-sort pattern.

## Changes

- support `behavior.name`, `topic.name`, `category.name`, and `test_type.type_value`
- validate relationship sort fields against the requested model
- dispatch supported fields through a shared relationship-sort helper
- add focused coverage for detection, validation, SQL generation, ordering, and QueryBuilder dispatch

Closes #2254.

## Validation

- GitHub Actions: Lint Check, Secret Scanning, Security Scanning (Trivy), and Backend Tests passed
- direct query and SQL matrix passed for all four fields in ascending and descending order
- Ruff format check passed
- Ruff check passed
- `git diff --check` passed

### DIFF
--- a/apps/backend/src/rhesis/backend/app/utils/query_utils.py
+++ b/apps/backend/src/rhesis/backend/app/utils/query_utils.py
@@ -454,9 +454,20 @@ class QueryBuilder:
             apply_virtual_count_sort,
             is_virtual_count_sort,
         )
+        from rhesis.backend.app.utils.relationship_sort import (
+            apply_virtual_relationship_sort,
+            is_virtual_relationship_sort,
+        )
 
         if self._sort_by and is_virtual_count_sort(self._sort_by):
             self.query = apply_virtual_count_sort(
+                self.query,
+                self.model,
+                self._sort_by,
+                self._sort_order,
+            )
+        elif self._sort_by and is_virtual_relationship_sort(self._sort_by):
+            self.query = apply_virtual_relationship_sort(
                 self.query,
                 self.model,
                 self._sort_by,
--- a/apps/backend/src/rhesis/backend/app/utils/query_validation.py
+++ b/apps/backend/src/rhesis/backend/app/utils/query_validation.py
@@ -8,6 +8,10 @@ from rhesis.backend.app.utils.count_sort import (
     VIRTUAL_COUNT_SORT_FIELDS,
     model_supports_count_sort,
 )
+from rhesis.backend.app.utils.relationship_sort import (
+    VIRTUAL_RELATIONSHIP_SORT_FIELDS,
+    model_supports_relationship_sort,
+)
 
 
 def validate_sort_field(model: Type, sort_by: str) -> None:
@@ -17,6 +21,16 @@ def validate_sort_field(model: Type, sort_by: str) -> None:
             raise HTTPException(
                 status_code=400,
                 detail=f"Invalid sort field: {sort_by}. Model does not support this count sort.",
+            )
+        return
+
+    if sort_by in VIRTUAL_RELATIONSHIP_SORT_FIELDS:
+        if not model_supports_relationship_sort(model, sort_by):
+            raise HTTPException(
+                status_code=400,
+                detail=(
+                    f"Invalid sort field: {sort_by}. Model does not support this relationship sort."
+                ),
             )
         return
 
--- a/apps/backend/src/rhesis/backend/app/utils/relationship_sort.py
+++ b/apps/backend/src/rhesis/backend/app/utils/relationship_sort.py
@@ -1,0 +1,54 @@
+"""Virtual sort fields backed by related model columns."""
+
+from __future__ import annotations
+
+from sqlalchemy import desc, select
+from sqlalchemy.orm import Query
+
+from rhesis.backend.app.models.behavior import Behavior
+from rhesis.backend.app.models.category import Category
+from rhesis.backend.app.models.topic import Topic
+from rhesis.backend.app.models.type_lookup import TypeLookup
+
+_RELATIONSHIP_SORT_FIELDS = {
+    "behavior.name": (Behavior, "name", "behavior_id"),
+    "topic.name": (Topic, "name", "topic_id"),
+    "category.name": (Category, "name", "category_id"),
+    "test_type.type_value": (TypeLookup, "type_value", "test_type_id"),
+}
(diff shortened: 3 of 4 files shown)


harry-rhesis:
Left a suggestion on the linked issue (#2254) worth considering before we go with disabling sort here: https://github.com/rhesis-ai/rhesis/issues/2254#issuecomment-5085714836

Short version: the backend already has a "virtual sort field" pattern (`comments_count`, `tasks_count`, `tags_count` in `apps/backend/src/rhesis/backend/app/utils/count_sort.py`) for sorting on fields that are not real columns on the model, using a correlated subquery rather than a join. `behavior.name` fits the same shape, and the field name would not even need to change on the frontend since `sort_by` is passed through as a plain string.

If that approach works out, it would let us drop the `sortable: false` change here and keep the column fully functional. Happy to help review or pair on it if useful. Also flagging, same as on the issue, that this repo requires DCO sign-off on commits.

harry-rhesis:
Thanks for the PR, @Missing-Identity, and for the clean regression test alongside it.

Left a suggestion on the linked issue (#2254) worth considering before we go with disabling sort here: https://github.com/rhesis-ai/rhesis/issues/2254#issuecomment-5085714836

Short version: the backend already has a "virtual sort field" pattern (`comments_count`, `tasks_count`, `tags_count` in `apps/backend/src/rhesis/backend/app/utils/count_sort.py`) for sorting on fields that are not real columns on the model, using a correlated subquery rather than a join. `behavior.name` fits the same shape, and the field name would not even need to change on the frontend since `sort_by` is passed through as a plain string.

If that approach works out, it would let us drop the `sortable: false` change here and keep the column fully functional. Would you be up for giving it a try? Happy to help review or pair on it if useful. Also flagging, same as on the issue, that this repo requires DCO sign-off on commits.

Missing-Identity:
Thanks for the detailed direction, Harry — I’ve reworked the PR around
the correlated-subquery approach.

`behavior.name` is now handled as a virtual relationship sort, the
frontend fallback has been removed, and focused backend regression
coverage is included. This also means previously persisted
`behavior.name` sort models now follow a valid backend path rather
than needing a frontend state migration.

I rebased onto the latest `main` and rewrote the branch as a single
DCO-signed commit. CI is running on the updated branch now.

harry-rhesis:
Thanks for the rework, Unmilan. This is a really clean follow-through on the virtual relationship sort approach, and the DCO-signed single commit looks good.

I validated the `behavior.name` path locally and it looks correct: correlated subquery, validation wired up, QueryBuilder dispatch in place, and the frontend no longer needs to disable the column. Once CI is green on this, we should be in good shape.

One gap before we merge: Topic, Category, and Test Type have the same bug. The Tests grid still sorts those columns by dotted fields that the API does not accept today:

- `topic.name`
- `category.name`
- `test_type.type_value`

Could you extend `relationship_sort` to cover those as well, using the same correlated-subquery pattern (`topic_id` / `category_id` / `test_type_id`)? That should close out the remaining sort failures on the Tests page in one go.

Also worth a look: peqy left a couple of review notes on this PR (silent no-op in `apply_virtual_relationship_sort` if called with an unsupported model, and earlier context around persisted sort models). The silent no-op matches the existing `count_sort` contract and is not a blocker for the normal QueryBuilder path, but if it is easy to tighten while you are in there, that would be welcome.

As soon as topic / category / test type are covered too, we should be cool to merge. Happy to re-review quickly once that lands.

harry-rhesis:
Thanks for sticking with this through the back-and-forth, Unmilan. The rework came out really clean, and the correlated-subquery approach fits right into the existing patterns. All four relationship fields are covered, tests are thorough, and CI is green.

Congrats on your first contribution to the project! Would love to see more from you.

A few small things before we merge:

1. **PR title and description are stale.** They still reference disabling the sort, but the code now enables backend support for it. Could you update the title to something like `fix(backend): support relationship field sorting` (matching the commit message) and revise the description to reflect the correlated-subquery approach?

2. **peqy's silent no-op note.** `apply_virtual_relationship_sort()` silently returns the query unchanged for unsupported models. This matches the `count_sort` contract, so it is not a blocker on the normal `QueryBuilder` path. But if it is easy to add a log or raise for direct callers while you are in there, that would tighten things up.

3. **Minor test gap.** There is no test for a completely unknown dotted field (e.g., `validate_sort_field(Test, "unknown.field")`). The existing validation handles it correctly, but an explicit test would lock that behavior in.

None of these are blockers for the implementation itself. Once the title/description are updated, this is good to go.

PR #2280, opened 2026-07-28
fix(auth): replace permanent Redis init latch with time-based cooldown (closes #2272)

### LINKED ISSUE #2272: Redis client permanently stops retrying after one failed ping, causing silent prod auth outages
## Summary
RedisConnectionManager.initialize() marks Redis as permanently unavailable after the first failed connection attempt and never retries. Any backend pod that fails its first Redis ping (e.g. during a Redis/Valkey restart or CPU-throttling blip) will reject every login for its entire remaining lifetime — even long after Redis is healthy again — until it's manually restarted.

## Steps to Reproduce
1. Backend pod starts while Redis/Valkey is briefly unreachable (restart, throttling, rolling-deploy race).
2. redis_manager.initialize() runs once at startup (main.py:372); the ping fails and _initialization_failed is set True (redis_client.py:28-29).
3. Redis/Valkey recovers seconds later.
4. Any POST /auth/exchange-code served by that pod still raises TokenStoreUnavailableError → 503, since is_available checks the latched flag, not live connectivity.

## Expected Behavior
Once Redis/Valkey recovers, affected pods should reconnect and resume normal operation automatically.

## Actual Behavior
Affected pods return 503 "Authentication temporarily unavailable" indefinitely until manually restarted.

## Logs / Evidence
- Prod logs showed `Redis not available: Error 111 connecting to rhesis-valkey-primary:6379` on specific pods, persisting well after Valkey was confirmed healthy.
- Direct diagnostic probe (`POST /auth/exchange-code` with a dummy code, 15–20x in a row) returned a roughly 50/50 mix of 400/503 across the 6 prod replicas — consistent with ~half the pods being permanently latched.
- A `kubectl rollout restart deployment/backend` immediately resolved it (all pods returned 400 on the same probe afterward), confirming the fix is "get a fresh process," not "Redis recovering."

## Impact
All logins can silently fail on affected replicas after any Redis/Valkey blip, with no self-healing. Root-caused during a 2026-07-27 production incident; required a manual rolling restart as a workaround.

## Possible Cause
RedisConnectionManager.initialize() in redis_client.py uses a one-shot _initialization_failed flag instead of a retryable state.

## Proposed Solution
- Replace the permanent latch with a time-based cooldown (e.g. retry after 30s) so a transient outage self-heals.
- Have used_token_store.py's claim_token_jti / store_auth_code_tokens / consume_auth_code_tokens call await redis_manager.initialize() before checking is_available, instead of relying solely on the one-time startup call — so they opportunistically reconnect.

## Acceptance Criteria
- [ ] A pod whose first Redis ping fails automatically reconnects once Redis is reachable again, without a pod restart.
- [ ] Regression test: first ping fails, Redis recovers, subsequent call succeeds.
- [ ] No permanently "poisoned" state survives past a bounded retry window.

feiiiiii5:
## Root Cause

`RedisConnectionManager.initialize()` set `_initialization_failed = True` on first connection failure and **never retried**. Any pod that missed its first Redis ping (during a restart, CPU throttle, or rolling-deploy race) would reject every login for its entire lifetime — even after Redis recovered — until manually restarted.

Root-caused during a **2026-07-27 production incident** where ~half the backend replicas returned 503 indefinitely after a brief Valkey blip.

## Fix

Replace the permanent `_initialization_failed` bool with a `_last_failed_at` timestamp + cooldown:

```python
# Before (buggy):
if self._initialized or self._initialization_failed:
    return  # permanent latch — never retries

# After (fixed):
if self._initialized:
    return
now = time.monotonic()
if self._last_failed_at and (now - self._last_failed_at) < _RETRY_COOLDOWN_SECONDS:
    return  # throttled — retry after 30s
```

- `_RETRY_COOLDOWN_SECONDS = 30` — prevents hammering Redis during outage
- On success, `_last_failed_at` resets to 0 — full self-healing
- No new dependencies, no API changes

## Test

- [x] 4 regression tests (`tests/backend/test_redis_client_retry.py`)
  - `test_first_failure_does_not_latch` — verifies no permanent latch
  - `test_retry_after_cooldown` — fail → cooldown → succeed
  - `test_no_retry_during_cooldown` — throttle prevents immediate retry
  - `test_is_available_false_until_connected` — property stays correct
- [x] All 4 tests pass
- [x] DCO signed

## Diff scope

2 files changed, +184/-5 lines (16 lines core fix + 168 lines tests)

## AI Disclosure

AI-assisted debug/initial draft/testing. Root cause from issue #2272 analysis (production incident report). Tests verified locally with mocked Redis. Human review of diff scope and correctness.


harry-rhesis:
Thanks for picking this up, @feiiiiii5! The root cause analysis is spot-on and the cooldown mechanism itself is the right approach. Great work on the tests too.

There is one gap that still needs to be addressed before this can be merged, though. It's the same thing peqy flagged: **the cooldown works, but nothing in the runtime ever triggers a retry.**

### The problem

`redis_manager.initialize()` is only called once, at startup in `main.py:372`:

```python
await redis_manager.initialize()  # Logs warning if fails, doesn't raise
```

The three call sites in `used_token_store.py` (lines 91, 132, 161) all check `redis_manager.is_available` before using the client, but they never call `initialize()`. So the actual request flow is:

1. Pod starts, `initialize()` runs, ping fails, `_last_failed_at` is set.
2. Redis recovers 5 seconds later.
3. A login request hits `claim_token_jti()` at line 91. It checks `is_available`, which is `False` (because `_initialized` is still `False`). It raises `TokenStoreUnavailableError`.
4. Nobody ever calls `initialize()` again, so the cooldown logic never runs. The pod is still permanently broken.

The cooldown is correct, it just needs a caller.

### Proposed fix

Have each `used_token_store.py` function call `await redis_manager.initialize()` before checking `is_available`. The cooldown inside `initialize()` already handles throttling, so this is safe to call on every request. The pattern would look like:

```python
async def claim_token_jti(...) -> bool:
    from rhesis.backend.app.services.connector.redis_client import redis_manager

    await redis_manager.initialize()  # no-op if connected; retries after cooldown

    if not redis_manager.is_available:
        raise TokenStoreUnavailableError("Token store temporarily unavailable")
    ...
```

Apply the same change to `store_auth_code_tokens` (line 132) and `consume_auth_code_tokens` (line 161).

This is also what the original issue description proposed under "Proposed Solution":

> Have `used_token_store.py`'s `claim_token_jti` / `store_auth_code_tokens` / `consume_auth_code_tokens` call `await redis_manager.initialize()` before checking `is_available`, instead of relying solely on the one-time startup call -- so they opportunistically reconnect.

### Tests

The four unit tests you have are good for validating the cooldown internals. To cover the end-to-end fix, please add at least one test that exercises the actual recovery path:

1. First `initialize()` call fails (simulating startup during a Redis blip).
2. Cooldown expires.
3. A `used_token_store` function (e.g. `claim_token_jti`) is called. It should trigger `initialize()` internally, reconnect, and succeed instead of raising `TokenStoreUnavailableError`.

This test would have caught the current gap and will prevent regressions if someone later removes the `initialize()` call from the token store functions.

### Housekeeping

Please also make sure all commits are signed off for DCO compliance (`git commit -s` or `git commit --amend -s` on the existing commit).

Thanks again for the solid investigation and clean implementation of the cooldown. Just needs this one more piece to close the loop.

feiiiiii5:
@harry-rhesis Thanks for the thorough review — you're absolutely right that the cooldown needs a caller to be useful. Fixed in `4906135`:

## Changes

1. **`used_token_store.py`**: Added `await redis_manager.initialize()` before each `is_available` check in all three functions:
   - `claim_token_jti` (line 91)
   - `store_auth_code_tokens` (line 133)
   - `consume_auth_code_tokens` (line 165)

2. **`test_used_token_store.py`**: Added `TestRedisRecovery` class with `test_claim_token_jti_triggers_reconnect` that exercises the full recovery path:
   - First `initialize()` fails (simulating startup during Redis blip)
   - `claim_token_jti` calls `initialize()` internally
   - Reconnect succeeds, SET operation completes

3. **DCO**: All commits signed off.

The cooldown + caller pattern now forms a complete self-healing loop. Let me know if anything else is needed.

feiiiiii5:
@peqy Thanks for the question — good catch on the mixed usage.

To clarify: in **redis-py 5.x** (pinned as `redis>=5.0.0,<6.0.0` in this project), there are actually **two different `from_url` APIs** with different signatures:

| API | Type | Awaitable? |
|-----|------|-----------|
| `redis.asyncio.from_url(url, ...)` | Module-level function | ✅ Yes (`async def`, returns a coroutine) |
| `redis.asyncio.Redis.from_url(url, ...)` | Classmethod on `Redis` | ❌ No (synchronous factory, returns a `Redis` instance directly) |

In this PR's code (`redis_client.py`):
```python
import redis.asyncio as redis
...
self._client = await redis.from_url(...)  # module-level async function — await is correct
```

The other call sites you're likely seeing without `await` are using the **classmethod** form:
```python
# redis_pubsub.py, publisher.py, cache.py, etc.
self._redis = AsyncRedis.from_url(...)  # classmethod — no await, also correct
```

Both are valid and do the same thing internally — the module-level `redis.asyncio.from_url()` is actually just a thin async wrapper that calls the `Redis.from_url()` classmethod under the hood. The codebase is consistent: every call site matches the correct calling convention for the API it uses. No `TypeError` risk here.

feiiiiii5:
Thanks @peqy — agreed on the test-patching point. The existing tests in this PR patch `redis.from_url` as `AsyncMock(return_value=mock_client)` (async callable), so `await redis.from_url(...)` yields the mocked client correctly. I'll keep that convention if any new test sites touch the module-level `from_url`. Appreciate the careful review.

harry-rhesis:
One small thing before merging: `tests/backend/test_redis_client_retry.py` sits at the top level of the test directory. The four cooldown tests it contains (first failure doesn't latch, retry after cooldown, no retry during cooldown, is_available correctness) are all testing `RedisConnectionManager` internals, so they belong under `tests/backend/services/connector/` to match the source layout.

That said, do we actually need them as a separate file? The recovery test you added in `tests/backend/auth/test_used_token_store.py` already covers the end-to-end path (startup failure, cooldown, reconnect via token store call). The cooldown unit tests are nice to have, but they're testing `time.monotonic()` arithmetic on private attributes, which makes them brittle if the implementation changes.

Could you either move the file to `tests/backend/services/connector/test_redis_client.py`, or if you think the recovery test in `test_used_token_store.py` already gives enough coverage, drop it entirely? Happy with either option.

harry-rhesis:
Following up on my previous comment about `tests/backend/test_redis_client_retry.py`. I checked the coverage more carefully and the two test files are actually complementary, not redundant:

- `TestRedisRecovery` in `test_used_token_store.py` mocks the entire `redis_manager` and only proves the wiring: that `claim_token_jti` calls `initialize()`. It doesn't exercise the actual cooldown logic at all.
- The four tests in `test_redis_client_retry.py` exercise the real `RedisConnectionManager` class: permanent latch is gone, retry works after cooldown, retry is blocked during cooldown, `is_available` stays correct.

Without the cooldown tests, someone could reintroduce the permanent `_initialization_failed` bool and `TestRedisRecovery` would still pass, since it mocks the manager entirely. So both files are needed.

The file just needs to move to `tests/backend/services/connector/test_redis_client.py` to match the source layout. Everything else looks good.

feiiiiii5:
Both points addressed in commit `3e40ea10b`:

1. **Test placement**: `tests/backend/test_redis_client_retry.py` moved to `tests/backend/services/connector/test_redis_client_retry.py`, mirroring the source layout.
2. **CI (Backend tests)**: the failure was `TypeError: object MagicMock can't be used in 'await' expression` in `TestClaimTokenJti` — `claim_token_jti` now awaits `redis_manager.initialize()`, but the five affected tests patched `redis_manager` with a plain `MagicMock`. They now set `mock_mgr.initialize = AsyncMock()`, matching the convention already used by `test_claim_token_jti_triggers_reconnect`.

Local verification: both test files pass (13/13) against the 3.12 venv built from `apps/backend/uv.lock`.

harry-rhesis:
@feiiiiii5 thanks for going the extra-mile to get this one merged!

PR #2311, opened 2026-07-30
Fix all tenant_isolation RLS policies crashing on missing GUC

harry-rhesis:
## Purpose

SSO login for new users crashes on `project_membership` (and potentially any other table) because the `tenant_isolation` RLS policy uses bare `current_setting('app.current_organization')::uuid`, which fails when no tenant GUC is set. PR #2300 fixed this for `organization_member` only. This PR fixes every remaining table.

## What Changed

- Added migration `d5e6f7a8b9c1` that dynamically finds all tables with a `tenant_isolation` policy and rewrites them using the NULLIF trusted-context pattern:
  ```sql
  NULLIF(current_setting('app.current_organization', true), '') IS NULL
  OR organization_id = NULLIF(current_setting('app.current_organization', true), '')::uuid
  ```
- Skips tables already fixed individually: `organization` (CASE pattern), `organization_member` (#2300), `role` and `role_permission` (custom join-based policies)
- Uses dynamic PL/pgSQL to avoid hardcoding table names. This catches all current and future tables created by the original `fcac5b8b5eb0` migration template.

## Additional Context

- Follows up on #2300 (organization_member fix)
- Production error from netgo deployment: `unrecognized configuration parameter "app.current_organization"` on `project_membership` during SSO auto-provisioning
- The original RLS migration (`fcac5b8b5eb0`) and backfill (`d4e5f6a7b8c3`) both used the bare pattern across all tables

## Testing

- Migration tested locally: upgrade, downgrade, and re-upgrade all succeed
- Verified via `pg_policies`: all 46 tables with `tenant_isolation` now show HARDENED (NULLIF) or CASE-OK status. Zero BARE policies remain.

### DIFF
--- a/apps/backend/src/rhesis/backend/alembic/versions/d5e6f7a8b9c1_harden_all_tenant_isolation_rls.py
+++ b/apps/backend/src/rhesis/backend/alembic/versions/d5e6f7a8b9c1_harden_all_tenant_isolation_rls.py
@@ -1,0 +1,125 @@
+"""Harden all tenant_isolation RLS policies against missing/empty GUC
+
+The original RLS migration (fcac5b8b5eb0) and the backfill migration
+(d4e5f6a7b8c3) created tenant_isolation policies using:
+
+    organization_id = current_setting('app.current_organization')::uuid
+
+This crashes in two ways when no tenant GUC is set (SSO callback,
+migrations, seeding):
+
+1. ``unrecognized configuration parameter`` -- current_setting without
+   missing_ok raises when the GUC has never been SET in the session.
+2. ``invalid input syntax for type uuid: ""`` -- even with missing_ok,
+   an empty string cannot be cast to uuid.
+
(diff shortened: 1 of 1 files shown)


harry-rhesis:
Closing in favor of a targeted fix.

The root cause is not the RLS policies. `organization_id` is resolved correctly (the failing query carries a concrete org UUID). What is missing is the Postgres session variable `app.current_organization` on the SSO callback's connection.

Every other caller of `on_user_org_assigned` sets the tenant GUCs first (`routers/organization.py:186`, `local_init.py:64`, and the contract stated at `local_init.py:213`). The SSO callback uses `get_db_session`, which sets no GUCs, and never added that step.

Relaxing all 46 policies to fail open would hide that bug and weaken tenant isolation everywhere. Replacement PR sets the GUCs in the callback instead and leaves the strict policies as a safety net.

PR #2318, opened 2026-07-30
Feat/Argocd Prd Promotion Gate

asad-miah:
This PR introduces changes from the `feat/argocd-prd-promotion-gate` branch.

## 📁 Files Changed (       6 files)

```
.github/workflows/publish-release.yml
.github/workflows/test-promote-stg-config.yml
kubernetes/clusters/prd/base.yaml
kubernetes/clusters/stg/base.yaml
kubernetes/README.md
RELEASING.md
```

## 📋 Commit Details

```
0b9cb6e15 - fix(dev): refresh bootstrap SHA to release/v0.11.0's current tip (Md Asaduzzaman Miah, 2026-07-30 16:26)
6c215ee32 - chore(dev): drop leftover test workflow from this branch (Md Asaduzzaman Miah, 2026-07-30 16:19)
468e69dac - feat(dev): sync promoted prd revision back into git (Md Asaduzzaman Miah, 2026-07-30 16:17)
ed55485b7 - fix(dev): retry argocd app set/sync on transient lock (Md Asaduzzaman Miah, 2026-07-30 15:24)
c27edd0dd - revert(dev): drop stg targetRevision pin, unneeded for test (Md Asaduzzaman Miah, 2026-07-30 14:28)
f01fddcf6 - test(dev): mirror prd promotion gate on stg for validation (Md Asaduzzaman Miah, 2026-07-30 14:11)
70b9c96d4 - commit hash based deployment for prd (Md Asaduzzaman Miah, 2026-07-30 14:07)
```

## ✅ Checklist

- [ ] All tests pass locally


.github/workflows/publish-release.yml
```diff
@@ -27,9 +27,168 @@ jobs:
...
+          git add "${FILES[@]}"
+          git commit -m "chore(prd): pin prd Application targetRevisions to $GITHUB_SHA"
+
+          # main can advance between the checkout above and this push; retry on
+          # non-fast-forward by rebasing onto the new tip.
+          for attempt in 1 2 3; do
+            git push origin HEAD:main && exit 0
+            echo "Push rejected (attempt $attempt) — rebasing onto origin/main and retrying..."
+            git fetch origin main
+            git rebase origin/main
```
harry-rhesis, line 162: **`git rebase` under `set -e` aborts on merge conflict, leaving prd promoted but git stale.**

Two releases running close together that both modify the same `targetRevision` lines would cause the second run's `git push` to be rejected. The retry path runs `git rebase origin/main`, which conflicts on the identical lines. `set -e` aborts the step immediately. ArgoCD is already live at the new SHA, but the git write-back never commits. `selfHeal` then reconciles against the stale git state and reverts the live Applications to the previous release's SHA.

Consider wrapping the rebase in an `|| { git rebase --abort; ... }` handler or switching to a merge strategy.

harry-rhesis (commented):
Code review (automated). 9 findings: 5 correctness, 2 maintenance/drift, 2 conventions. The top 3 share a common theme: the promote-prd-config job has no graceful partial-failure handling under `set -e`, which can leave prd in an inconsistent state.

.github/workflows/publish-release.yml
```diff
@@ -27,9 +27,168 @@ jobs:
...
+            prd-internal-dns
+            prd-grafana-resources
+            prd-alloy
+            prd-cluster-dns
+            prd-external-dns
+            argocd
+            rhesis
+          )
+
+          for app in "${APPS[@]}"; do
```
harry-rhesis: **A permanent failure on any app in the loop aborts the step, leaving prd split across two revisions.**

If app #9 hits a non-retryable ArgoCD error after exhausting retries, `set -e` aborts the entire step. Apps 1-8 are already live-synced to `$GITHUB_SHA`; apps 9-17 stay on the previous release. The git write-back never runs either.

Consider collecting errors and attempting all apps before deciding whether to fail, so you at least get a complete picture of what succeeded and what didn't.

.github/workflows/publish-release.yml
```diff
@@ -27,9 +27,168 @@ jobs:
...
+          # transiently fail with "another operation is already in progress" if
+          # an app's own selfHeal reconcile is mid-flight when this runs (observed
+          # in testing against stg) — not caught by the concurrency group above.
+          retry_argocd() {
+            for attempt in $(seq 1 5); do
+              EXIT_CODE=0
+              OUTPUT=$("$@" 2>&1) || EXIT_CODE=$?
+              echo "$OUTPUT"
+              [ "$EXIT_CODE" -eq 0 ] && return 0
+              if ! echo "$OUTPUT" | grep -qi "another operation"; then
```
harry-rhesis: **`retry_argocd` only retries "another operation" errors. Sync timeouts and network failures fail immediately.**

A large chart like kube-prometheus-stack exceeding the 300s timeout exits non-zero with a deadline-exceeded message. The `grep -qi 'another operation'` doesn't match, so it returns 1 on the first attempt. Same for transient DNS/TCP errors.

Consider broadening the retry condition to cover timeout and connection errors, or defaulting to retry on any failure with a specific exclusion list for genuinely non-retryable cases.

.github/workflows/publish-release.yml
```diff
@@ -27,9 +27,168 @@ jobs:
...
+            git rebase origin/main
+            sleep 2
+          done
+          echo "Push failed after 3 attempts."
+          exit 1
+
   publish:
     if: startsWith(github.ref, 'refs/heads/release/')
-    needs: [backend-deploy, frontend-deploy, worker-deploy]
+    needs: [backend-deploy, frontend-deploy, worker-deploy, promote-prd-config]
```
harry-rhesis, line 205: **Adding `promote-prd-config` to `publish`'s `needs` means a git-push failure blocks the release tag, even though container deploys succeeded.**

If the git push to main fails after retries (e.g. persistent non-fast-forward), no GitHub release is cut, no tag is created, and the release PR stays open. Prd is already running the new code with no corresponding release record.

Consider making the publish step tolerate a promote-prd-config failure (or at least still creating the release/tag), since the deployment itself already succeeded.

.github/workflows/publish-release.yml
```diff
@@ -27,9 +27,168 @@ jobs:
...
+            retry_argocd argocd app sync "$app" \
+              --server "$ARGOCD_SERVER" --auth-token "$ARGOCD_AUTH_TOKEN" --grpc-web --timeout 300
+          done
+
+      # Writes the promoted SHA back to git so the checked-in manifests match what
+      # was just synced above — this is what keeps disaster recovery correct (a
+      # `kubectl apply` of these files, see kubernetes/README.md) and what prd-base's
+      # own selfHeal will treat as desired state on its next reconcile. It does not
+      # itself perform the promotion (the explicit set/sync above does); run this
+      # promptly after so prd-base's selfHeal doesn't revert the live overrides
```
harry-rhesis: **`selfHeal` race between ArgoCD promotion and git write-back.**

The ArgoCD set/sync loop completes, but the git write-back (token gen, checkout, sed, commit, push with retries) can take minutes. If `selfHeal` fires before the commit lands on main, it re-applies the old `targetRevision` from the checked-in manifests and silently reverts the promotion.

This is a known constraint, but worth considering whether temporarily disabling `selfHeal` on `prd-base` during the promotion window (or doing the git write-back before the ArgoCD sync) would close the gap.

.github/workflows/publish-release.yml
```diff
@@ -27,9 +27,168 @@ jobs:
...
+          }
+
+          # Set/sync every app explicitly rather than relying on prd-base's sync to
+          # cascade the new revision to its children: prd-base's own git state (at
+          # $GITHUB_SHA) still carries whatever targetRevision was last committed,
+          # so applying it alone would leave children pinned to the previous
+          # release. prd-base first (root app-of-apps), then every nested app it
+          # manages. cnpg-operator is intentionally excluded — manual sync, manual
+          # promotion (see kubernetes/README.md).
+          APPS=(
```
harry-rhesis: **APPS array, FILES array, and `kustomization.yaml` are three independently maintained lists of the same prd Applications.**

Adding a new prd Application to `kustomization.yaml` without updating the workflow arrays means it never gets pinned during promotion. Nothing in CI detects this drift.

Consider deriving the lists from a single source (e.g. parsing `kustomization.yaml`) or adding a CI check that verifies all three are in sync.

.github/workflows/publish-release.yml
```diff
@@ -27,9 +27,168 @@ jobs:
...
+      - name: Pin every prd Application to this release's commit and sync
+        env:
+          ARGOCD_SERVER: ${{ secrets.ARGOCD_SERVER_PRD }}
+          ARGOCD_AUTH_TOKEN: ${{ secrets.ARGOCD_AUTH_TOKEN_PRD }}
+        run: |
+          # Retry up to 5 times with 30s back-off: `argocd app set`/`app sync` can
+          # transiently fail with "another operation is already in progress" if
+          # an app's own selfHeal reconcile is mid-flight when this runs (observed
+          # in testing against stg) — not caught by the concurrency group above.
+          retry_argocd() {
```
harry-rhesis: **ArgoCD CLI install and retry logic duplicated from `k8s-deploy.yml`.**

The install step (version v2.14.4, SHA256 pin, curl/chmod) and the retry loop are byte-for-byte copies. A version bump or retry fix in one file won't propagate to the other. Consider extracting to a composite action or a shared script.

.github/workflows/publish-release.yml
```diff
@@ -27,9 +27,168 @@ jobs:
...
+
+      - name: Pin every prd Application to this release's commit and sync
+        env:
+          ARGOCD_SERVER: ${{ secrets.ARGOCD_SERVER_PRD }}
+          ARGOCD_AUTH_TOKEN: ${{ secrets.ARGOCD_AUTH_TOKEN_PRD }}
+        run: |
+          # Retry up to 5 times with 30s back-off: `argocd app set`/`app sync` can
+          # transiently fail with "another operation is already in progress" if
+          # an app's own selfHeal reconcile is mid-flight when this runs (observed
+          # in testing against stg) — not caught by the concurrency group above.
```
harry-rhesis: **Comment references a "concurrency group above" that doesn't exist in this workflow.**

Copied from `k8s-deploy.yml` where that group exists. This workflow has no concurrency block, and nothing prevents two concurrent `workflow_dispatch` triggers from racing through `promote-prd-config` simultaneously.

.github/workflows/publish-release.yml
```diff
@@ -27,9 +27,168 @@ jobs:
...
+            sed -i "s|targetRevision: .*|targetRevision: $GITHUB_SHA|" "$f"
+          done
+
+          if git diff --quiet -- "${FILES[@]}"; then
+            echo "All prd Applications already pinned to $GITHUB_SHA — nothing to commit."
+            exit 0
+          fi
+
+          git add "${FILES[@]}"
+          git commit -m "chore(prd): pin prd Application targetRevisions to $GITHUB_SHA"
```
harry-rhesis: **Automated commit uses scope `prd`, which isn't in the allowed scope list.**

Per AGENTS.md, allowed scopes are `backend`, `frontend`, `sdk`, `tests`, or `dev`. This step commits `chore(prd): pin prd Application targetRevisions to $GITHUB_SHA` on every release. The matching scope would be `dev` (infrastructure/DevOps).

harry-rhesis (commented):
Thanks for putting this together! The prd promotion workflow is a solid addition. A few observations below, mostly around edge cases in failure handling. (This review supersedes the previous one.)

.github/workflows/publish-release.yml
harry-rhesis, line 162: Nice work on the retry logic here. One edge case to consider: if two releases run close together and both modify the same `targetRevision` lines, the second run's `git push` gets rejected. The retry path runs `git rebase origin/main`, which would conflict on those identical lines. Under `set -e`, the step aborts immediately. Since the ArgoCD set/sync step already completed, prd would be live at the new SHA but the git write-back would never commit. `selfHeal` could then reconcile against the stale git state and revert the live Applications.

Wrapping the rebase in an `|| { git rebase --abort; ... }` handler or using a merge strategy could help here.

.github/workflows/publish-release.yml
harry-rhesis: Worth thinking about what happens if one app in the loop hits a genuinely non-retryable error after exhausting retries. Since `set -e` is active, the entire step aborts. Apps that were already synced are live at `$GITHUB_SHA`, while the rest stay on the previous release. The git write-back doesn't run either.

Collecting errors and attempting all apps before deciding whether to fail would give a complete picture of what succeeded and what didn't. That way a single flaky app doesn't block the rest.

.github/workflows/publish-release.yml
harry-rhesis: The retry logic is a good idea. Right now it only matches "another operation" errors. Sync timeouts (e.g. a large chart like kube-prometheus-stack exceeding the 300s timeout) and transient network failures would fail immediately on the first attempt, even though they'd likely succeed on retry.

Broadening the retry condition to cover timeout and connection errors, or defaulting to retry-on-any-failure with a specific exclusion list, would make this more resilient.

.github/workflows/publish-release.yml
harry-rhesis, line 205: Something to consider: adding `promote-prd-config` to `publish`'s `needs` means a git-push failure blocks the release tag and PR merge, even though the container deploys already succeeded. Prd would be running new code with no corresponding release record.

It might be worth making `publish` tolerate a `promote-prd-config` failure (or at least still creating the release/tag), since the deployment itself already landed.

.github/workflows/publish-release.yml
harry-rhesis: There's a timing window worth being aware of: the ArgoCD set/sync loop completes, but the git write-back (token gen, checkout, sed, commit, push with retries) can take a few minutes. If `selfHeal` fires before the commit lands on main, it re-applies the old `targetRevision` from the checked-in manifests and silently reverts the promotion.

Temporarily disabling `selfHeal` on `prd-base` during the promotion window, or doing the git write-back before the ArgoCD sync, could close this gap.

.github/workflows/publish-release.yml
harry-rhesis: Heads up: the APPS array, FILES array, and `kustomization.yaml` are now three independently maintained lists of the same prd Applications. Adding a new app to `kustomization.yaml` without updating the workflow arrays would mean it never gets pinned during promotion, and nothing in CI would catch the drift.

Deriving the lists from a single source (e.g. parsing `kustomization.yaml`) or adding a CI check to verify they stay in sync would help prevent that.

.github/workflows/publish-release.yml
harry-rhesis: Minor maintenance note: the ArgoCD CLI install step (version v2.14.4, SHA256 pin, curl/chmod) and the retry loop are duplicated from `k8s-deploy.yml`. A version bump or retry fix in one file wouldn't propagate to the other. Extracting to a composite action or shared script would keep them in sync.

.github/workflows/publish-release.yml
harry-rhesis: Small copy-paste artifact: the comment references a "concurrency group above" that exists in `k8s-deploy.yml` but not in this workflow. Also worth noting that this workflow has no concurrency block, so two concurrent `workflow_dispatch` triggers could race through `promote-prd-config` at the same time.

.github/workflows/publish-release.yml
harry-rhesis: Tiny nit: the automated commit uses scope `prd` (`chore(prd): pin prd Application...`), but per AGENTS.md the allowed scopes are `backend`, `frontend`, `sdk`, `tests`, or `dev`. The matching scope here would be `dev` (infrastructure/DevOps).

asad-miah:
Thanks for the thorough review, @harry-rhesis! Pushed a fix in b007961e9 that addresses these:

**Fixed:**
- **selfHeal race** (line 120) — this was a bigger problem than my earlier code comment suggested: it wasn't a brief race, `prd-base`'s selfHeal would have kept reverting children indefinitely, since it reconciles against an immutable commit. Fixed by moving the git write-back *before* the ArgoCD sync, and pinning `prd-base` to the resulting write-back commit instead of `$GITHUB_SHA` directly. Its tree is now self-consistent from the moment it's synced, so the cascade to children works correctly without fighting selfHeal. Details in the updated job comment and `kubernetes/README.md`.
- **Partial failure mid-loop** (line 108) — moot now: this fix also removed the `APPS` array/loop. `prd-base` is the only app explicitly set/synced; children cascade from its own sync, so there's no multi-app loop left to fail partway through.
- **Retry only matches "another operation"** (line 67) — broadened: retries on any failure now except a short exclusion list of clearly permanent errors (permission/auth/not-found/invalid-argument), so a slow sync or network blip gets retried instead of failing on attempt 1.
- **Commit scope `prd`** (line 175) — changed to `chore(dev): ...` per AGENTS.md's allowed scopes.
- **Stale "concurrency group" comment** (line 60) — gone as part of rewriting that block.

**Not fixed, judgment calls left as-is:**
- **`git rebase` under `set -e`** (line 183) — still relies on implicit `set -e` to fail the step on conflict, rather than an explicit abort handler. The job still fails correctly, just with a less clean message; only bites if two releases run within minutes of each other.
- **`promote-prd-config` in `publish`'s `needs`** (line 191) — left coupled on purpose: if prd promotion fails, we don't want to cut a release tag/PR merge for a deploy that didn't fully land.

**Deferred as follow-ups, not blocking this PR:**
- **Three independently maintained lists** (line 88) — now down to two (`FILES` array + `kustomization.yaml`) since the `APPS` array is gone. Still worth a CI check eventually to catch drift when a new prd Application is added.
- **Duplicated ArgoCD CLI install/retry logic vs `k8s-deploy.yml`** (line 61) — real duplication, no correctness impact. Worth extracting to a composite action later.

harry-rhesis:
Thanks for the thorough work here, @asad-miah! The promotion workflow has come a long way across these iterations. Here's a summary of where things stand.

## Addressed

- **selfHeal race**: git write-back now runs before ArgoCD sync, and prd-base is pinned to the write-back commit. This closes the window where selfHeal could revert children.
- **Partial failure mid-loop**: the APPS loop is gone entirely. Only prd-base is explicitly set/synced; children cascade from its sync.
- **Narrow retry condition**: retry logic now covers any failure except a short exclusion list of permanent errors (auth, not-found, etc.), so timeouts and network blips get retried.
- **Commit scope `prd`**: changed to `chore(dev)` per AGENTS.md.
- **Stale "concurrency group" comment**: removed.
- **RELEASING.md link anchor**: fixed to `#production-promotion-gate`.
- **Child apps not actually promoted** (peqy's earlier critical): fixed by the write-back-first approach.
- **`git push` non-fast-forward flaky** (peqy): retry loop with fetch+rebase added.

## Acknowledged, left as-is

- **`git rebase` under `set -e`**: accepted. Since the rebase now happens before prd-base is promoted, a conflict fails the job before ArgoCD is touched. No "prd live but git stale" state.
- **`promote-prd-config` in publish's `needs`**: intentional coupling so a failed promotion blocks the release tag.

## Acknowledged, deferred

- **Two independently maintained lists** (FILES array + kustomization.yaml): down from three now that the APPS array is gone. A CI check to catch drift is worth doing as a follow-up.
- **Duplicated ArgoCD CLI install/retry logic**: real duplication with `k8s-deploy.yml`, no correctness impact. Worth extracting to a composite action later.

## Still open

1. **`base.yaml` self-reference (peqy, critical)**: `base.yaml` is in the `FILES` write-back and gets `targetRevision: $GITHUB_SHA`. But prd-base manages its own manifest (base.yaml is a resource in kustomization.yaml). When prd-base syncs at `PROMOTED_SHA`, it re-applies base.yaml which says `targetRevision: $GITHUB_SHA`. That overwrites prd-base's own revision back to `$GITHUB_SHA`, which predates the write-back commit and still has the old child targetRevisions. selfHeal would then revert children indefinitely. This needs a fix before merge: either remove base.yaml from the FILES array (so prd-base's revision is only controlled by the explicit `argocd app set`), remove base.yaml from kustomization.yaml (so prd-base doesn't manage itself), or write `PROMOTED_SHA` into base.yaml instead of `$GITHUB_SHA` (which requires a two-pass commit).

2. **Main-tip drift (peqy, improvement)**: `PROMOTED_SHA` is created from the current tip of `main`. If `main` has picked up any `kubernetes/clusters/prd/**` changes beyond targetRevision pins since the release branch was cut, those would ship to prd unvalidated. A guard that fails when `git diff $GITHUB_SHA..origin/main -- kubernetes/clusters/prd` contains non-targetRevision changes would catch this.

3. **Broad `sed` replacement (peqy, improvement, raised across 4 rounds)**: `sed -i "s|targetRevision: .*|...|"` replaces every match in each file. If a file ever gains a second `targetRevision:` field (e.g. from a Helm source with multiple sources), it would be silently corrupted. At minimum, asserting exactly one match per file (`grep -c`) would fail fast. Using `yq` to set `.spec.source.targetRevision` would be the most robust option.

Items 1-3 are from peqy's latest CHANGES_REQUESTED reviews, which landed after the most recent commit. Item 1 is a correctness issue that could cause prd to revert on every release, so it should be addressed before merge.

asad-miah (commented):
Re: the `\s` finding — both GNU grep (Ubuntu / GitHub Actions runners) and BSD grep (macOS) support `\s` as a whitespace shorthand. Verified against all 7 prd Application YAMLs and in a `ubuntu:22.04` Docker container — each file returns exactly 1 match with `grep -c '^\s*targetRevision:'`.

The guard works correctly as written. No change needed.

asad-miah:
> Thanks for the thorough work here, @asad-miah! The promotion workflow has come a long way across these iterations. Here's a summary of where things stand.
> 
> ## Addressed
> * **selfHeal race**: git write-back now runs before ArgoCD sync, and prd-base is pinned to the write-back commit. This closes the window where selfHeal could revert children.
> * **Partial failure mid-loop**: the APPS loop is gone entirely. Only prd-base is explicitly set/synced; children cascade from its sync.
> * **Narrow retry condition**: retry logic now covers any failure except a short exclusion list of permanent errors (auth, not-found, etc.), so timeouts and network blips get retried.
> * **Commit scope `prd`**: changed to `chore(dev)` per AGENTS.md.
> * **Stale "concurrency group" comment**: removed.
> * **RELEASING.md link anchor**: fixed to `#production-promotion-gate`.
> * **Child apps not actually promoted** (peqy's earlier critical): fixed by the write-back-first approach.
> * **`git push` non-fast-forward flaky** (peqy): retry loop with fetch+rebase added.
> 
> ## Acknowledged, left as-is
> * **`git rebase` under `set -e`**: accepted. Since the rebase now happens before prd-base is promoted, a conflict fails the job before ArgoCD is touched. No "prd live but git stale" state.
> * **`promote-prd-config` in publish's `needs`**: intentional coupling so a failed promotion blocks the release tag.
> 
> ## Acknowledged, deferred
> * **Two independently maintained lists** (FILES array + kustomization.yaml): down from three now that the APPS array is gone. A CI check to catch drift is worth doing as a follow-up.
> * **Duplicated ArgoCD CLI install/retry logic**: real duplication with `k8s-deploy.yml`, no correctness impact. Worth extracting to a composite action later.
> 
> ## Still open
> 1. **`base.yaml` self-reference (peqy, critical)**: `base.yaml` is in the `FILES` write-back and gets `targetRevision: $GITHUB_SHA`. But prd-base manages its own manifest (base.yaml is a resource in kustomization.yaml). When prd-base syncs at `PROMOTED_SHA`, it re-applies base.yaml which says `targetRevision: $GITHUB_SHA`. That overwrites prd-base's own revision back to `$GITHUB_SHA`, which predates the write-back commit and still has the old child targetRevisions. selfHeal would then revert children indefinitely. This needs a fix before merge: either remove base.yaml from the FILES array (so prd-base's revision is only controlled by the explicit `argocd app set`), remove base.yaml from kustomization.yaml (so prd-base doesn't manage itself), or write `PROMOTED_SHA` into base.yaml instead of `$GITHUB_SHA` (which requires a two-pass commit).
> 2. **Main-tip drift (peqy, improvement)**: `PROMOTED_SHA` is created from the current tip of `main`. If `main` has picked up any `kubernetes/clusters/prd/**` changes beyond targetRevision pins since the release branch was cut, those would ship to prd unvalidated. A guard that fails when `git diff $GITHUB_SHA..origin/main -- kubernetes/clusters/prd` contains non-targetRevision changes would catch this.
> 3. **Broad `sed` replacement (peqy, improvement, raised across 4 rounds)**: `sed -i "s|targetRevision: .*|...|"` replaces every match in each file. If a file ever gains a second `targetRevision:` field (e.g. from a Helm source with multiple sources), it would be silently corrupted. At minimum, asserting exactly one match per file (`grep -c`) would fail fast. Using `yq` to set `.spec.source.targetRevision` would be the most robust option.
> 
> Items 1-3 are from peqy's latest CHANGES_REQUESTED reviews, which landed after the most recent commit. Item 1 is a correctness issue that could cause prd to revert on every release, so it should be addressed before merge.

1. base.yaml self-reference — removed from prd's kustomization.yaml (dev/stg unchanged), with comments in both the kustomization file and README explaining why prd is the exception.
2. Main-tip drift — new guard step fails the job if main has picked up unrelated kubernetes/clusters/prd/** changes since the release branch was cut.
3. Broad sed — asserts exactly one targetRevision: match per file before editing, fails fast otherwise.

asad-miah:
Checked this — the claim isn't accurate for this environment, so no fix needed here.

`promote-prd-config` runs on `arc-runner-prd`, which uses GitHub's official runner image (`ghcr.io/actions/actions-runner`) — Ubuntu 22.04, so `grep` there is GNU grep. GNU grep supports `\s` as whitespace even in basic (non-extended) mode; it's a documented, long-standing GNU extension, not something limited to `-E`/PCRE. I tested the exact pattern against the real indented file content on two separate grep implementations (BSD grep and GNU-compatible grep) and both matched correctly — 1 match, not 0:

```
$ grep -c '^\s*targetRevision:' kubernetes/clusters/prd/arc.yaml
1
```

So `grep -c '^\s*targetRevision:'` does not find 0 matches and would not abort prd publishes. `[[:space:]]` is the more strictly POSIX-portable form and would work identically, but since the actual runner is GNU grep on Ubuntu, this isn't a blocker — leaving it as-is.

asad-miah:
> Checked this — the claim isn't accurate for this environment, so no fix needed here.
> 
> `promote-prd-config` runs on `arc-runner-prd`, which uses GitHub's official runner image (`ghcr.io/actions/actions-runner`) — Ubuntu 22.04, so `grep` there is GNU grep. GNU grep supports `\s` as whitespace even in basic (non-extended) mode; it's a documented, long-standing GNU extension, not something limited to `-E`/PCRE. I tested the exact pattern against the real indented file content on two separate grep implementations (BSD grep and GNU-compatible grep) and both matched correctly — 1 match, not 0:
> 
> ```
> $ grep -c '^\s*targetRevision:' kubernetes/clusters/prd/arc.yaml
> 1
> ```
> 
> So `grep -c '^\s*targetRevision:'` does not find 0 matches and would not abort prd publishes. `[[:space:]]` is the more strictly POSIX-portable form and would work identically, but since the actual runner is GNU grep on Ubuntu, this isn't a blocker — leaving it as-is.

Checked this — the claim isn't accurate for this environment, so no fix needed here.

promote-prd-config runs on arc-runner-prd, which uses GitHub's official runner image (ghcr.io/actions/actions-runner) — Ubuntu 22.04, so grep there is GNU grep. GNU grep supports \s as whitespace even in basic (non-extended) mode; it's a documented, long-standing GNU extension, not something limited to -E/PCRE. I tested the exact pattern against the real indented file content on two separate grep implementations (BSD grep and GNU-compatible grep) and both matched correctly — 1 match, not 0:

$ grep -c '^\s*targetRevision:' kubernetes/clusters/prd/arc.yaml
1
So grep -c '^\s*targetRevision:' does not find 0 matches and would not abort prd publishes. [[:space:]] is the more strictly POSIX-portable form and would work identically, but since the actual runner is GNU grep on Ubuntu, this isn't a blocker — leaving it as-is.

PR #2354, opened 2026-08-04
Add usage accounting for metered resources

harry-rhesis:
## Purpose

Sub-plan 2 of usage accounting. Sub-plan 1 (#2350) shipped the `QuotaResource` enum, the quota registry and YAML-driven tier limits. This adds the other half: a durable record of what each organization has actually consumed, and a read-only endpoint to report it.

No enforcement here. Nothing is blocked, no request is rejected. That is Sub-plan 3.

## What Changed

**Counter storage.** A `usage` table holding one row per organization, resource and billing period, unique on `(organization_id, resource, period_start)` so accrual can upsert against it. It deliberately does not use `OrganizationMixin`: the service writes through raw atomic upserts that the ORM auto-filter listeners are not built for, so a FORCE'd `tenant_isolation` RLS policy is the tenant boundary instead. It is also excluded from the generic recycle-bin routes, where an org member could otherwise hard-delete their own row to zero their counters.

**Two kinds of resource.** Flow resources (test executions, tracing spans, test generation, model tokens) are cumulative counters for the current UTC calendar month. Stock resources (seats, projects, endpoints) are live entity counts computed on read, since there is nothing to accrue or reset. `GET /usage` returns both with a `kind` field so consumers group them without keeping their own copy of that split.

**One accrual path.** `dispatch_accrual` is the only entry point call sites use. It queues the write on a Celery worker and never raises, so a failed counter can neither slow down nor break the operation it measures. `increment_usage` is the database primitive behind it, called only by that worker, where committing the session and raising on failure are safe because the session belongs to nothing else.

**Token usage from the SDK.** `BaseLLM` takes an optional `on_usage` callback that providers invoke at the point they parse `usage` out of their own API response. A plain constructor kwarg keeps the returned object a real provider instance, so every `isinstance(model, BaseLLM)` check downstream keeps working. Payloads are normalized to a `TokenUsage` before the callback sees them, reusing the existing `extract_token_usage` helper rather than growing a second copy of provider-dialect matching.

**Where tokens are counted.** Hosted models (Rhesis or Polyphemus with no user-supplied API key) get the callback wired in at construction, because Rhesis pays for that inference. A model using the org's own provider key gets none, since that usage is billed to them directly.

## Additional Context

Three counting bugs surfaced while wiring this up and are fixed here:

- `/services/generate/content` discarded usage entirely, so the `rhesis-default` provider never accrued anything. That endpoint has two kinds of caller and both must count. A developer calling it directly is billed against their own org. A relaying `RhesisLLM` (self-hosted delegation, or a same-deployment loopback) needs its own local counter to move too, so the endpoint composes rather than replaces: it accrues normally *and* forwards the usage via an `X-Rhesis-Usage` header. Two books for two orgs, not a double-count of one. The header carries it because that endpoint returns bare content with no envelope for a sibling `usage` field.
- Polyphemus's batch path read `item['usage']['total_tokens']` directly and dropped any item that reported only prompt/completion counts.
- `RhesisLLM` read usage off the response body, where a dict is the caller's own schema-validated payload, so a schema containing a `usage` field could be mistaken for token counts.

Two notes for reviewers:

- Billing periods are UTC-anchored rather than local. With local dates, workers in different timezones disagree about which month a call near midnight belongs to and accrue to two different rows.
- Because accrual is queued, the counter is eventually consistent and can lag by seconds. That is the right tradeoff for reporting and for the soft enforcement planned in Sub-plan 3, but worth confirming when that gate is built.

On not using OpenTelemetry for this: the repo has two OTel systems and neither can be the billing source. `apps/backend/telemetry/` is opt-in product analytics that drops spans when a user disables telemetry. `sdk/telemetry/` traces the *customer's* application, so those tokens are the customer's own spend rather than Rhesis-hosted inference. Enforcement also needs a transactional read to return a 402, which a metrics backend cannot serve. The one thing worth borrowing was the naming, so `TokenUsage` uses the field names of the OpenAI Responses API `usage` object, which is also what OTel's `gen_ai.usage.*` conventions use.

The frontend dashboard that consumes `GET /usage` is a follow-up PR stacked on this branch.

## Testing

```bash
# backend
cd apps/backend && uv run pytest ../../tests/backend/services/test_usage.py \
  ../../tests/backend/routes/test_usage.py ../../tests/backend/app/test_usage_tracking.py \
  ../../tests/backend/tasks/ ../../tests/backend/routes/test_services.py \
  ../../tests/backend/utils/test_llm_delegation.py ../../tests/backend/ee/licensing/

# sdk
cd sdk && uv run pytest ../tests/sdk/ --ignore=../tests/sdk/integration
```

707 backend and 2264 SDK tests pass locally. `ruff check` and `ruff format` clean on all touched files.

Manual check: call a hosted model, then `GET /usage`, and confirm `model_tokens` reflects the call after the worker drains.

### DIFF
--- a/apps/backend/src/rhesis/backend/alembic/versions/77df3dbea77d_add_usage_table.py
+++ b/apps/backend/src/rhesis/backend/alembic/versions/77df3dbea77d_add_usage_table.py
@@ -115,6 +115,11 @@ def upgrade() -> None:
             ["organization_id"],
         )
 
+    # Not a duplicate of Base.deleted_at's `index=True`: raw op.create_table
+    # above does not honor that ORM-level flag, so nothing indexes this
+    # column until this explicit op.create_index runs. Same two-step
+    # (unindexed column in create_table, then this idempotent check) as
+    # c7f4d9b2e1a3_add_auth_client_table.py's ix_auth_client_deleted_at.
     deleted_at_index_exists = conn.execute(
         sa.text(
             "SELECT 1 FROM pg_indexes WHERE tablename='usage' AND indexname='ix_usage_deleted_at'"
--- a/apps/backend/src/rhesis/backend/alembic/versions/77df3dbea77d_add_usage_table.py
+++ b/apps/backend/src/rhesis/backend/alembic/versions/77df3dbea77d_add_usage_table.py
@@ -1,0 +1,149 @@
+"""add usage table
+
+Per-organization, per-resource, per-billing-period cumulative usage
+counter backing the read-only ``GET /usage`` endpoint. ``resource``
+stores ``QuotaResource`` string values (see
+``rhesis.backend.app.quota``); the unique constraint on
+``(organization_id, resource, period_start)`` is what
+``increment_usage`` upserts against.
+
+``organization_id`` cascades on delete so removing an org does not
+leave orphaned counters behind.
+
+RLS: ``usage`` gets the same ``tenant_isolation`` policy (ENABLE +
+FORCE) as every other tenant table, following the per-table pattern
+established after the original blanket-RLS migration (see e.g.
+``d9e0f1a2b3c4_add_rbac_catalog_tables.py``) -- new tables are not
+covered automatically. This matters more than usual here: the usage
+service intentionally runs its reads/writes under
+``bypass_tenant_filter()`` (the ORM-level auto-filter, which the raw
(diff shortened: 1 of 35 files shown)


harry-rhesis:
@peqy could you take a look at this PR?

PR #2362, opened 2026-08-04
Fix: Email Invitation Failed

asad-miah:
This PR introduces changes from the `fix/email-invitation-failed` branch.

## 📁 Files Changed (       1 files)

```
apps/backend/src/rhesis/backend/app/utils/validation.py
```

## 📋 Commit Details

```
072b4a7b4 - fix(backend): use public DNS for email deliverability check (Md Asaduzzaman Miah, 2026-08-04 13:16)
```

## ✅ Checklist

- [ ] All tests pass locally

### DIFF
--- a/apps/backend/src/rhesis/backend/app/auth/providers/email.py
+++ b/apps/backend/src/rhesis/backend/app/auth/providers/email.py
@@ -8,6 +8,7 @@ including user registration and login.
 import logging
 from typing import Optional
 
+import anyio
 from fastapi import HTTPException, Request, status
 from sqlalchemy.orm import Session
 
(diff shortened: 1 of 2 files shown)


harry-rhesis (approved):
Looks good. The blocking DNS call in an async endpoint was a real problem and `anyio.to_thread.run_sync` is the right fix. The custom resolver to bypass internal cluster DNS makes sense given the BIND9 setup, and the 3s lifetime is a reasonable cap.

Checked the other callers of `validate_and_normalize_email` -- the only other one with `check_deliverability=True` is the sync `create_user()` endpoint, which FastAPI already runs in a threadpool, so no change needed there.

Thanks for the thorough comments explaining the DNS and timeout context.

PR #2374, opened 2026-08-05
Let the architect read annotations and create complete metrics

nicolai-rhesis:
## Purpose

Two gaps in what the architect agent can see and produce.

**It could not see human review feedback.** Everything the architect can do comes from one whitelist, `mcp_server/mcp_tools.yaml` — a route absent from that file is unreachable. `GET /annotations/` was absent, and there is no trace or telemetry tool at all, so trace reviews were entirely out of reach. Reviews on test results leaked through incidentally (`test_reviews` sits on `TestResultBase`, so `get_test_result` returns it) but nothing told the agent it was there and there was no way to filter for annotated results.

**Metrics it created arrived nearly empty.** Description, evaluation steps, reasoning instructions and result explanation were all blank on generated metrics — the fields exist and the API accepts them, but nothing asked for them.

### Annotations become readable

- `list_annotations` filters added: `test_run_id`, `test_result_id`, `trace_id`, `trace_db_id`. Pushed into the per-source branch `WHERE` clauses so they prune before the `UNION ALL` and hit existing indexes.
- `test_run_id` and `test_result_id` deliberately narrow **both** branches. `trace` carries its own `test_run_id`/`test_result_id`, so one run-scoped call returns reviews on that run's test results *and* on the traces the run produced — which is what makes this useful for failure analysis.
- `trace_id`/`trace_db_id` are trace-only, so they disable the test-result branch, reusing the narrowing the existing `source` handling already does. Following its documented rule: turn the other branch off, never force a branch on, because the flags may already encode a permission denial.
- Trace-only filters raise 403 without `telemetry:read` rather than returning a misleading empty list. `test_run_id` spans both sources so it degrades to the trace side instead of denying.
- Trace annotations now report their real `test_run_id`/`test_result_id` instead of `NULL`. The branch filtered on those columns but projected null, so a run-scoped call returned trace reviews that could not say which run they came from. `TraceResponse` already exposes both ids to the same `telemetry:read` holders, so this discloses nothing new; they stay null only for traces outside a test execution.
- Exposed as the read-only `list_annotations` MCP tool, plus a guideline telling the agent to check human feedback when analysing a run and to treat a human verdict as outranking a metric score.

The write side is deliberately **not** included — an agent closing out someone else's review is a separate decision.

### Metrics get created rich

- `create_metric` now documents `description`, `evaluation_steps`, `reasoning` and `explanation`. These were already in the tool's input schema via `MetricCreate`, but `MetricBase` carries no Pydantic field descriptions, so they reached the agent bare, with no reason to fill them. The YAML overrides are the only place that guidance can live.
- `GeneratedMetric` (behind `generate_metric`/`improve_metric`) had **no** `reasoning` or `explanation` field at all and never asked for `evaluation_steps`. All three are now required, so structured output cannot skip them, and both Jinja templates ask for them explicitly.
- The improve path had a field whitelist that never showed `reasoning`/`explanation` to the LLM. Left alone, requiring them in the schema would have blanked those fields on every improve, since the template tells the model to preserve what the edit request does not mention — it cannot preserve what it is not shown.

### Entity links stop lying about detail pages

Both prompt templates told the agent that behaviors and metrics have no detail pages, so it referred to them by name and suppressed links a user could have followed. Both have `[identifier]` routes. The two templates also disagreed with each other — `streaming_response.j2` listed metrics as linkable while `telemachus-guidelines.j2` did not. Both now carry behavior, metric and trace link patterns, and the negative claim is narrowed to test results, which genuinely have no detail page.

## Additional Context

- No migration. Every column and schema field already existed; this is about what the agent is told and what it can filter on.
- `list_annotations` inherits the endpoint's existing fail-closed behaviour: it needs an ambient project and 400s without one. Architect sessions satisfy that — `LocalToolProvider` sends `X-Project-Id` when it has a project, wired from the session in `services/architect/runner.py`.
- Making the three descriptive fields required on `GeneratedMetric` is a behaviour change for any direct `MetricSynthesizer` caller: an LLM response missing one is now a validation error rather than a silent `None`.
- `TestEntityLinkGuidanceMatchesFrontend` derives the linkable-entity set from `apps/frontend/src/app/(protected)/*/[identifier]`, so an SDK test now reads the frontend route tree. It skips when that tree is absent (SDK checked out standalone). This is deliberate — the drift it catches is exactly what caused the stale claim.
- One incidental reformat: `ruff format` rewrapped the `NEW_TOOLS` frozenset in `test_mcp_tools_yaml.py`, a block this PR does not otherwise touch.

## Testing

```bash
cd apps/backend
uv run pytest ../../tests/backend/routes/test_annotations.py \
              ../../tests/backend/services/test_mcp_tools_yaml.py \
              ../../tests/backend/routes/test_metric.py
cd ../../sdk
uv run pytest ../tests/sdk/metrics/test_synthesizer.py \
              ../tests/sdk/agents/test_architect_prompt_loader.py
```

97 backend and 49 SDK tests pass. Docker must be running for the backend suite.

New coverage:

- Run-scoped annotations span both sources and exclude an unrelated run; result-scoped includes traces linked to it; each trace identifier returns traces only; contradictory filters return empty rather than erroring.
- Trace-only filter without `telemetry:read` → 403 (parametrised over both identifiers); run-scoped with only `telemetry:read` → trace side, no 403.
- `list_annotations` builds against the real OpenAPI, so a path or param drift fails the suite instead of silently dropping the tool — the failure mode the whitelist makes easy.
- The four descriptive fields reach the `create_metric` schema *with* descriptions; omitting any of them from `GeneratedMetric` is a validation error; the improve template renders all of them with their current values.

Worth a manual check: open the architect on a project with annotations and ask what people flagged on the last run, then have it create a metric and confirm the detail page has no blank fields.

### DIFF
--- a/apps/backend/src/rhesis/backend/app/mcp_server/mcp_tools.yaml
+++ b/apps/backend/src/rhesis/backend/app/mcp_server/mcp_tools.yaml
@@ -1013,13 +1013,16 @@ tools:
       - test_result_id: one result plus traces linked to it.
       - trace_id (32-char OpenTelemetry hex id) or trace_db_id (internal
         UUID): one trace only. Use whichever id you already have; both are
-        returned on every annotation.
+        returned on every trace annotation.
       - source: restrict to "test_result" or "trace" when you want one side.
 
       Key response fields: comments (the human's actual words),
       status.name (Pass/Fail), resolved, target.type (what was reviewed —
       test_result, trace, metric, or turn), plus behavior_name and the
-      test_run_id/test_result_id/trace_id deep-link ids.
+      test_run_id/test_result_id/trace_id deep-link ids. Annotations on a
+      test-execution trace carry the run and result ids too, so you can
+      walk from a trace review back to the run — they are null only for
+      traces recorded outside a test run.
 
       CHAIN: after get_test_result_stats shows failures, call this with the
       same test_run_id to see whether a human already diagnosed them — then
--- a/apps/backend/src/rhesis/backend/app/mcp_server/mcp_tools.yaml
+++ b/apps/backend/src/rhesis/backend/app/mcp_server/mcp_tools.yaml
@@ -1024,6 +1024,10 @@ tools:
       walk from a trace review back to the run — they are null only for
       traces recorded outside a test run.
 
+      When linking to a trace in the UI, use trace_db_id (a UUID) in the
+      URL path, not trace_id (a 32-char hex string). Only trace_db_id
+      resolves to a trace page.
+
       CHAIN: after get_test_result_stats shows failures, call this with the
       same test_run_id to see whether a human already diagnosed them — then
       use get_test_result on the ids it returns to read the full exchange.
--- a/apps/backend/src/rhesis/backend/app/mcp_server/mcp_tools.yaml
+++ b/apps/backend/src/rhesis/backend/app/mcp_server/mcp_tools.yaml
@@ -505,9 +505,43 @@ tools:
       assignee_id, status_id, model_id, backend_type_id, metric_type_id).
       If you only have a natural-language description, prefer
       generate_metric instead.
+
+      ALWAYS send the descriptive fields alongside the scoring
+      configuration: description, evaluation_steps, reasoning, and
+      explanation. They are optional to the API but a metric without them
+      is a poor metric — reviewers cannot tell what it measures, and the
+      judge LLM has less to work with, so scores get less consistent.
+      Sending only name + evaluation_prompt + score_type produces a
+      threadbare metric and is a bug, not a shortcut. Fill every field you
+      have grounds to fill.
     parameters:
       name:
         description: "Metric name, unique within the organization (required, string)"
+      description:
+        description: >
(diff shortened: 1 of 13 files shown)


nicolai-rhesis:
Good catch on the trace deep-link ids — that was a real inconsistency, not a deliberate permission choice. Fixed in 11da3eb66.

The trace branch projected `NULL::uuid` for both while the query filtered the same branch on `t.test_run_id` / `t.test_result_id`, so a run-scoped call returned trace annotations that could not say which run they came from. It also contradicted the tool description, which promised those ids on every annotation.

There was no permission argument for keeping them null: `TraceResponse` (`schemas/telemetry.py:111-119`) already exposes `test_run_id`, `test_result_id` and `test_id` to the same `telemetry:read` holders who can reach the trace branch at all, so projecting them discloses nothing that endpoint did not already hand over. They now stay null only for traces recorded outside a test execution, which is the column's own meaning.

Two tests cover it: the run-scoped test now asserts the trace annotation carries both ids, and a new `test_operation_trace_annotation_has_null_run_and_result` pins the null case so the distinction stays intentional.

PR #2377, opened 2026-08-05
Add platform key support for local-mode hosted models

alexeyhramov123:
## Purpose

Self-hosted/local deployments currently have no way to authenticate against Rhesis-hosted models (the `rhesis` provider and unauthenticated `polyphemus`) without a SaaS session. This adds an org-scoped Rhesis platform API key that local deployments can configure, so hosted models work locally instead of silently failing or being unusable.

## What Changed

- Backend: store an encrypted Rhesis platform API key on the organization row, with a local-mode gate and a `platform_key` service that validates and caches key status (validity, Polyphemus authorization) without hitting the network on hot paths.
- Backend: add `GET/PUT/DELETE /platform/rhesis-key` endpoints to read, set, and clear the org's platform key.
- Backend: in local mode, resolve Rhesis-hosted system models and unauthenticated Polyphemus models with the org-scoped platform key when one is configured, accruing usage like any other hosted call; non-local (SaaS) behavior is unchanged.
- Backend: annotate `available`/`availability_reason` on `GET /models` (and `GET /models/{id}`) so models that can't authenticate in local mode are flagged with a machine-readable reason (`rhesis_key_missing`, `rhesis_key_invalid`, `polyphemus_not_authorized`).
- Backend: unit tests for the platform key service (storing/clearing the key, masking, cached validation state, best-effort validation that never raises on network errors).
- Frontend: API client and `usePlatformKey` hook for the platform key endpoint.
- Frontend: grey out unavailable models on the Models page with their availability reason.
- Frontend: add a card to the Models page for configuring and validating the Rhesis platform key.

## Additional Context

Local/self-hosted only — deployments running `BACKEND_ENV=production` (SaaS) see no behavior change; every model stays available there.

## Testing

- `uv run pytest tests/backend/services/test_platform_key.py` (from `apps/backend/`).
- Manually: run the backend with `BACKEND_ENV=local`, open the Models page, configure a Rhesis platform key via the new card, and confirm previously-greyed `rhesis`/`polyphemus` models become available and usage accrues.


apps/backend/src/rhesis/backend/app/routers/platform.py
```diff
@@ -0,0 +1,79 @@
...
+    the cached result rather than re-hitting the network.
+    """
+    status = platform_key_service.get_platform_key_status(
+        db, current_user.organization_id, refresh=True
+    )
+    return PlatformKeyStatus(**status)
+
+
+@router.put("/rhesis-key", response_model=PlatformKeyStatus)
+def set_rhesis_key(
```
harry-rhesis, line 63: PUT/DELETE here only require `require_current_user_or_token` + `require_local_mode`, not the owner-only permission check that other org-settings mutations go through (`main.py`'s `apply_authz_backstop` on `PUT /organizations/{id}`). Right now any authenticated org member, not just the owner, can overwrite or clear the org's platform API key. Worth routing this through the same backstop.

apps/backend/src/rhesis/backend/app/services/platform_key.py
```diff
@@ -0,0 +1,290 @@
...
+        return result
+
+    if resp.status_code in (401, 403):
+        return {"valid": False, "polyphemus_authorized": False}
+    if resp.status_code >= 400:
+        # Server-side or unexpected error: key validity is undeterminable.
+        result["polyphemus_authorized"] = _probe_polyphemus_authorized(key)
+        return result
+
+    result["valid"] = True
```
harry-rhesis, line 285: `httpx.get()` doesn't follow redirects by default. If the platform host returns a 3xx, the status falls through both the 401/403 and >=400 branches, and `resp.json()` on the redirect body raises into a bare `except`, leaving `valid=True`. An invalid or unreachable key could get cached as valid. Might be worth handling 3xx explicitly or setting `follow_redirects=True`.

apps/backend/src/rhesis/backend/app/routers/model.py
```diff
@@ -33,6 +35,72 @@
...
+_PLATFORM_PROVIDERS = ("rhesis", "polyphemus")
+
+
+def _provider_of(model) -> str | None:
+    """Return a model's provider type value, guarding a missing provider_type."""
+    provider_type = getattr(model, "provider_type", None)
+    return getattr(provider_type, "type_value", None) if provider_type else None
+
+
+def _annotate_model_availability(db: Session, organization_id, models_list) -> None:
```
harry-rhesis: `_annotate_model_availability`/`_provider_of` put availability logic in the router, but the rest of this PR follows the routers -> services -> crud layering (see `platform_key.py`). Moving this into a service would let it get unit tested the same way and keeps it reusable for any future model-list endpoint.

apps/frontend/src/app/(protected)/models/components/ModelCard.tsx
```diff
@@ -12,7 +12,15 @@ import { Capability } from '@/constants/capabilities';
...
 import { PROVIDER_ICONS } from '@/config/model-providers';
-import type { ValidationStatus } from '../types';
+import { getAvailabilityReasonCopy, type ValidationStatus } from '../types';
+
+// Local/self-hosted deployments can't run the email-based Polyphemus access
+// request (mail isn't configured), so the "Request Access" affordance is hidden
+// there — backend availability (greyed with a reason) is the single source of
+// truth for whether a Rhesis/Polyphemus model can be used. Same signal as the
+// models page.
+const isLocalMode =
```
harry-rhesis: Local-mode detection relies on two separate env vars, backend `BACKEND_ENV` and frontend `NEXT_PUBLIC_FRONTEND_ENV`, that aren't linked. If one is set and the other isn't, the backend can grey out models expecting a platform key while the frontend never mounts the card to configure one. Might be worth deriving both from a single source or documenting the coupling clearly.

apps/backend/src/rhesis/backend/app/services/platform_key.py
```diff
@@ -0,0 +1,290 @@
...
+    """Store *key* (encrypted) on the organization row and cache its status.
+
+    Refresh-on-write: this is a write endpoint, so it probes the key once
+    (``validate_platform_key``) and persists the validation result alongside
+    the key. Subsequent status reads and the GET /models availability
+    annotation then read those cached columns instead of re-probing.
+    """
+    org = _load_org(db, organization_id)
+    if org is None:
+        raise ValueError("Organization not found")
```
harry-rhesis, line 191: `set_platform_api_key` raises a bare `ValueError` when the org isn't found, but `routers/platform.py` doesn't catch it, so this becomes an unhandled 500 instead of a clean 4xx.

apps/backend/src/rhesis/backend/app/utils/user_model_utils.py
```diff
@@ -591,6 +596,20 @@ def _fetch_and_configure_model(
 
     # Special handling for Rhesis system models
     if _is_rhesis_system_model(provider, api_key):
+        # Local/self-hosted mode: authenticate the prepopulated Rhesis-hosted
+        # system models with the org-scoped platform key when one is configured,
+        # accruing usage like any other hosted call. Non-local (SaaS) behavior
+        # is unchanged: fall back to the accrual-wrapped default.
+        if get_application_settings().is_local:
```
harry-rhesis: This local-mode key lookup block is repeated at lines 603, 630, and 771 with only the provider/model_type/on_usage varying. Worth factoring into one helper so a future fix doesn't need to be applied three times.

apps/backend/src/rhesis/backend/app/routers/platform.py
```diff
@@ -0,0 +1,79 @@
...
+    payload: PlatformKeyUpdate,
+    db: Session = Depends(get_db_session),
+    current_user: User = Depends(require_current_user_or_token),
+) -> PlatformKeyStatus:
+    """Store the org platform key (encrypted) and return its validated status."""
+    platform_key_service.set_platform_api_key(db, current_user.organization_id, payload.key)
+    status = platform_key_service.get_platform_key_status(db, current_user.organization_id)
```
harry-rhesis: `set_platform_api_key`/`clear_platform_api_key` already load and mutate the org row, then `get_platform_key_status` re-queries it from scratch right after. Could reuse the already-loaded object and skip the extra round trip.

apps/backend/src/rhesis/backend/app/services/platform_key.py
```diff
@@ -0,0 +1,290 @@
...
+        logger.warning("Polyphemus authorization probe failed: %s", exc)
+        return None
+    if resp.status_code in (401, 403):
+        return False
+    # A non-auth-enforcing response (e.g. a public 200 root) cannot be
+    # attributed to the Bearer token being valid -- treat as inconclusive.
+    return None
+
+
+def validate_platform_key(key: str) -> dict:
```
harry-rhesis, line 251: The primary probe and fallback probe are both blocking `httpx.get` calls with 5s timeouts, run one after the other. Worst case that's about 10 seconds holding a worker and DB session. Might be worth trimming the timeout or making these async.

apps/backend/src/rhesis/backend/app/services/platform_key.py
```diff
@@ -0,0 +1,290 @@
...
+
+# How long a cached validation result is trusted before an opt-in refresh
+# (GET /platform/rhesis-key with refresh=True) will re-probe. GET /models never
+# refreshes -- it always reads the cache.
+_CACHE_TTL_SECONDS = 600  # 10 minutes
+
+# Default hosted Polyphemus URL, mirrored from the SDK provider so the backend
+# does not need to import the SDK just to read a constant.
+_DEFAULT_POLYPHEMUS_URL = "https://polyphemus.rhesis.ai"
+
```
harry-rhesis, line 44: `_load_org` does a plain query without the soft-delete filtering that `crud.get_organization` applies elsewhere, so a soft-deleted org could still resolve a "valid" key status here.

apps/backend/src/rhesis/backend/app/services/platform_key.py
```diff
@@ -0,0 +1,290 @@
...
+    Single source of truth for "is a platform key present": prefers the
+    org-stored (encrypted) key and falls back to the process-wide
+    ``RHESIS_API_KEY`` env var. Cheap: one row read plus an env lookup, no
+    network. Returns ``None`` when neither source provides a key.
+    """
+    return _resolve_key(_load_org(db, organization_id))
+
+
+def is_platform_key_present(db: Session, organization_id) -> bool:
```
harry-rhesis: `is_platform_key_present`, `get_cached_key_valid`, and `get_cached_polyphemus_authorized` don't have any callers yet, only `get_platform_api_key` and `get_availability_signals` are used. If they're not needed for a follow-up, might be worth trimming them so the next person doesn't have to guess which one to call.

harry-rhesis (commented):
Thanks for this, @alexeyhramov123. Nice work putting the platform key service and model availability logic together, the test coverage on platform_key.py is thorough. I left a few inline comments, the two that matter most before merge are the missing owner-only check on the key endpoints and the redirect-handling gap in key validation. The rest are smaller consistency and cleanup notes.

PR #2386, opened 2026-08-06
Centralize LLM token accrual behind one sink and ambient org

### LINKED ISSUE #2382: Centralize LLM token accrual instead of wiring it per call site
## Summary

Move MODEL_TOKENS accrual from per-call-site wiring to a central mechanism, so any LLM call made through the backend is counted automatically instead of only the call sites that remembered to pass `on_usage`.

## Background / Context

Accrual today works by passing `on_usage=make_usage_accrual_callback(org_id)` into each model constructor, wired in `apps/backend/src/rhesis/backend/app/utils/user_model_utils.py`. Every construction site has to remember it. Three places had not, all found while testing the usage dashboard (PR #2355):

1. Batch and sequential test execution fell back to a bare `DEFAULT_*_MODEL` string when a test config had no resolvable user. The string only becomes a model much later, inside Penelope or the metric judge, where no org context is left. Fixed in #2355 by routing those fallbacks through `resolve_default_hosted_model`.
2. The batch runner constructs `MetricEvaluator` without `db`/`organization_id`, so per-metric `model_id` judge overrides are silently ignored in the async path (`tasks/execution/batch/runner.py`). Still open, see #2 in Deliverables.
3. `PenelopeAgent` calls `get_model()` with no callback when handed a string or `None` (`penelope/src/rhesis/penelope/agent.py`). Not reachable today, but nothing prevents it.

These are one bug three times. Patching call sites leaves the fourth to be found in production as a silently undercounted invoice.

There is a second layer with the same shape on the SDK side: `_emit_usage` is called by each provider individually. Adding emission for LiteLLM-backed providers required patching `a_generate`, `generate_batch`, and `_a_generate_stream` separately, and `polyphemus.py`/`native.py` are their own implementations. A new provider emits nothing at all, silently.

## Goals

- Emission is structurally unmissable: a new SDK provider cannot ship without it.
- Attribution is ambient: adding a new backend call path cannot silently skip accrual.
- Unattributed usage is visible rather than dropped.

## Deliverables

1. **SDK**: emit in one place. Template-method shape in `BaseLLM` where the public `generate`/`a_generate`/`generate_batch` do the emit and providers implement an inner method that only returns the response. Note this changes the SDK's provider contract, so it needs a version bump and a look at any external subclasses.
2. **Backend**: one process-wide usage sink registered at startup, plus ambient org context read at emission time, replacing the per-instance closure. Attribution gets bound in the two entry points that already bind scope (FastAPI dependency, `celery/signals.py`) instead of at N construction sites. Verify the context survives `asyncio.to_thread` and the batch executor's fan-out.
3. **Keep the hosted/not-hosted decision where it is.** This part cannot move to the provider level: whether tokens count depends on whose credentials paid, not on what the model is. An org's own OpenAI key means they pay OpenAI directly and it must not hit their Rhesis quota; the deployment default runs on our credentials and must. Same provider class, opposite answer, decided by how the model was selected. It stays at the resolution layer in `user_model_utils.py`, but becomes one flag set at one chokepoint rather than a callback threaded through every call site.
4. **An unattributed-usage counter**, incremented when a call emits with no org bound. The current explicit design fails visibly (you notice zero counts); an ambient one can silently mis-attribute, and that regression must be observable.
5. **Fix the batch `MetricEvaluator` gap** (item 2 above). This one needs its own design: there is no live DB session in the batch path by construction, since `prefetch_execution_context` expunges everything so work can run across threads and coroutines. The fix that matches the existing pattern is resolving per-metric `model_id` overrides eagerly during prefetch, while the session is open, the same way `execution_model`/`evaluation_model` already are, and storing the resolved instances on `ExecutionContext` so `prepare_metrics` skips the DB lookup at evaluation time.

- [ ] A test asserts that a provider which does not call the emit hook fails, or that emission happens for every public generate method on `BaseLLM` without per-provider wiring.
- [ ] A test constructs a model through a path that passes no `on_usage` and still sees accrual dispatched for the ambient org.
- [ ] A test asserts usage emitted with no ambient org increments the unattributed counter rather than accruing to an empty org id.
- [ ] Per-metric `model_id` judge overrides accrue in the batch path, not just the sequential one.
- [ ] An org with its own provider API key accrues zero MODEL_TOKENS, verified as a regression test on the credential-provenance rule.

## Risks / Considerations

- Silent mis-attribution is the main new failure mode, hence deliverable 4. An ambient mechanism that loses its context does not error, it just books tokens somewhere wrong or nowhere.
- The SDK change touches a public contract. External subclasses of `BaseLLM` would break if the provider-implemented method is renamed without a compatibility shim.
- Double counting if both mechanisms are live during the transition. The per-instance `on_usage` wiring needs removing in the same change that adds the sink, not after.

## Additional Context

Found while verifying MODEL_TOKENS on the usage dashboard (PR #2355). The fallback fix shipped there; this issue covers the mechanism.

harry-rhesis:
## Purpose

Closes #2382.

MODEL_TOKENS accrual worked by closing over an organization id at model-construction time and passing the resulting callback into each model constructor. Every construction site had to remember it, and sites forgot. Three were found while testing the usage dashboard in #2355; two were patched there, the mechanism was not. Exploring the rest turned up a larger population than the issue assumed: around twenty `get_model(...)` call sites in `apps/backend/` build a model outside the resolution chain with no accrual at all, and two SDK providers emit nothing whatsoever. `LiteLLMProxy` parses `usage` out of its response and discards it, so a real hosted API has reported zero tokens for its entire existence, and `HuggingFaceLLM` computes exact counts and stores them on an attribute nobody reads. Neither failed any test, because emission was a per-provider obligation documented only in docstrings.

The goal is that adding a call path or a provider cannot silently skip accrual, and that usage which cannot be attributed is visible rather than dropped.

## What Changed

**SDK: one sink instead of per-call-site wiring.** `set_default_usage_callback()` registers a process-wide sink that receives usage from every model in the process, including models built by code that has never heard of usage accounting. It fires alongside any instance `on_usage` rather than replacing it, so attaching a listener to one model cannot silently detach it from the host's accounting. `BaseLLM.usage_metered` lets a host record whose credentials paid; the SDK never sets or reads it. A conformance test walks the provider registry and fails on any new emission root without a test driving its transport. `LiteLLMProxy` and `HuggingFaceLLM` now emit, and `LiteLLM.__init__` accepts `on_usage` like every other provider.

**Backend: attribution read at emission time.** A ContextVar holds the organization to bill, bound in the two places that already establish tenant context. The FastAPI dependency has to be `async`, and there is a test asserting it stays that way: a sync dependency runs in the anyio threadpool, where a ContextVar write cannot reach the request's own task. That is exactly why `app/scope.py`'s ContextVar is documented as unusable from handlers, and `apps/backend/AGENTS.md` has been updated to say the prohibition is about sync dependencies specifically. Celery binds on `task_prerun` and unbinds on `task_postrun`, because a prefork worker reusing a process would otherwise bill the next task to the previous task's organization. `ThreadPoolExecutor` and `run_in_executor` do not copy context, so the metric strategies and the Penelope target wrap their work with `with_usage_attribution`. Each boundary has its own propagation test, including the two that do not propagate on their own.

**Whether to bill stays at the resolution layer**, as a stamp rather than a threaded callback. This part cannot move to the provider: an org running OpenAI on its own key pays OpenAI directly, while the deployment default runs on our credentials, and that is the same provider class with opposite answers. An unstamped model bills and logs `usage.unstamped_model` with the provider and model name, so a missed call site is visible rather than a silent undercount. The exception is a model holding an API key that is not ours, which is not billed. That comparison needed care in both directions: `RhesisLLM` and `PolyphemusLLM` both fall back to `os.getenv("RHESIS_API_KEY")`, so treating any key as the org's own would have silently stopped billing the main hosted provider.

**Two billing correctness fixes.** `model_connection.py` runs a real generation against credentials a user has just typed into the connection-test form; those are now stamped unmetered, since billing them would charge an org for tokens their own provider already charged them for. And an orgless-user bug that @peqy flagged in review of #2355 and that is still on `main` at `user_model_utils.py:694`: `str(user.organization_id)` yields `"None"`, a truthy string that passes every `if organization_id:` guard and only fails later when the accrual task casts it to a UUID. `resolve_default_hosted_model` no longer takes an organization id at all, and `bind_usage_org` rejects stringified nulls so a future caller cannot reintroduce it.

**Unattributed usage is counted, not dropped.** Emissions with no organization bound log `usage.unattributed` with structured fields. `JsonLogFormatter` previously discarded `extra={...}` entirely, so it now forwards an allowlist of keys, which is what makes the signal queryable rather than only greppable. Deliberately not written to the `usage` table: it has an FK to `organization.id` and an RLS policy whose `USING` doubles as the insert check, so there is no row shape for "no org", and a sentinel org would corrupt the table everyone bills from.

No log line contains an API key. There is a test that sweeps every logging branch with a key set and checks both the formatted message and the structured fields.

## Additional Context

Scoped down from a five-PR plan to one, on the basis that a process-wide sink makes the remaining bare `get_model()` sites correct-and-noisy rather than silently wrong, which turns "route twenty call sites through a chokepoint" into optional cleanup. Two items stay open on #2382: the SDK template-method refactor with `__init_subclass__` enforcement, which is a breaking contract change worth its own PR, and the batch `MetricEvaluator` per-metric override bug, which is a correctness defect rather than an emission one and whose accrual half fixes itself here.

Behaviour change worth knowing about: an unstamped model now bills the ambient org by default. That is the correct default for the actual population of unstamped sites, which are all deployment defaults running on our credentials, and the alternative default would rebuild the silent undercount this PR removes. The logging makes the remaining ones findable.

This supersedes `test_no_org_still_builds_the_model_but_wires_no_accrual`, which pinned the old behaviour of no-org meaning no accrual wiring.

## Testing

6932 backend tests and 2340 SDK tests pass. 35 of those are new, covering accrual decisions, context propagation per boundary, both binding points, API keys never reaching logs, and provider emission conformance.

Three regressions were verified to actually be caught rather than assumed:

- Deleted the `LiteLLMProxy` emit line and confirmed the conformance test fails.
- Replaced the key comparison with a naive "has a key" check and confirmed the hosted-default test fails.
- Ran the whole backend suite on a clean `main` to confirm the one remaining failure, `test_status.py::test_create_status_with_long_name`, is a pre-existing ordering flake and not from this branch.

To exercise it by hand: run a test set, then check `GET /usage` reflects MODEL_TOKENS for the org, and grep worker logs for `usage.unattributed` and `usage.unstamped_model`, both of which should be absent on the normal paths.


harry-rhesis:
Thanks, all three addressed in 9daec83bb.

**The question was a real gap, not a hypothetical.** I went looking for providers that read an env var into `self.api_key` and found `litellm_proxy.py:58`: `self.api_key = api_key or os.getenv("LITELLM_PROXY_API_KEY")`. So a deployment defaulting to that provider holds a key that is ours but does not match `RHESIS_API_KEY`, and any of the unstamped bare `get_model()` call sites on it would have read as org-owned and silently stopped billing. Reachable today, since `explorer/suggestions.py:74` builds the system default that way. The comparison is now against every key the deployment supplies. I also added a test that scans the SDK providers for that assignment pattern and fails if one reads an environment variable the list does not know about, so it cannot go stale the next time a provider is added. Verified it fails when I drop the new entry.

**Celery `task_id`:** guarded with an early return. I did not take the `id(task)` fallback, because Celery instantiates one task object per task type per worker, so two concurrent runs of the same task would share a key and reset each other tokens, which is worse than not binding. Skipping leaves the usage unattributed and logged, rather than billed to the wrong org.

**`bind_usage_org`:** now accepts a `UUID` as well as a string, which is what the column actually yields.

CI was green before these and the new tests pass locally (6936 backend, 2340 SDK). The one backend failure is `test_status.py::test_create_status_with_long_name`, a pre-existing ordering flake that I confirmed fails identically on a clean `main`.

harry-rhesis:
All three addressed in `fb48cda5d`.

**Boundary test bypass** — real gap, thanks. The guard only inspected `ast.ImportFrom`, so `import rhesis.sdk.models.factory as f` then `f.get_model(...)` walked past it. I flag the module import itself rather than the attribute access: matching `.get_model` by name would false-positive on every `crud.get_model` call, whereas importing `rhesis.sdk.models` or `rhesis.sdk.models.factory` is unambiguous and is the only way to reach the factory that way. Verified against all three bypass forms (aliased module, aliased package, plain import) and confirmed `BaseLLM`, `crud.get_model` and embedder imports still pass.

**`_warned_unstamped` thread safety** — took the "accept dupes and comment" option. The race is between the membership test and the add, and `set.add` on an already-hashed tuple is atomic under the GIL, so it cannot corrupt the set or raise; the worst case is a duplicate log line. Dedup there is noise control rather than correctness, so a lock on the logging path would cost more than it saves. Now says so in the code.

**Bill-on-unstamped** — took the comment option rather than the flag. The branch now names the invariant that makes it safe (the boundary test enforces it, so it is not an assumption) and, more usefully, records when it stops holding: we are heading towards not offering Rhesis-hosted models at all, and once every model carries an org key, "unstamped" flips from "ours" to "theirs" and skipping becomes the safer default. Worth noting the blast radius today is bounded, since quota limits are display-only with no enforcement, so a wrong value shows on a dashboard rather than blocking anyone. A feature flag felt like the wrong shape for something with a known expiry date.

6942 backend + 2340 SDK green, same single pre-existing `test_create_status_with_long_name` flake that also fails on a clean `main`.

harry-rhesis:
Closing the loop on the `CHANGES_REQUESTED` review, which I fixed but never actually replied to here — and on the refactors that followed. All review threads are now resolved.

**[Critical] Celery workers would not accrue tokens** — correct, and worse than the review suggested. Two separate bugs, fixed in `0a36f37e2`:

- `worker_process_init` is sent only by the prefork and solo pools, and `apps/worker/start.sh` runs `--pool threads` for both the main and architect workers, so the sink was never installed in any worker. Now also connected to `celeryd_init`, which fires for every pool.
- Independently: Celery's tracer fires `task_prerun` *before* `Task.before_start`, and `before_start` is what copies `organization_id` off the message onto the request. So the handler was reading an attribute that did not exist yet and every Celery task went unattributed regardless of pool. Now resolved from the message, mirroring `before_start`'s own precedence so retries agree.

Both had passing tests beforehand, because those tests hand-built a request with the attribute already set — they asserted my assumption rather than Celery's behaviour. Replaced with one that drives a real task through the tracer in eager mode, plus one that sends the signals and asserts the sink is installed. Verified both fail against the old code.

A related audit found two more thread hops dropping context (`services/generation.py` used `run_in_executor`, which unlike `asyncio.to_thread` does not copy contextvars) — `4cdd18195`.

**`with_usage_attribution` wrap-time capture** — doc warning added in `b6edc694a`, pointing at the test that pins the semantics.

**`_our_api_keys()` recomputed per emission** — resolved by deletion in `9645d2380` rather than by caching. Tracing every path that can carry an org-supplied key (Model row, connection-test form) showed all of them stamp, so the heuristic was asking whether a key was foreign about a model that structurally cannot hold one. Removing it also took out the `_DEPLOYMENT_KEY_ENV_VARS` list and the test that regex-parsed SDK provider source to keep it fresh — a guard with its own silent failure mode. The one genuinely unstamped path (Penelope building its own default) is now stamped at both call sites in `a3ec7d817`.

**SDK global mutable state** — `bb71eb4cd`. Took the "mention it in the docs" option: the docstring now says what follows from process-wide scope (last caller wins, so it is for an application at startup, not a library or handler; concurrent per-tenant sinks want per-instance `on_usage`; tests must uninstall). Skipped the contextmanager helper as speculative API surface.

Also in this range, from my own review rather than yours: `4dee70ff8` consolidated ~10 string→model unwrap sites into `ensure_language_model` behind the boundary test, and `df9e209ed` deduped `resolve_default_hosted_model` against it, removing a double-construction on the failure path.

PR #2432, opened 2026-08-11
Prevent Architect tool-call failures on the first attempt

harry-rhesis:
## Purpose

The Architect fails a tool call, retries, fails again, and eventually succeeds. The task completes, but every failed attempt burns one of the 15 iterations allowed per turn and shows the user a red marker.

One theme is behind almost all of it: **the model writes its arguments from the rendered tool schema, and that schema was consistently vaguer than what the endpoint actually enforces.** Where the schema under-specified, the model guessed; the guess was wrong; the server rejected it.

A concrete case from a real session. The user asked for a multi-turn test set about insurance fraud. `generate_test_set` was rejected with "At least one behavior must be specified" (`routers/test_set.py:132`, which fires when `config.behaviors` is empty). The model had been shown `config: object` with the requirement buried in a prose bullet list, while the schema itself declared nothing required. The agent then told the user this was a platform bug — "the system registry takes a few moments to propagate the new behavior-metric linkages" — which is invented; that endpoint reads `config.behaviors` straight off the request body and never queries the database. It then worked around the failure by switching to `create_test_set_bulk`, which only stores prompts the model writes out by hand. The user asked for a generated multi-turn set and received a single hand-written scenario, announced as "Good news!"

Scope is prevention. How the remaining failures are displayed is deliberately out of scope.

## What Changed

**Make the schema tell the truth**

- `_format_tools` recurses into nested objects and array items, so `tests: array[object]` now shows that every item requires `behavior`, `category` and `topic`. Costs +830 chars across the whole 48-tool catalog (+1.7%), because most `object` params are `Dict[str, Any]` with nothing to expand.
- `generate_test_set` resolves `project_id` from the request's project scope via the existing `get_project_context` dependency. The Architect already sends `X-Project-Id`, so it no longer has to supply a field nothing told it about. Fewer required fields beats better docs.
- `name` is required on a `TestSetGenerationRequest` subclass used only by the bulk route, and `config.behaviors` is required with `min_length=1` — both matching what the routers already enforced.
- `test_set_type` and `test_type` are typed as their enums, so `"Single-Turn" | "Multi-Turn"` reaches the schema instead of a bare `string`. The `mode="before"` normalizers are kept, so loose casing and snake_case still pass exactly as before.
- `create_metric`'s conditional requirements are documented: numeric metrics need `min_score`, `max_score` and `threshold`; categorical ones need at least two `categories` and one `passing_categories`. Flat JSON Schema cannot express this, and it was stated nowhere, making a numeric metric a guaranteed first-attempt 422. Metrics get created before test sets in every plan.
- Behaviors and metrics gain duplicate-name guidance (resolve and reuse, never retry with a suffixed name), and the undescribed parameters on the creation tools are described.

**Let the model learn from its own failure**

- `_format_history` wrote `Called: <tool_name>` and nothing else, so on a rejection the model saw the error but not the payload. It now echoes a truncated argument preview for failed calls only.
- The iteration prompt says what to do with a failure: fix the named argument, never re-send an identical payload, treat an ordering or already-exists message as an instruction rather than something to retry.
- Three new rules stop the workaround behaviour: never attribute a failure to a sync delay or platform quirk you cannot see, never substitute a different tool because one failed (`create_test_set_bulk` called out by name), never present a reduced result as success.

**Stop losing and re-failing payloads**

- `_parse_arguments` silently returned `{}` when the model's JSON-string arguments failed to parse, so the server replied "field required" for fields the model believed it had sent. It now logs, and coerces a valid-but-non-object result to `{}` instead of letting it kill the whole `AgentAction`.
- Arguments are checked against the tool's own `inputSchema` before dispatch, naming the missing field without an HTTP round trip. Deliberately narrow — only missing required fields. Types and enum values are left alone because the API normalises much of what looks wrong here, and rejecting those locally would invent failures the server would have accepted.
- A 5xx or dropped connection is retried once instead of becoming a permanent failure that costs an iteration.

## Additional Context

- Follow-up to #2428, which covered the turn-type guidance and left enforcement out.
- Two things found and corrected mid-change: the recursion initially hid `sources[].id` because `id` is in the server-managed field list, so that filter now applies only at the top level; and a `min_length=1` on the bulk `tests` array was reverted after finding that `sdk/src/rhesis/sdk/entities/test_set.py:983` deliberately pushes empty lists, with integration coverage.
- Breaking-ish: `GenerationConfig.behaviors` and `TestSetGenerationRequest.name` now fail as 422 rather than 400. Both were already rejected; only the status code changes. The frontend sends both in every flow and surfaces either identically.
- The new `test_mcp_tools_yaml.py` coverage test is scoped to the creation path. `update_*` and endpoint config still carry ~60 undescribed parameters; widen the set as they are cleaned up rather than relaxing the rule.

## Testing

```bash
cd sdk && make test                    # 2354 passed
uvx ruff check && uvx ruff format --check   # clean on all touched files
```

**Backend tests have not been run — Docker was unavailable in this environment.** That leaves the `project_id` fallback, the enum-typed schemas and the new `test_mcp_tools_yaml.py` assertions unverified by the suite. I checked every one of those assertions against the real generated tool catalog using a stub app, and all pass, but that is not the same as running them. Please run:

```bash
cd apps/backend
uv run pytest ../../tests/backend/services/test_mcp_tools_yaml.py -v
uv run pytest ../../tests/backend/routes/test_test_set.py ../../tests/backend/services/test_test_set.py -v
```

Manual check — the original failing prompt:

1. `./rh dev up`, open the Architect
2. "let's create a multi-turn test set about insurance fraud", pick one behavior, approve
3. `generate_test_set` should now succeed on the first call. It should never fall back to `create_test_set_bulk`, and never describe a failure as a platform sync issue.

### DIFF
--- a/apps/backend/src/rhesis/backend/app/mcp_server/local_tools.py
+++ b/apps/backend/src/rhesis/backend/app/mcp_server/local_tools.py
@@ -9,9 +9,10 @@ Auth is handled via a delegation token passed as a Bearer header,
 so tenant isolation is preserved without requiring a static API key.
 """
 
+import asyncio
 import json
 import logging
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Optional
 
 import httpx
 
@@ -27,6 +28,11 @@ from .tools import (
 )
 
 logger = logging.getLogger(__name__)
+
+# One retry is enough for the blips this is aimed at. More would stall the
+# turn behind a genuinely broken dependency.
+_TRANSIENT_RETRIES = 1
+_RETRY_DELAY_SECONDS = 0.5
 
 
 class LocalToolProvider(MCPTool):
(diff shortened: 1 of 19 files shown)


harry-rhesis:
Both valid, both fixed in `dfc62ae` and `c086e0d`.

**1. Redundant router checks on `POST /test_sets/generate`**

Right on all counts, including the whitespace-only detail — that was the one case still reaching the router, and exactly the drift you described. Took the first option rather than aligning to 422:

- `TestSetGenerationRequest.name` is now `Annotated[str, StringConstraints(strip_whitespace=True, min_length=1)]`, so the strip happens before the length check and a blank name is a 422 like every other validation failure.
- Dropped both router checks. The handler just reads `request.name`.

I also removed the same `config.behaviors` check in `routers/services.py`. You only flagged `/test_sets/generate`, but the `GenerationConfig.behaviors` change made it unreachable on the sampling route too, so leaving it would have been the same inconsistency one file over.

Verified: `"  Fraud Scenarios  "` → `'Fraud Scenarios'`; whitespace-only, empty name, and empty `behaviors` all 422; the sampling route's optional `name` unchanged.

One consequence worth recording: a missing `config.behaviors` is now a 422 rather than a 400 with a hand-written message. Still actionable for the agent — the SDK's pre-dispatch check catches the *absent* case before any HTTP call and names `config.behaviors` directly, and pydantic names the field for the *empty-list* case, which that check deliberately leaves to the server.

**2. Backend test importing from the SDK**

Agreed, and it was a bit worse than the coupling itself — it reached for a private name (`_SERVER_MANAGED_FIELDS`) across a package boundary. Now a local `SERVER_MANAGED` constant in the test class, with a comment pointing at the SDK as the source of truth.

I deliberately skipped a drift guard, since asserting the two match would reintroduce the import. The duplication fails in the safe direction: if the SDK *adds* a field, this test loudly demands a description for a parameter the agent never sees. Only a *removal* would open a silent gap, which is the rarer direction.

PR #2445, opened 2026-08-11
Fix: Duplicate Env Between Config Secrets

### LINKED ISSUE #2420: Dedupe env var keys between Helm ConfigMap and ExternalSecret
## Summary

Several env var keys are declared in **both** the Helm ConfigMap and the cluster `ExternalSecret`.
Every deployment lists `secretRef` after `configMapRef` in `envFrom`, and later sources win in
Kubernetes, so the Secret silently overrides the chart. Editing those keys in `values-*.yaml` is a
no-op with no warning.

## Background / Context

Found while fixing the Vertex embedding 404 (`locations/eu` does not serve `text-embedding-005`).
The fix sets a region in `values-*.yaml`, which would have done nothing: `VERTEX_AI_LOCATION` was
also projected from `prd-rhesis-vertex-ai-location`.

#2419 removes `VERTEX_AI_LOCATION` and `VERTEX_AI_PROJECT` from all three clusters'
`ExternalSecret`. That pair was safe to remove because the ConfigMap value provably matched what
production was actually using (the 404 reported `projects/playground-437609/locations/eu/...`,
identical to the chart). The remaining keys could not be verified without reading Secret Manager, so
they were left in place.

Affected `envFrom` order (all 7 deployments): `charts/rhesis/templates/{backend,worker,frontend,chatbot,polyphemus,otel-collector,telemetry-processor}/deployment.yaml`

## Remaining duplicated keys

| env | keys in both ConfigMap and ExternalSecret |
| --- | --- |
| dev | `APP_DB_USER`, `DB_HOST`, `DB_NAME`, `OTEL_PROCESSOR_ENDPOINT` |
| stg | `ANALYTICS_DB_HOST`, `ANALYTICS_DB_NAME`, `ANALYTICS_DB_PORT`, `OTEL_PROCESSOR_ENDPOINT` |
| prd | `ANALYTICS_DB_HOST`, `ANALYTICS_DB_NAME`, `ANALYTICS_DB_PORT`, `OTEL_PROCESSOR_ENDPOINT` |

Reproduce:

```bash
env=prd
helm template rhesis charts/rhesis -f charts/rhesis/values-$env.yaml \
  | awk '/kind: ConfigMap/,/^---/' \
  | grep -oE '^  [A-Z][A-Z0-9_]+:' | tr -d ' :' | sort -u > /tmp/cm.txt
grep -oE 'secretKey: [A-Z][A-Z0-9_]+' \
  kubernetes/clusters/$env/external-secrets/rhesis-app-secrets.yaml \
  | awk '{print $2}' | sort -u > /tmp/sec.txt
comm -12 /tmp/cm.txt /tmp/sec.txt
```

## Goals

- `values-*.yaml` is the single source of truth for non-secret config.
- A new overlap cannot be introduced without CI failing.

## Steps / Action Plan

1. For each remaining key, read the Secret Manager value and compare it to the chart value.
2. Where they match, delete the key from the `ExternalSecret` (no behaviour change).
3. Where they differ, decide which is authoritative, align the chart, then delete the duplicate.
4. Add a lint check that intersects rendered ConfigMap keys with `ExternalSecret` `secretKey`s and
   fails on any overlap. `.github/workflows/lint.yml` is the natural home.

- [ ] No key appears in both the rendered ConfigMap and the `ExternalSecret`, for dev, stg and prd.
- [ ] Secrets stay in the `ExternalSecret`; non-secret config lives only in the ConfigMap.
- [ ] CI fails when a duplicate key is introduced.
- [ ] The reproduce command above prints nothing for all three envs.

## Risks / Considerations

`DB_HOST`, `DB_NAME`, `APP_DB_USER` and the `ANALYTICS_DB_*` keys point at real databases. If the
Secret value differs from the chart value, removing the Secret key repoints a live connection. Verify
each value before deleting anything, and do not batch this with unrelated changes.

asad-miah:
This PR introduces changes from the `fix/duplicate-env-between-config-secrets` branch.

## 📁 Files Changed (       5 files)

```
.github/workflows/lint.yml
kubernetes/clusters/dev/external-secrets/rhesis-app-secrets.yaml
kubernetes/clusters/prd/external-secrets/rhesis-app-secrets.yaml
kubernetes/clusters/stg/external-secrets/rhesis-app-secrets.yaml
scripts/check-k8s-secret-overlap.sh
```

## 📋 Commit Details

```
46affb65b - chore: remove unused RHESIS_ADMIN/USER/ANALYTICS_USER passwords (Md Asaduzzaman Miah, 2026-08-11 16:07)
1a471f43d - fix(dev): remove keys duplicated between ConfigMap and ExternalSecret (Md Asaduzzaman Miah, 2026-08-11 16:04)
```

## ✅ Checklist

- [ ] All tests pass locally


scripts/check-k8s-secret-overlap.sh
```diff
@@ -0,0 +1,46 @@
...
+  secrets_file="kubernetes/clusters/$env/external-secrets/rhesis-app-secrets.yaml"
+  # --show-only isolates the chart's own ConfigMap template, not the raw combined
+  # output — a plain `awk '/kind: ConfigMap/,/^---/'` would also catch subchart
+  # ConfigMaps (postgresql/valkey/otel), whichever renders first.
+  cm_keys=$(
+    helm template rhesis charts/rhesis -f "charts/rhesis/values-$env.yaml" \
+      --show-only templates/configmap.yaml \
+      | grep -oE '^  [A-Z][A-Z0-9_]+:' | tr -d ' :' | sort -u
+  )
+  secret_keys=$(
```
harry-rhesis, line 29: Minor robustness concern: `grep -oE` returns exit code 1 when it finds no matches. With `set -euo pipefail`, if a secrets file had no `secretKey:` entries matching the pattern (unlikely but possible during a refactor), the script would exit before reaching the overlap check, rather than reporting "OK."

Same applies to the `cm_keys` pipeline on line 24 if the ConfigMap template somehow rendered with no matching keys.

A `|| true` guard would make this resilient:
```bash
secret_keys=$(
  grep -oE 'secretKey: [A-Z][A-Z0-9_]+' "$secrets_file" \
    | awk '{print $2}' | sort -u
) || true
```

kubernetes/clusters/dev/external-secrets/rhesis-app-secrets.yaml
```diff
@@ -93,13 +84,6 @@ spec:
       remoteRef:
         key: dev-rhesis-session-secret-key
 
-    - secretKey: RHESIS_ADMIN_PASSWORD
```
harry-rhesis, line 96: These three password keys (`RHESIS_ADMIN_PASSWORD`, `RHESIS_USER_PASSWORD`, `ANALYTICS_USER_PASSWORD`) are removed across all three envs as "unused." Could you confirm how that was verified? A grep across the Helm templates, deployment specs, and app code showing nothing references them would be enough. Just want to make sure no workload silently loses an env var it reads.

scripts/check-k8s-secret-overlap.sh
```diff
@@ -0,0 +1,46 @@
...
+
+FAILED=0
+
+for env in dev stg prd; do
+  secrets_file="kubernetes/clusters/$env/external-secrets/rhesis-app-secrets.yaml"
+  # --show-only isolates the chart's own ConfigMap template, not the raw combined
+  # output — a plain `awk '/kind: ConfigMap/,/^---/'` would also catch subchart
+  # ConfigMaps (postgresql/valkey/otel), whichever renders first.
+  cm_keys=$(
+    helm template rhesis charts/rhesis -f "charts/rhesis/values-$env.yaml" \
```
harry-rhesis, line 25: Worth noting: `helm template` needs the subchart tarballs in `charts/rhesis/charts/`. They're committed today, so this works. If the project ever `.gitignore`s them (common pattern), the CI job would need a `helm dependency build` step before this line. Not a problem now, just a future-proofing note.

harry-rhesis (commented):
Good catch on the silent-override problem, Asad. The fix is clean and the CI guardrail is a nice addition. A few things below.

**What looks good:**
- Fixes a real, subtle bug: `secretRef` after `configMapRef` in `envFrom` means the Secret silently wins, making Helm values edits no-ops with no warning.
- Clean commit separation: config-overlap fix in one commit, unused password cleanup in another.
- Good "deliberately absent" comments explaining *why* keys were removed, so the next person doesn't re-add them.
- The CI script uses `--show-only templates/configmap.yaml` to avoid matching subchart ConfigMaps. That detail matters.
- `set -euo pipefail` from the start.

scripts/check-k8s-secret-overlap.sh
asad-miah: Good catch, fixed in 65e2725f0 — added `|| true` to both the `cm_keys` and `secret_keys` pipelines so a zero-match grep can't kill the script under `set -euo pipefail`.

kubernetes/clusters/dev/external-secrets/rhesis-app-secrets.yaml
asad-miah: Verified by grepping the whole repo (Helm chart templates/values, backend, frontend, SDK, infrastructure/, docker-compose, .env.example, gsm-secrets-sync.sh, service-secrets-* files, init scripts) for RHESIS_ADMIN_PASSWORD, RHESIS_USER_PASSWORD, and ANALYTICS_USER_PASSWORD — zero references anywhere outside these ExternalSecret files.

The actual DB role bootstrap doesn't touch these three keys at all:
- Dev: the Bitnami postgres subchart's init container (charts/rhesis/values-dev.yaml) creates the rhesis-admin/rhesis-user/rhesis-analytics-user roles from ADMIN_DB_PASS/APP_DB_PASS/ANALYTICS_DB_PASS.
- Stg/prd: CNPG's declarative role management pulls its own credentials from a separate ExternalSecret, kubernetes/clusters/{stg,prd}/external-secrets/cnpg-role-credentials.yaml, which projects the same underlying GCP Secret Manager secrets under different key names into a different target Secret. That file is untouched by this PR.

So the GCP secrets themselves stay live and in use for stg/prd — just not through rhesis-app-secrets. Removing these three keys from rhesis-app-secrets is dead wiring, not a live credential.

scripts/check-k8s-secret-overlap.sh
asad-miah: Good note, thanks. Tarballs are committed today so this holds, but I'll keep it in mind if we ever move to .gitignore-ing charts/rhesis/charts/ — would need a helm dependency build step in CI before this line.

PR #2464, opened 2026-08-12
Add per-page social preview images to the docs site

nicolai-rhesis:
## Purpose

Every docs page shipped the same `og:image` — the static `GH_Short_Demo.png` video thumbnail — because `getOpenGraphImage()` took a page path and ignored it. A link to Endpoints, Core Concepts or any of the 76 glossary terms produced an identical card in Slack, LinkedIn and search previews.

## What Changed

- **`/api/og` renders a card per page.** Title, description and section come from the page's own MDX, resolved through the same content scanner the sitemap and llms.txt already use — so adding a page needs no frontmatter and a renamed heading updates the card automatically. All 273 content pages resolve.
- **Card design follows the marketing OG images**: white canvas with the blue and warm blooms, mark plus wordmark, section pill, headline that steps 76→46px as titles grow, description, `docs.rhesis.ai` footer. Colours are literal hex in `lib/og-theme.js` copied from `tokens.css`, because satori never sees a stylesheet.
- **Escape hatches.** A page can set `ogImage: /path.png` in frontmatter for a hand-made image, and any render failure redirects to the old static image rather than serving a 500 to a crawler.
- **Fonts.** satori cannot decode woff2 and renders variable fonts at their default instance, so the card uses four committed TTF subsets (~93 KB total) built by `docs/scripts/generate-og-fonts.py`. A test asserts they exist and are real TTFs — losing one would silently turn every card back into the fallback.
- **Two description bugs fixed on the way** (separate commit). Both already affected `<meta name="description">` and llms.txt, not just the cards: generated glossary pages leaked their `export const metadata = {…}` block into the description, and lead paragraphs wrapped in `**bold**` shipped the asterisks.
- **`npm test` in `docs/src` runs again** — `node --test <dir>` no longer accepts a directory, so the whole docs suite errored out instead of running.

## Additional Context

- Deliberately not per-page static files or page screenshots: a docs page shrunk to 1200×630 is unreadable, and 273 hand-made PNGs go stale the moment a page is renamed.
- `OG_VERSION` in `lib/og-theme.js` is the cache-buster. Crawlers cache social images by URL, so any design change needs a bump or LinkedIn and Slack keep serving the old picture.
- `next/og` loads a resvg wasm blob the tracer misses, hence the new `outputFileTracingIncludes` — without it the route works in dev and 500s in the container.
- Rules for all of the above are written up under "Social preview images" in `docs/AGENTS.md`.

## Testing

- `cd docs/src && npm test` — 53 pass, including per-page resolution across all 273 pages, the font/asset assertions, brand-suffix stripping, the glossary description regression, and path-traversal rejection.
- `npm run lint`, `npm run format:check`, `npm run lint:images` clean.
- `npm run build` clean; `/api/og` builds as a dynamic route and the wasm lands in `.next/standalone`.
- Ran the standalone server the way the container does (`node .next/standalone/.../server.js`) and rendered cards from it — that is the path that usually breaks in Docker.
- Spot-check locally: `npm run dev` then open `/api/og?p=docs/endpoints&v=1`, `/api/og?p=glossary/llm-as-a-judge&v=1`, `/api/og?v=1`, and a bogus path to see the fallback card.
- After deploy, re-scrape one URL through LinkedIn's Post Inspector to flush its cache.


nicolai-rhesis:
Self-review found four issues, all now fixed and pushed.

**The fallback never fired.** `ImageResponse` renders inside its body stream, so satori and resvg errors escape the `try/catch` around the constructor entirely. Verified by corrupting `Sora-700.ttf`: the route returned **500** with a broken body, not the documented fallback. The response is now buffered with `arrayBuffer()` before returning, which is what makes the catch reachable — same test now returns 302 to the static image. Per @peqy's note, that redirect is `Cache-Control: no-store` so a transient failure can't be cached as permanent.

**Card URLs went stale for up to a year.** They keyed only on path plus `OG_VERSION` but responded with `s-maxage=31536000`, so editing a page's title or description left the old card in the CDN and in every social crawler — contradicting the whole point of resolving from MDX. URLs now carry `h=`, a hash of the card text, so an edit changes the URL on its own. Tested for collisions across all 273 pages.

**11 glossary descriptions were truncated.** The exported-metadata regex cut at the first backslash, so definitions containing an apostrophe (`'the AI\'s certainty'`) ended mid-sentence with a stray backslash — on the card and in llms.txt. Now escape-aware and bounded to the metadata object, so a `description:` in a later code sample can't be picked up either.

**Emphasis unwrapping mispaired asterisks.** Stripping inline code first left `**` around a hole, so `docs/metrics/metric-scope` read `Every metric declares **. Every test set declares .` — words dropped, literal asterisks leaked. Inline code now keeps its text.

Also took @peqy's suggestion to tighten the `ogImage` absolute-URL check to `/^https?:\/\//`, and documented that frontmatter `description:` is single-line only (no page uses a block scalar; one would fall through to the paragraph scan rather than break).

58 tests pass, with regressions covering each of the above.

PR #2468, opened 2026-08-13
Add configurable brand colour, icon and product name

nicolai-rhesis:
## Purpose

White-label deployments need their own colour, icon and product name without a per-customer image build. This adds three environment variables that the frontend reads at request time, so the same image serves every deployment.

They are deliberately **not** `NEXT_PUBLIC_*`: that prefix inlines a value into the client bundle at build time, which would mean one Docker image per deployment. This follows the rule already stated in `next.config.ts` — "one Docker image works across all environments without build args" — and rides the existing `window.__ENV__` runtime-config mechanism.

They also do **not** go in GCP Secret Manager. None of the three is a secret; all three end up in the served HTML. They live in the chart's ConfigMap, which every app deployment already loads via `envFrom`, so a self-hosted deployment sets them in its own `values.yaml`.

| Variable | Example | Default |
|---|---|---|
| `BRAND_PRIMARY_COLOR` | `#6A1B9A` | *(unset — Rhesis blue)* |
| `BRAND_FAVICON_URL` | `https://cdn.example.com/favicon.png` | `/logos/rhesis-logo-favicon.svg` |
| `BRAND_PRODUCT_NAME` | `Acme` | `Rhesis AI` |

## What Changed

- **`config/branding.ts`** — validates and resolves all three, falling back to the Rhesis default with a warning. A malformed value cannot break a deployment; `javascript:` and plain `http://` favicon URLs are rejected outright.
- **`styles/brand-palette.ts`** — derives the full palette from the one hex. The lighten/darken factors are fitted to the existing Figma palette (`darken(0.26)` on `#0080AF` lands one unit off the designed `#005F82`), so a custom colour inherits the relationships the palette was built with. Button label text is chosen by contrast ratio, so a pale brand colour gets dark labels instead of unreadable white ones.
- **`styles/theme.ts`** — `getDesignTokens(mode, brandColor?)`. Also routes the ~10 component overrides that set brand colours directly (app bar, contained/outlined/text primary buttons, checkboxes, pill toggles, input focus border) through the derived values; without that they would have stayed blue while the palette changed.
- **Favicon** — root layout's static `metadata` became `generateMetadata()` so the icon resolves per request.
- **Titles** — `%s | Rhesis AI` becomes `%s | <product name>`. A configured name also replaces the Rhesis tagline in the page description.
- **Brand mark** — new `BrandMark` component used by the sidebar (collapsed and expanded), onboarding sidebar/step header, and the sign-in header.
- **Auth pages** — accent tokens, header wordmark, brand mark and footer follow the branding. A rebranded wordmark stops linking to rhesis.ai.
- **Chart / compose / docs** — `config.brandPrimaryColor`, `config.brandFaviconUrl`, `config.brandProductName` into the ConfigMap; `.env.example` and both deployment docs pages.

## Additional Context

**The default Rhesis theme is unchanged.** Passing no brand colour returns the Figma literals verbatim — there is a test asserting the palette is identical, including that `alpha('#0080AF', 0.04)` reproduces the original `rgba(0, 128, 175, 0.04)` string character for character.

**Deliberately left on Rhesis identity**, all noted in the docs:

- Chart palettes — a readable series palette needs distinguishable hues, not tints of one colour.
- The sign-in background artwork — light mode's is a raster PNG that can't be recoloured in code.
- The Documentation, Blog and GitHub links in the sign-in header — removing navigation is a product decision, not a theming one.

**`BrandMark` uses a plain `` for remote URLs**, not `next/image`. The optimizer refuses any host absent from `images.remotePatterns`, and the host here comes from a values file — unknowable at build time. Widening `remotePatterns` to `**` would turn the optimizer into an open image proxy. Favicons are a few KB, so there is nothing to optimise.

**Two things worth a reviewer's eye:**

1. The white-vs-dark label decision uses MUI's default `contrastThreshold` of 3. A brand colour whose contrast against white lands near 3:1 — mid-bright oranges, ambers and mid greens — therefore gets white labels by a narrow margin, where dark labels would score much better. The threshold stays at 3 because raising it to 4.5 would flip Rhesis's own `#0080AF` (4.47:1) to dark text, a regression on the default theme. Forcing one or the other per deployment would need a follow-up variable.
2. `helm` was not available locally, so `helm template` was never run against the ConfigMap change. The two lines follow the exact form of their neighbours and the values keys exist, but it is unverified.

Branding is per-deployment, not per-organisation — on multi-tenant rhesis.ai every org would see the same colour. That matches the Kubernetes-deployment framing of the request; org-scoped branding would be a different change built on `organization.logo_url`.

## Testing

`npx tsc --noEmit` clean, no new lint warnings, **220 suites / 2018 tests pass** (35 new).

Verified against a running dev server with all three variables set — served HTML carried the configured `rel="icon"` href, the brand colour in the CSS with zero Rhesis blue remaining, the configured `<title>`, and the branded `window.__ENV__` payload. With the variables unset the same page served the Rhesis favicon and blue, and no `brandPrimaryColor` key.

To try it locally, add to `apps/frontend/.env.local` and restart the frontend:

```
BRAND_PRIMARY_COLOR=#6A1B9A
BRAND_FAVICON_URL=https://cdn.example.com/favicon.png
BRAND_PRODUCT_NAME=Acme
```

Check the sign-in page (accents, header wordmark, footer, tab icon), then log in for the sidebar brand mark, app bar and buttons. Remove the lines and confirm nothing about the Rhesis look changed.

**Not verified end-to-end:** the `<page> | <product name>` title suffix on an inner page, and the sidebar brand mark in the running app — both need an authenticated session, which the local backend wasn't up for. The template string and the sidebar `src` are both covered by unit tests and read straight from the same resolved values as the verified pieces.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -44,9 +44,9 @@ SESSION_SECRET_KEY=
 # URL or a root-relative path, and also becomes the sidebar brand mark.
 # BRAND_PRODUCT_NAME replaces "Rhesis AI" in page titles (max 60 characters).
 
-# BRAND_PRIMARY_COLOR=#005B33
+# BRAND_PRIMARY_COLOR=#6A1B9A
 # BRAND_FAVICON_URL=https://www.example.com/favicon.png
-# BRAND_PRODUCT_NAME=Netgo
+# BRAND_PRODUCT_NAME=Acme
 
 # =============================================================================
 # DATABASE
--- a/.env.example
+++ b/.env.example
@@ -35,6 +35,20 @@ SESSION_SECRET_KEY=
 # BACKEND_PORT=8080
 
 # =============================================================================
+# BRANDING
+# =============================================================================
+# Optional. Rebrands the frontend without rebuilding the image — both are read
+# at request time. Leave unset for the standard Rhesis look; a malformed value
+# is ignored with a warning.
+# BRAND_PRIMARY_COLOR must be 6-digit hex. BRAND_FAVICON_URL must be an https://
+# URL or a root-relative path, and also becomes the sidebar brand mark.
+# BRAND_PRODUCT_NAME replaces "Rhesis AI" in page titles (max 60 characters).
+
+# BRAND_PRIMARY_COLOR=#005B33
+# BRAND_FAVICON_URL=https://www.example.com/favicon.png
+# BRAND_PRODUCT_NAME=Netgo
+
+# =============================================================================
 # DATABASE
 # =============================================================================
 
--- a/apps/frontend/src/app/(protected)/onboarding/components/OnboardingSidebar.tsx
+++ b/apps/frontend/src/app/(protected)/onboarding/components/OnboardingSidebar.tsx
@@ -1,8 +1,9 @@
 'use client';
 
-import Image from 'next/image';
 import ArrowBackIcon from '@mui/icons-material/ArrowBack';
 import { Box, Typography } from '@mui/material';
+import BrandMark from '@/components/common/BrandMark';
+import { useNavigationItems } from '@/contexts/NavigationItemsContext';
 import { DEFAULT_AUTHENTICATED_PATH } from '@/constants/paths';
 import { ONBOARDING_STEPS } from './onboarding-steps';
 import { onboardingSidebarSx } from './onboarding-styles';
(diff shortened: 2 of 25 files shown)


nicolai-rhesis:
Superseded by #2469, which has the same content on a clean branch.

PR #2348, opened 2026-08-03
Feat/Grafana Dashboard For Rhesis

asad-miah:
This PR introduces changes from the `feat/grafana-dashboard-for-rhesis` branch.

## 📁 Files Changed (      14 files)

```
kubernetes/clusters/prd/cnpg-cluster/cnpg-cluster.yaml
kubernetes/clusters/prd/external-secrets/grafana-viewer-credentials.yaml
kubernetes/clusters/prd/external-secrets/kustomization.yaml
kubernetes/clusters/prd/external-secrets/rhesis-app-secrets.yaml
kubernetes/clusters/prd/grafana-resources/grafana-viewer-credentials.yaml
kubernetes/clusters/prd/grafana-resources/kustomization.yaml
kubernetes/clusters/prd/grafana-resources/postgres-prd-datasource.yaml
kubernetes/clusters/stg/cnpg-cluster/cnpg-cluster.yaml
kubernetes/clusters/stg/external-secrets/grafana-viewer-credentials.yaml
kubernetes/clusters/stg/external-secrets/kustomization.yaml
kubernetes/clusters/stg/external-secrets/rhesis-app-secrets.yaml
kubernetes/clusters/stg/grafana-resources/grafana-viewer-credentials.yaml
kubernetes/clusters/stg/grafana-resources/kustomization.yaml
kubernetes/clusters/stg/grafana-resources/postgres-stg-datasource.yaml
```

## 📋 Commit Details

```
f779aea10 - update stg grafana pass (Md Asaduzzaman Miah, 2026-08-03 17:16)
748040115 - feat(infra): add read-only postgres datasource for grafana (stg) (Md Asaduzzaman Miah, 2026-08-03 16:23)
774d605bc - added grafana viewr password (Md Asaduzzaman Miah, 2026-08-03 14:59)
015bd809e - feat(infra): add read-only postgres datasource for grafana (prd) (Md Asaduzzaman Miah, 2026-08-03 14:22)
```

## ✅ Checklist

- [ ] Code follows the project's style guidelines
- [ ] Self-review of code has been performed
- [ ] Code is commented, particularly in hard-to-understand areas
- [ ] Corresponding changes to documentation have been made
- [ ] Tests have been added/updated for new functionality
- [ ] All tests pass locally

PR #2476, opened 2026-08-13
Add cross-org usage dashboard to Grafana

harry-rhesis:
## Purpose

Usage accrues per org into the `usage` table, but the only way to look at it is the in-app Usage page, which shows one org its own numbers. There is no way for us to see consumption across every tenant: who is heaviest, what the platform total is, whether a month looks abnormal. This adds that view to the Grafana we already run, on top of the read-only Postgres datasource from #2348.

This is the Observability slice of the usage-limits work, redesigned. The original plan called for `prometheus_client`, a `/metrics` endpoint, in-process gauges updated on increment, and a Celery beat task to resync them after restarts. That does not fit this codebase, for four independent reasons: `/metrics` is already taken by `routers/metric.py`; there is no Celery beat process anywhere, so one would mean a new container and deployment; the Celery worker never calls `bootstrap_ee`, so a beat task would resolve community limits for every org rather than their real ones; and the backend runs several replicas, so gauges updated on increment would each hold a different subset of the state and Prometheus would scrape disagreeing values. Alertmanager is also disabled, so the planned PrometheusRule thresholds would have fired into nothing.

Current usage is database state, not process state. Reading it with SQL is both simpler and more accurate, needs no new dependency and no backend code, and sidesteps the unbounded per-organization label cardinality the metrics approach would have introduced.

## What Changed

- **`kubernetes/base/grafana-resources/usage-dashboard.yaml`** (new) — the repo's first `GrafanaDashboard`. Six panels plus a notes panel: platform totals and a per-org breakdown for the current billing period, top consumers, a 12-month trend by resource and by org, and a month-over-month table.
- **`kubernetes/clusters/dev/grafana-resources/`** (new datasource + credentials) — #2348 covered stg and prd only, so dev had nothing for the dashboard to query.
- **`charts/rhesis/values-dev.yaml`** — an `init-grafana-viewer.sh` initdb script creating the `grafana-viewer` role. Dev has no CNPG, so it cannot come from `managed.roles`; this mirrors how `rhesis-user` and `rhesis-analytics-user` are already created there.
- **`infrastructure/config/gsm-secrets.json.example`** — added the `dev` entry for `GRAFANA_VIEWER_PASSWORD`, which stg and prd already had.

No backend, frontend, SDK or EE changes. No migration, no new dependency.

Two design notes worth reviewing:

The dashboard lives in `base/` even though each env names its datasource differently, because the datasource is chosen by a template variable rather than hardcoded. That variable queries type `postgres`, which is what all three `GrafanaDatasource` CRs declare, so it resolves whether or not the running Grafana maps the newer `grafana-postgresql-datasource` plugin alias.

Every panel joins `organization` and filters `deleted_at` on both sides. Without the join, a deleted org's live usage rows inflate the platform total and make it disagree with the per-org table directly beneath it. My fixture reproduced exactly that: `tracing_spans` read 19777 instead of 12000.

## Additional Context

- Builds on #2348 (read-only Postgres datasources), #2422 and #2437 (`bypassrls: true` on `grafana-viewer`, without which every cross-org query returns zero rows under the `usage` table's `FORCE ROW LEVEL SECURITY`).
- `BYPASSRLS` on the dev role is required for that same reason, not for convenience.
- Limits are deliberately absent. They live in the EE tier config and the org's signed license, neither readable from SQL, so a used-vs-limit percentage is not derivable here. An org's own limits are already on its Usage page. A panel on the dashboard says so.
- `seats`, `projects` and `endpoints` are absent too. They are live counts off the `user`, `project` and `endpoint` tables rather than accrued into `usage`, so they have no history to chart.
- Alerting is out of scope. Alertmanager is disabled and Grafana has no contact points, so alert rules need a destination decision first.
- Roles and grants are not in Terraform. `terraform/infrastructure/` is GCP-only, with no `postgresql` provider and no `postgresql_role`/`postgresql_grant` resources anywhere, so this follows the existing manifest-and-initdb pattern rather than pioneering a new home for one role.
- Noticed while in here, not fixed: the existing stg and prd `grafana-viewer-credentials.yaml` use `secretKey: password`, so `gsm-secrets-sync.sh` looks up `.stg.password`, which does not exist in the JSON. Their `GRAFANA_VIEWER_PASSWORD` entries are unreachable by that script as written, and were presumably set by hand. The dev wiring added here uses a matching name so the sync resolves it. Left the deployed ones alone.

## Testing

**Three prerequisites before any panel shows data.** These are environment state, not code:

1. Run `./infrastructure/config/gsm-secrets-sync.sh` for dev once the real password value is filled in. That creates `dev-grafana-viewer-password` and binds ESO access.
2. stg and prd need their one-time `GRANT` run. `bypassrls: true` reconciles onto a running cluster via `managed.roles`, but the `GRANT` lives in `postInitApplicationSQL`, which only fires on a fresh initdb. Check with `SELECT has_table_privilege('grafana-viewer','public.usage','SELECT'), rolbypassrls FROM pg_roles WHERE rolname = 'grafana-viewer';` — both must be true, otherwise panels fail with `permission denied for table usage`.
3. Dev's already-initialized cluster needs the `CREATE ROLE` and `GRANT` block run once by hand, same fresh-PVC caveat.

**What I verified locally:**

- All six panel queries and both template-variable queries against a seeded Postgres 17: zero failures, correct numbers. Soft-deleted usage rows and deleted orgs excluded, `COALESCE(display_name, name)` falling back correctly for an org with no display name, month-over-month deltas correct at +200 and -280.
- The dev role SQL exactly as it ships, against a table with `FORCE ROW LEVEL SECURITY` and the real `tenant_isolation` policy: `grafana-viewer` reads across all orgs, writes are refused with `permission denied`, and as a control an identically granted role *without* `BYPASSRLS` sees **zero rows**. That control is what justifies the `BYPASSRLS` line rather than assuming it.
- `kubectl kustomize kubernetes/clusters/{dev,stg,prd}/grafana-resources` all render the dashboard alongside their own datasource, and the embedded `spec.json` re-parses from each rendered output. Note the render path is the `grafana-resources` overlay, not the cluster root, which only renders the wrapper Argo Applications.
- `helm template` with dev values, confirming the init script and the `GRAFANA_VIEWER_PASSWORD` env var.

**To check after it syncs to dev:** the datasource passes Grafana's "Save & test", the dashboard appears in the Rhesis folder, the current-period table lists more than one org, and the 12-month trend is non-empty (the `91607f0dd412` and `9550c62e80a5` backfills already populated trailing months). Then cross-check one org's current-period figure against that org's own Usage page under `organizations/usage`. Both read the same rows, so a mismatch means a panel query is wrong.


asad-miah:
**Security: SQL injection via unescaped `$resource` variable**

Panels "Top consumers: $resource", "$resource by organization", and "Month over month: $resource" interpolate `$resource` raw (`u.resource = '$resource'`), while `$org` right next to it is correctly escaped as `${org:sqlstring}`. Grafana doesn't validate URL-supplied variable values (`?var-resource=...`) against the dropdown, and `grafana-viewer` has `BYPASSRLS` + `SELECT` on all public tables — so an overridden value could UNION-inject to read cross-tenant data from any table, not just `usage`.

**Fix:** use `${resource:sqlstring}` (or `:singlequote`) in all three panels, matching the `$org` pattern already used in this file.

harry-rhesis:
Good catch, and thanks. Valid, fixed in c27a7101b: all three panels now use `${resource:sqlstring}`, matching the `$org` pattern two lines away. That omission was a plain inconsistency on my part rather than a judgement call. I added the `$org` escaping in response to an earlier review comment and left `$resource` raw in the same query.

I reproduced it against a scratch Postgres before and after, and two details are worth recording.

**The UNION mechanism does not work as written, but the vulnerability is real.** A `UNION` payload terminated with `--` fails, because `--` only comments to end of line and the query continues with `AND u.period_start = ...` on the next. What does work is an OR term, which needs no tail-swallowing at all:

```
?var-resource=x OR (SELECT count(*) FROM public.token WHERE secret LIKE sk-live%) > 0 OR a=a
```

Since `AND` binds tighter than `OR`, the middle term stands alone and its truthiness decides the whole `WHERE`. That is a boolean oracle: read any table in `public` one predicate at a time. So the conclusion in your report is right, the exploit is blind rather than a direct dump.

**It also bypassed the soft-delete filtering.** With that payload the panel returned a row for `Deleted Co`, an organization with `deleted_at` set, because the injected `OR` short-circuits the `u.deleted_at IS NULL AND o.deleted_at IS NULL` conjunction. So the reach was not only cross-table but cross-tenant within `usage` itself.

After the fix the same payload returns zero rows, and all six panels still return correct figures against a fixture covering soft-deleted rows, deleted orgs, and two orgs sharing a display name.

Your point about `BYPASSRLS` plus full-schema SELECT being the amplifier is the right read, and it is the strongest argument yet for narrowing that grant. I have added it to the discussion on #2479, which tracks whether that posture should hold across all three environments. Worth noting the grant cannot simply be scoped to `usage` and `organization` at initdb time, since `ALTER DEFAULT PRIVILEGES` cannot name tables that do not exist yet, so narrowing it needs a different mechanism rather than a smaller grant list.

PR #2477, opened 2026-08-13
Log and mask server-side errors across all routers

akwasigroch:
## Purpose

Router error handling had two problems. **We were blind:** 98 broad `except Exception` blocks across the routers, 27 of which logged nothing at all, and no catch-all handler behind any of them — an uncaught error became a bare Starlette 500 that left no trace. There was no request id, so even a logged error couldn't be tied to a user's report. **We leaked:** 43 sites fed an exception caught by a broad handler straight into `detail=`, handing callers SQLAlchemy statements, OAuth provider internals, Celery broker URLs and connection targets. `auth.py` returned `f"Authentication failed: {str(e)}"`; `job.py` returned `f"Celery health check failed: {str(e)}"`.

An unexpected error should now tell the user nothing and the logs everything, tied together by an id the user can quote.

## What Changed

**Foundation (`3a10c76c5`)**

- `RequestIDMiddleware` assigns every request an id, echoed as `X-Request-ID`. Pure ASGI rather than `BaseHTTPMiddleware` — the latter runs the downstream app in a new anyio task, and the id has to stay readable both downstream in log records and back up in the exception handlers.
- A global `Exception` handler logs the full traceback and returns `{"detail": "An unexpected error occurred.", "error_id": "..."}`. A 5xx `HTTPException` handler does the same. **Sub-500s pass through completely untouched**, so deliberate validation messages like "Test set name already exists" still reach users unchanged.
- `internal_error(exc, context=...)` for the cases where a router genuinely adds context a global handler can't infer.
- Log records carry `request_id`, so service- and crud-level lines correlate to the same request.
- `tests/backend/security/test_router_error_leaks.py` AST-scans every router for the leak pattern and fails on regressions.

**Router cleanup (43 leaks removed across 4 commits)**

Most of the fix is deletion — once a global handler exists, a broad handler that logs nothing and re-wraps the error is pure noise. Net **−436 lines** in the routers.

## Additional Context

**Three latent bugs surfaced.** Broad handlers were swallowing the endpoints' own deliberate `HTTPException`s and re-wrapping them as 500s: `job.py`'s `503 "No Celery workers available"` and `404 "Task not found"`, and `test_run.py`'s `404 "Test run not found"`. Deleting the handlers fixes all three.

**Behavior change:** all 5xx response bodies are now generic, including hand-written ones like `"Failed to send test request to SDK"`. Nothing is lost — the original moves to the logs under the `error_id` — but it is visible. Two tests asserted the old behavior and were updated. The narrower alternative (genericise only exception-derived text) was rejected because it lets every *new* leak ship freely until someone notices.

**The guard already caught a regression.** The OWASP router (#2202) landed on `main` while this branch was in flight, shipping the same pattern on a 502 and a 500. The guard flagged both on merge — that is the case it exists for.

**Known gap:** `model.py::test_model_connection` returns HTTP 200 with `{"status": "error", "message": str(e)}`. The guard cannot see it — there is no `detail=`. Fixing it would change the response contract, so it is left for a follow-up.

Frontend needs no change to keep working; it reads `.detail`, which still exists. Surfacing `error_id` in error toasts is a separate follow-up.

## Testing

3907 backend tests pass, 42 skipped. Zero leaks across all 52 routers.

```bash
cd apps/backend
uv sync --extra all --extra ee
uv run pytest ../../tests/backend/security/ ../../tests/backend/routes/ \
  ../../tests/backend/app/ ../../tests/backend/auth/ ../../tests/backend/utils/ \
  ../../tests/backend/tasks/ -q -p no:randomly
```

To confirm the guard actually bites, reintroduce a leak and watch it fail:

```bash
sed -i '' 's/detail="Failed to generate content"/detail=f"Failed to generate content: {e}"/' \
  apps/backend/src/rhesis/backend/app/routers/services.py
cd apps/backend && uv run pytest ../../tests/backend/security/test_router_error_leaks.py -q
git checkout apps/backend/src/rhesis/backend/app/routers/services.py
```

End to end: `./rh dev status` for this checkout's ports, then trigger a failing endpoint and confirm the response body is generic, `X-Request-ID` is set, and the same id appears in the logs beside a full traceback.

---

## Update: code-review fixes (commits 7–8)

A `/code-review high` pass over this branch found eight issues. Six are fixed here; two are deliberately deferred.

**The blanket 5xx masking was too blunt in one direction.** `POST /endpoints/test` and `invoke_endpoint` exist to report what is wrong with the *caller's* endpoint — a refused connection, a rejected token — and masking those told the user nothing while protecting nothing of ours. `UpstreamHTTPException` now marks that case and the global handler passes its detail through.

The exemption is deliberately narrow, because the obvious version of it would have introduced a new leak. Two places where the boundary was not where it looked:

- `EndpointService` wraps *our own* exceptions in `EndpointInvocationError` as `error_type="internal_error"` (`services/endpoint/service.py:348`), so the router branches on that discriminator rather than trusting the exception type.
- `testing.py`'s `try` covered endpoint construction and input enrichment as well as the invocation, so the exemption is scoped to the `invoke()` call alone and setup failures still go through `internal_error`.

A test asserts a plain 500 is still masked, so the exemption cannot quietly widen.

**Also fixed:** CORS `expose_headers` omitted `X-Request-ID`, so browser JS could never read the id the middleware exists to emit; `internal_error(status_code=400)` answered "An unexpected error occurred.", describing a client error in server-error words; `internal_error` logged and then the handler logged the same failure again; both the middleware and the handler set `X-Request-ID`, so every 5xx carried it twice plus a redundant `X-Error-Id`; and `POST /tests/bulk` with an empty list returned 500 instead of 400 (pre-existing, but this branch stripped the message that made it diagnosable).

**Deferred, tracked for follow-up:** `ValueError` bleed paths — five sites in `services/` launder a provider or parser error into a `ValueError` message that narrow router handlers then hand to the user (`raise ValueError(f"LLM generation failed: {e}")`). Same concern as this PR, one hop removed, and invisible to the guard because it crosses a `raise`/`catch` boundary. Also deferred: inbound `X-Request-ID` is adopted from any caller, so a client can reuse an id and make correlation ambiguous.

## Revised review order

1. **`error_handlers.py` + `utils/request_context.py`** (commits 1, 8) — the contract. The middleware-ordering argument for reading the id from the request scope, the `status_code < 500` passthrough, and the `UpstreamHTTPException` exemption and its two guards.
2. **`test_router_error_leaks.py`** (commit 5) — defines what counts as a leak.
3. **Batch commits 2–4** — mostly deletions; reading the added lines is enough. `garak.py` is the riskiest (unwrapped `try` blocks with dedented bodies).
4. **Commits 6–7** — the OWASP fix (the guard catching a regression) and the bulk-create status fix.

## Verification

5795 backend tests pass, 43 skipped. Zero leaks across all 52 routers.

---

## Update: second review pass (seven follow-up commits)

A review of the whole branch found the masking went too far in one direction and not far enough in another. Both are fixed here, along with a bug that made this PR's central promise false in every deployed environment.

**Logs held no traceback in production.** `JsonLogFormatter` built its payload from `record.getMessage()` and never read `record.exc_info`, and `jsonLoggerEnabled: true` is set in dev, stg and prd. So `logger.exception` emitted one line with no stack, no frames and no exception class — the client was told nothing and the log kept only `str(exc)`. That is the half of "mask the response, log the reason" that was silently not happening, and it bit hardest exactly where this branch deleted broad handlers on the grounds that the global handler would log the same traceback. Tracebacks now land in a `stack_trace` field, redacted. The request id is also passed explicitly on the unhandled path, because `ServerErrorMiddleware` runs after `RequestIDMiddleware` clears the ContextVar — the field had been null on precisely the lines an `error_id` points at.

**Deliberate 5xx messages had no way through.** `PublicHTTPException` passes a literal detail through and logs at WARNING with no stack, since a hand-written 503 is not a fault to debug and its traceback only points back at the raise. `UpstreamHTTPException` becomes one case of it. This restores "Garak package is not installed", "No Celery workers available" and "Failed to send test request to SDK" — strings this branch left in the code while the handler replaced them, so the code read as if they still arrived. `internal_error` gains `public_detail` for the same reason at 4xx.

**A wrong credential for a service the user connected themselves was masked.** `handle_mcp_exception` deliberately remaps an MCP 401/403 to a 502 so the reason survives and the frontend does not clear the session — and the global handler masked exactly that, so a bad Notion token read as our service being down. It returns `UpstreamHTTPException` now, and only for `MCPApplicationError`, whose detail really is the tool's parsed response. `services.py` had stripped provider reasons at 400, a status this PR's own contract says passes through, turning "invalid api key" into "Failed to generate content" while logging it at ERROR with a stack as though it were ours. `tools.py` reports an unreachable instance again, in a drawer whose only job is showing why a connection failed.

**The branch had introduced a leak of its own.** `services/endpoint/testing.py` wrapped `invoke()` in a bare `except Exception` and declared whatever it caught to be upstream detail, shipping `str(exc)` on a 500 — including SQLAlchemy errors carrying the statement and its bound parameters, and `HTTPException`s that stringify to `"400: ..."`, so a status code arrived inside a 500 body. The invokers *return* an error response for genuine upstream failures, so almost nothing that block caught was upstream at all. `rest_invoker` did the same thing through a channel nothing was watching: `str(e)` in an HTTP 200 body, invisible to the guard because there is no `detail=`. `test_endpoint_mapping`, whose only job is explaining what is wrong with the mappings the user just typed, had been left out of the exemption entirely.

**Caller faults were logged as server faults.** 422s logged at ERROR with a traceback immediately before the function whose docstring says a 422 is not a server fault. Login failures reported an IdP outage as a 400 with one stackless warning, where `main` logged an ERROR with a traceback — both axes wrong at once. A single connection-refused against a user's endpoint produced three ERROR records and two tracebacks; it is now one warning.

**Redaction was wrong in both directions.** It mangled ordinary diagnostics — `api key: not configured` became `api key: [REDACTED] configured` — while missing the secrets that actually arrive: provider errors read `Incorrect API key provided: sk-…`, with filler between the keyword and the value, and header dumps put a quote between the name and `Bearer …`. The URL pattern never matched SQLAlchemy's real `postgresql://` scheme, and the Celery broker is `redis://`. Both directions are fixed, with key shapes matched without needing a keyword at all.

**The guard let 14 of 20 crafted leaks through.** The likeliest regression was the simplest: only `detail=` keywords were inspected, so `HTTPException(500, f"...{e}")` with a positional detail was invisible. `BROAD_EXCEPTIONS` held only `Exception` and `BaseException`, treating `SQLAlchemyError` and `httpx.HTTPError` as narrow and deliberate despite carrying SQL text and connection strings. Taint followed plain assignment only, so `+=`, tuples, subscripts and attributes all escaped. Scanning now covers `routers/`, `utils/` and `services/` recursively; `services/invokers/` is exempt because every string there reports on the user's own endpoint.

### Two things worth pushing back on

The guard no longer flags a 4xx `detail` built from an exception, which it used to. Sub-500 details pass through by design and three live sites depend on it, but this is less coverage than before, not a refactor.

`get_client_credentials_token` changed from 500 to 502, with two tests updated. More accurate — it is the caller's own token endpoint failing — but it is a status change on a public path.

### Known gaps, deliberately not fixed here

15 pre-existing leaks across 8 files put `str(e)` in a 200 body or a result object. They are the same bug this PR removes, through a channel it never examined, and they are ratcheted in `KNOWN_LEAKS` so the count can only fall. Separately: `routers/file_import.py` orders `except ValueError` ahead of `except ValidationError`, so a caller fault lands in the 404 arm; the SDK's `mcp/agent.py` wraps any exception as `MCPValidationError` at 422, a sub-500 that passes through unmasked; `services/github.py` answers 200 with an empty body for an invalid `repo_url`; and an unhandled 500 comes from `ServerErrorMiddleware`, outside `CORSMiddleware`, so browser JS can read neither its body nor `X-Request-ID` — meaning the `expose_headers` change earlier in this PR does not help that path. A test now pins that behaviour so it cannot change silently.

## Verification (current)

6165 passed, 43 skipped, 1 xfailed. Ruff clean on all changed files apart from six findings that reproduce on `origin/main`.

```bash
cd apps/backend
uv sync --extra all --extra ee
uv run pytest ../../tests/backend/security/ ../../tests/backend/routes/ ../../tests/backend/app/ \
  ../../tests/backend/auth/ ../../tests/backend/utils/ ../../tests/backend/tasks/ \
  ../../tests/backend/services/ -q -p no:randomly
```

The traceback fix is the one thing no test in the earlier rounds would have caught, so it is worth seeing directly:

```bash
cd apps/backend && uv run python -c "
import json, logging, sys
from rhesis.backend.logging.logging_config import JsonLogFormatter, RedactingFormatter
f = RedactingFormatter(JsonLogFormatter())
try:
    raise ValueError('boom api_key=sk-proj-aBcD1234EfGh5678')
except ValueError:
    r = logging.LogRecord('t', logging.ERROR, 'p', 1, 'ctx', (), sys.exc_info())
d = json.loads(f.format(r))
print('frames:', 'File' in d['stack_trace'], '| redacted:', '[REDACTED]' in d['stack_trace'])
"
```

### DIFF
--- a/apps/backend/src/rhesis/backend/app/error_handlers.py
+++ b/apps/backend/src/rhesis/backend/app/error_handlers.py
@@ -10,7 +10,7 @@ from fastapi.exceptions import RequestValidationError
 from starlette.exceptions import HTTPException as StarletteHTTPException
 from starlette.responses import JSONResponse
 
-from rhesis.backend.app.utils.request_context import get_request_id, request_id_of
+from rhesis.backend.app.utils.request_context import request_id_of
 
 logger = logging.getLogger(__name__)
 
@@ -36,7 +36,21 @@ def public_message(status_code: int) -> str:
     return PUBLIC_ERROR_MESSAGES[500]
 
 
-class UpstreamHTTPException(HTTPException):
+class PublicHTTPException(HTTPException):
+    """A 5xx whose detail is written for the caller to read.
+
+    Masking is the default because a detail built from an exception leaks
+    whatever the exception happened to carry. That reasoning does not apply to
+    a fixed string someone wrote on purpose -- "Garak package is not installed",
+    "No Celery workers available" -- which names the one thing the caller can
+    act on and reveals nothing.
+
+    The detail must be a literal, not interpolated from an exception. The moment
+    it carries `str(exc)` this is a leak with a blessing on it.
+    """
+
+
+class UpstreamHTTPException(PublicHTTPException):
     """A failure of the *caller's* system, not ours -- detail is theirs to see.
 
     Endpoint testing and invocation exist to report what is wrong with a user's
@@ -49,7 +63,13 @@ class UpstreamHTTPException(HTTPException):
     """
 
 
-def internal_error(exc: Exception, *, context: str, status_code: int = 500) -> HTTPException:
+def internal_error(
+    exc: Exception,
+    *,
+    context: str,
+    status_code: int = 500,
+    public_detail: str | None = None,
+) -> HTTPException:
     """Log ``exc`` and return an HTTPException that reveals none of it.
 
     Only for cases where the router genuinely adds context a global handler
@@ -59,12 +79,20 @@ def internal_error(exc: Exception, *, context: str, status_code: int = 500) -> H
     The level follows the status: a 4xx is the caller's mistake, so it gets one
     warning line and no stack. If a failure is ours to debug it belongs in the
     5xx branch, which is what puts a traceback in the log.
+
+    ``public_detail`` replaces the generic message with one the caller can act
+    on -- "No generation model is configured", not "The request could not be
+    processed." It must be a literal: interpolating ``exc`` into it defeats the
+    whole point of routing through this function.
     """
     if status_code >= 500:
-        logger.exception("[%s] %s: %s", get_request_id() or "-", context, exc)
+        logger.exception("%s: %s", context, exc)
     else:
-        logger.warning("[%s] %s: %s", get_request_id() or "-", context, exc)
-    http_exc = HTTPException(status_code=status_code, detail=public_message(status_code))
+        logger.warning("%s: %s", context, exc)
+    # A 5xx detail only survives the global handler on a PublicHTTPException, so
+    # asking for one and getting a plain HTTPException would silently mask it.
+    cls = PublicHTTPException if public_detail else HTTPException
+    http_exc = cls(status_code=status_code, detail=public_detail or public_message(status_code))
     # Already in the log with `context` attached; without this the handler below
     # logs the same failure a second time, saying less.
     http_exc.rhesis_logged = True
@@ -84,6 +112,10 @@ async def unhandled_exception_handler(request: Request, exc: Exception) -> JSONR
         request.method,
         request.url.path,
         exc,
+        # RequestIDMiddleware's `finally` has already cleared the ContextVar by
+        # now, so the logging filter cannot recover the id for these records --
+        # the ones an error_id actually points at. Pass it in by hand.
+        extra={"request_id": request_id},
     )
     # Header set here, not by the middleware: an unhandled exception is turned
     # into a response by ServerErrorMiddleware, which sits outside it.
@@ -105,18 +137,20 @@ async def http_exception_handler(request: Request, exc: StarletteHTTPException):
         return await default_http_exception_handler(request, exc)
 
     request_id = request_id_of(request)
-    upstream = isinstance(exc, UpstreamHTTPException)
+    deliberate = isinstance(exc, PublicHTTPException)
 
     if not getattr(exc, "rhesis_logged", False):
-        logger.error(
-            "[%s] %s on %s %s: %s",
-            request_id or "-",
+        # A deliberate 5xx is not a fault of ours to debug: an unreachable user
+        # endpoint or a missing optional package is a warning with a readable
+        # reason, not an ERROR with a stack that only points back at the raise.
+        logger.log(
+            logging.WARNING if deliberate else logging.ERROR,
+            "%s on %s %s: %s",
             exc.status_code,
             request.method,
             request.url.path,
             exc.detail,
-            # An upstream failure is not our stack; the detail is the useful part.
-            exc_info=not upstream,
+            exc_info=not deliberate,
         )
 
     # No X-Request-ID here: this response is built inside ExceptionMiddleware,
@@ -124,7 +158,7 @@ async def http_exception_handler(request: Request, exc: StarletteHTTPException):
     return JSONResponse(
         status_code=exc.status_code,
         content={
-            "detail": exc.detail if upstream else public_message(exc.status_code),
+            "detail": exc.detail if deliberate else public_message(exc.status_code),
             "error_id": request_id,
         },
         headers=getattr(exc, "headers", None),
--- a/apps/backend/src/rhesis/backend/app/error_handlers.py
+++ b/apps/backend/src/rhesis/backend/app/error_handlers.py
@@ -50,16 +50,23 @@ class UpstreamHTTPException(HTTPException):
 
 
 def internal_error(exc: Exception, *, context: str, status_code: int = 500) -> HTTPException:
-    """Log ``exc`` in full and return an HTTPException that reveals none of it.
+    """Log ``exc`` and return an HTTPException that reveals none of it.
 
     Only for cases where the router genuinely adds context a global handler
     can't infer. Otherwise don't catch at all — the global handler logs the
     same traceback and the `except` block is just noise.
+
+    The level follows the status: a 4xx is the caller's mistake, so it gets one
+    warning line and no stack. If a failure is ours to debug it belongs in the
+    5xx branch, which is what puts a traceback in the log.
     """
-    logger.exception("[%s] %s: %s", get_request_id() or "-", context, exc)
+    if status_code >= 500:
+        logger.exception("[%s] %s: %s", get_request_id() or "-", context, exc)
+    else:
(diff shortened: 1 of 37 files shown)


harry-rhesis (commented):
Nice work on this. The design is sound, the testing is thorough, and the router cleanup is satisfying to read. None of the items below are blockers.

## Double-logging on a few `services.py` paths (non-blocking)

`services.py:248-250` (ModelConfigurationError) and `services.py:468-470` (RuntimeError) both log with `exc_info=True` and then raise a 500 HTTPException without `rhesis_logged = True`. The global handler logs the same error a second time with a traceback. Same thing at line 557-559 (recent activities), though that one uses a fixed-string detail so it's safe either way.

All three are pre-existing and behaviorally correct (the global handler masks the detail). But replacing them with `internal_error` would cut the duplicate log lines. Fine to defer.

## `handle_execution_error` fallback still constructs the leak pattern (non-blocking)

`execution_validation.py:139-142`:

```python
error_msg = str(error) if str(error) else "An unexpected error occurred"
return HTTPException(status_code=500, detail=f"Failed to {operation}: {error_msg}")
```

The global handler masks this today. But if a future caller uses `handle_execution_error` outside the request lifecycle (background task, CLI), the leak surfaces. The AST guard can't see through function calls, so it won't catch it. I know the ValueError bleed paths are already tracked for follow-up; the broad fallback at the end of this function is a separate instance of the same concern.

## Header-duplication test could cover `UpstreamHTTPException` (non-blocking)

`test_request_id_header_is_not_duplicated` checks `/boom`, `/server-error`, and `/bad-request`. Adding `/upstream` to the list would confirm no duplication on the upstream exemption path too.

---

Thanks for the thorough writeup in the PR description, especially the middleware-ordering rationale and the note on the three latent bugs this surfaced.

PR #2496, opened 2026-08-15
fix(dev): fail CI on ConfigMap and ExternalSecret key overlap

### LINKED ISSUE #2420: Dedupe env var keys between Helm ConfigMap and ExternalSecret

feiiiiii5:
Fixes #2420 (step 4: the CI gate)

# What

An env key declared in both the chart ConfigMap and an ExternalSecret is silently shadowed: Kubernetes lets the later envFrom source win, so editing values-*.yaml for that key is a no-op with no warning. This PR adds a lint gate so a new overlap fails CI, and pins the duplicates that exist today in an explicit allowlist.

# Change

New scripts/check-env-overlap.py renders the chart per env (helm template) and fails when a key appears in both the rendered ConfigMap data and any ExternalSecret secretKey in kubernetes/clusters/<env>/external-secrets. The allowlist matches today's reality exactly (dev: APP_DB_USER/DB_HOST/DB_NAME/OTEL_PROCESSOR_ENDPOINT; stg and prd: ANALYTICS_DB_* + OTEL_PROCESSOR_ENDPOINT). Two failure modes: a new overlap, or an allowlist entry that no longer overlaps (so the allowlist cannot go stale as the duplicates from the issue are resolved).

lint.yml gains a helm-overlap job gated on charts/** and kubernetes/** changes, and the job joins the lint gate's needs list.

# Verification

Local run against the current tree: all three envs report OK with the exact allowlist. Removing one allowlisted key fails with "new ConfigMap/ExternalSecret overlap" (exit 1). Steps 1-3 of the issue (comparing values in Secret Manager and deleting the duplicates) still need cloud access and are left to the team; this gate makes the cleanup safe once done.

# Scope

scripts/check-env-overlap.py (new) and .github/workflows/lint.yml. No chart or kubernetes manifest values change.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -40,6 +40,10 @@ jobs:
             skill:
               - 'skills/rhesis/**'
               - 'scripts/skill/**'
+            helm:
+              - 'charts/**'
+              - 'kubernetes/**'
+              - 'scripts/check-env-overlap.py'
 
       - name: Resolve lint targets
         id: resolve
@@ -208,8 +212,27 @@ jobs:
       - name: Validate skill
         run: python scripts/skill/validate.py
 
+  helm-overlap:
+    needs: changes
+    if: ${{ github.event_name == 'workflow_dispatch' || contains(needs.changes.outputs.targets, 'helm') }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+
+      - name: Install Helm
+        uses: azure/setup-helm@v4
+        with:
+          version: latest
+
+      - name: Install Python dependencies
+        run: python -m pip install pyyaml
+
+      - name: Check ConfigMap / ExternalSecret overlap
+        run: python scripts/check-env-overlap.py
+
   lint:
-    needs: [changes, python, migrations, frontend, styles, docs, skill]
+    needs: [changes, python, migrations, frontend, styles, docs, skill, helm-overlap]
     if: always()
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -246,7 +246,8 @@ jobs:
                         "frontend=${{ needs.frontend.result }}" \
                         "styles=${{ needs.styles.result }}" \
                         "docs=${{ needs.docs.result }}" \
-                        "skill=${{ needs.skill.result }}"; do
+                        "skill=${{ needs.skill.result }}" \
+                        "helm-overlap=${{ needs.helm-overlap.result }}"; do
             echo "$RESULT"
             case "${RESULT#*=}" in
               success|skipped) ;;
--- a/scripts/check-env-overlap.py
+++ b/scripts/check-env-overlap.py
@@ -48,6 +48,8 @@ def configmap_keys(env: str) -> set[str]:
             "charts/rhesis",
             "-f",
             f"charts/rhesis/values-{env}.yaml",
+            "--show-only",
+            "templates/configmap.yaml",
         ],
         cwd=REPO_ROOT,
         capture_output=True,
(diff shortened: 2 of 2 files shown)


feiiiiii5:
Addressed in commit `38383832c`:
- `lint` now checks `helm-overlap` in its result loop, so a failing overlap gate fails lint.
- `configmap_keys` renders only `templates/configmap.yaml` (`--show-only`), so dependency ConfigMaps cannot produce false positives.
- overlap uses only `spec.data[].secretKey`; entries without `secretKey` are warned about and skipped.

Local verification: `python scripts/check-env-overlap.py` exits 0 for dev/stg/prd; ruff check + format are clean.

feiiiiii5:
Closing as duplicate — upstream main already ships `scripts/check-k8s-secret-overlap.sh` (bash), which covers the same Helm ConfigMap / ExternalSecret overlap detection that this PR's `scripts/check-env-overlap.py` (Python + allowlist) was built for.

The peqy review found the scoping correct, but the upstream script is the canonical implementation and this PR would add a second, parallel tool for the same job. Branch `fix/env-overlap-dedupe` is kept for reference.

PR #2497, opened 2026-08-15
fix(dev): drop unused provider API keys from deployment secrets

### LINKED ISSUE #2466: Remove unused provider API keys from the backend environment
## Summary

The backend process carries provider API keys (`OPENAI_API_KEY`, `GEMINI_API_KEY`, `GOOGLE_API_KEY`, `AZURE_OPENAI_API_KEY`) that most deployments do not need. Any code path that reaches a provider without an explicit key silently picks these up, which means a tenant's model can end up running on our credentials.

Each deployment should ship exactly the credential its own default model requires, and nothing else.

## Background

Every provider falls back to the process environment when handed a falsy key. In LiteLLM (`litellm/main.py:2128-2133`):

```python
api_key = (
    api_key
    or litellm.api_key
    or litellm.openai_key
    or get_secret("OPENAI_API_KEY")
)
```

Our own SDK providers do the same (`sdk/src/rhesis/sdk/models/providers/openai.py:20`, `gemini.py:42`).

So a tenant Model row with no API key and no endpoint resolves, runs on our account, and looks to the tenant like a working configuration. We pay the provider, and nothing is billed internally because the row belongs to the tenant and tenant rows are correctly never charged.

#2461 closes the known path into this: model creation rejects a row with neither a key nor an endpoint, and the resolution layer refuses to build one. This issue is the layer underneath, which does not depend on us having enumerated every path. **The fallback can only borrow what is present in the environment.** Remove the keys and the whole class of leak goes away, including paths nobody has thought of yet.

## What needs doing

- [ ] For each deployment (dev, stg, prd), read the configured `DEFAULT_GENERATION_MODEL`, `DEFAULT_EVALUATION_MODEL`, `DEFAULT_EXECUTION_MODEL` and `DEFAULT_EMBEDDING_MODEL` and note which provider credential each actually requires. The in-code defaults are `rhesis/rhesis-default` (`apps/backend/src/rhesis/backend/app/config/settings.py:284-298`), which needs none of these keys.
- [ ] Drop every provider key from that deployment's secrets that no default model needs. Sources to update: `infrastructure/k8s/manifests/secrets/rhesis-secrets.yaml.example` and the per-cluster external secrets under `kubernetes/clusters/*/external-secrets/`.
- [ ] Confirm nothing else in the deployment consumes them. In application source these variables appear only as provider-constructor fallbacks, which is the behaviour we are removing.
- [ ] Keep them in local dev and CI where they are legitimately used for running against real providers.

## Notes

This matters for multi-tenant (SaaS) deployments specifically. In a self-hosted deployment a tenant row falling back to the environment is falling back to the operator's own key, which is not a leak.

Worth checking the logs for the warning added by #2461 before and after, to see whether any real row has been hitting this.

## Related

- #2461 adds the write-time rejection and the runtime refusal
- #2382 centralized token accrual, where this was found

feiiiiii5:
Fixes #2466

# What

Removes the unused provider API keys from deployment secrets. GEMINI_API_KEY and GOOGLE_API_KEY are dropped from the dev, stg and prd ExternalSecrets, and OPENAI_API_KEY / AZURE_OPENAI_API_KEY / GEMINI_API_KEY / GOOGLE_API_KEY are dropped from the example secret manifest.

# Why

Every deployment's DEFAULT_*_MODEL is vertex_ai/gemini-3.1-flash-lite (generation, evaluation, execution) and vertex_ai/text-embedding-005 (embedding). The vertex_ai provider authenticates with GOOGLE_APPLICATION_CREDENTIALS; the only consumers of the API keys were the google-genai / openai provider environment fallbacks, so a tenant model row with no key could silently resolve and run on our credentials. #2461 closes the known row-level path; removing the keys from the environment closes the whole class.

# Verification

The values files for all three environments confirm the vertex_ai defaults (charts/rhesis/values-{dev,stg,prd}.yaml). Source grep confirms the four key names appear only as provider-constructor fallbacks (sdk/src/rhesis/sdk/models/providers/gemini.py) and in the leak description comment (apps/backend/.../user_model_utils.py) — no other consumer. scripts/check-env-overlap.py still passes for all three envs.

# Scope

kubernetes/clusters/{dev,stg,prd}/external-secrets/rhesis-app-secrets.yaml, infrastructure/k8s/manifests/secrets/rhesis-secrets.yaml.example, charts/rhesis/values-dev.yaml (comment). Left untouched: the Secret Manager bootstrap tooling under infrastructure/config/ still lists GEMINI_API_KEY for secret creation; removing it is a separate onboarding-tooling decision.

### DIFF
--- a/charts/rhesis/values-dev.yaml
+++ b/charts/rhesis/values-dev.yaml
@@ -40,7 +40,7 @@ config:
   # GCP
   vertexAiProject: "playground-437609"
 
-  # AI model — Vertex defaults in dev (GEMINI_API_KEY + GOOGLE_APPLICATION_CREDENTIALS in ESO);
+  # AI model — Vertex defaults in dev (GOOGLE_APPLICATION_CREDENTIALS in ESO);
   # avoids requiring RHESIS_API_KEY for rhesis/rhesis
   defaultGenerationModel: "vertex_ai/gemini-3.1-flash-lite"
   defaultEvaluationModel: "vertex_ai/gemini-3.1-flash-lite"
--- a/infrastructure/k8s/manifests/secrets/rhesis-secrets.yaml.example
+++ b/infrastructure/k8s/manifests/secrets/rhesis-secrets.yaml.example
@@ -69,6 +69,18 @@ data:
   # nothing else: a provider constructor falls back to the process environment
   # when handed a falsy key, so an unused key here silently lets tenant models
   # run on your account.
+  #
+  # Uncomment the entry for each provider you actually use:
+  #
+  # Azure OpenAI (if using Azure)
+  # AZURE_OPENAI_API_KEY: <BASE64_ENCODED_AZURE_OPENAI_API_KEY>
+  #
+  # OpenAI (if using OpenAI directly)
+  # OPENAI_API_KEY: <BASE64_ENCODED_OPENAI_API_KEY>
+  #
+  # Google Gemini (if using Google AI)
+  # GEMINI_API_KEY: <BASE64_ENCODED_GEMINI_API_KEY>
+  # GOOGLE_API_KEY: <BASE64_ENCODED_GOOGLE_API_KEY>
 
   # =============================================================================
   # 📧 EMAIL CONFIGURATION
--- a/infrastructure/k8s/manifests/secrets/rhesis-secrets.yaml.example
+++ b/infrastructure/k8s/manifests/secrets/rhesis-secrets.yaml.example
@@ -65,17 +65,10 @@ data:
   # =============================================================================
   # 🤖 AI MODEL CONFIGURATION
   # =============================================================================
-  # Choose one or more AI providers based on your needs
-
-  # Azure OpenAI (if using Azure)
-  AZURE_OPENAI_API_KEY: <BASE64_ENCODED_AZURE_OPENAI_API_KEY>
-
-  # OpenAI (if using OpenAI directly)
-  OPENAI_API_KEY: <BASE64_ENCODED_OPENAI_API_KEY>
-
-  # Google Gemini (if using Google AI)
-  GEMINI_API_KEY: <BASE64_ENCODED_GEMINI_API_KEY>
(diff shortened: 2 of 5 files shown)


feiiiiii5:
Confirmed on the current branch:
- All Helm workloads (backend, worker, frontend, chatbot, polyphemus, otel-collector, telemetry-processor) consume `rhesis-app-secrets` via `envFrom`. Their configured defaults in dev/stg/prd are all `vertex_ai/gemini-*`, and `GOOGLE_APPLICATION_CREDENTIALS` stays in the same ExternalSecret, so none of them needs `GEMINI_API_KEY`/`GOOGLE_API_KEY` to keep running.
- The SDK and standalone agents still read `GEMINI_API_KEY`/`GOOGLE_API_KEY` when someone explicitly selects the `gemini` provider, but that is exactly the path this PR closes for platform tenants: after this change it fails loudly instead of silently falling back to a platform key.
- `apps/worker/k8s/deployment.yaml` reads both keys from `rhesis-worker-secrets`, not from `rhesis-app-secrets`, and this PR leaves that manifest untouched, so that worker keeps its own provisioning.

asad-miah (commented):
Thanks a lot for the fix. @feiiiiii5 
one small nit:
The example secrets file used to show self-hosters exactly which env var to set per provider (Azure/OpenAI/Gemini). The new comment explains why to keep it minimal but no longer names any of those vars, so someone self-hosting with a non-Vertex provider now has to dig into SDK source to find the right key name. Could you please take a look on it.

feiiiiii5:
Good catch — restored the per-provider variable names in 3297a0976: the example file now keeps the minimization guidance and lists AZURE_OPENAI_API_KEY / OPENAI_API_KEY / GEMINI_API_KEY / GOOGLE_API_KEY as commented entries, so self-hosters can still see exactly which env var to set per provider without shipping unused keys.

harry-rhesis (approved):
Thanks for this, @feiiiiii5. Clean rationale, good verification, and the scope is well-contained. The example manifest update with the comments is a nice touch too.

Issue #2429, opened 2026-08-11
Enforce that a test set's type matches its tests' types


harry-rhesis:
## Summary

Nothing prevents a test set whose `test_set_type` is `Multi-Turn` from containing Single-Turn tests, or vice versa. `create_test_set_bulk` accepts a mismatched payload and persists it.

## Background / Context

A test set carries a `test_set_type`, and every test under it carries its own `test_type`. These are resolved independently and never checked against each other.

`bulk_create_tests` (`apps/backend/src/rhesis/backend/app/services/test.py:685-702`) types each test on its own, in this precedence: the test's explicit `test_type`, then auto-detection from content (`test_configuration.goal` means Multi-Turn, a `prompt` means Single-Turn), then the parent set's type, then the default. The parent type is only a last-resort fallback, so a test carrying a `prompt` always resolves to Single-Turn even when the set is declared `Multi-Turn`. There is no validator on `TestSetBulkCreate` comparing the two.

This is not only cosmetic. Preflight decides which checks apply from the **set** type (`_is_multi_turn` in `apps/backend/src/rhesis/backend/app/routers/preflight.py:42`), while execution decides the run mode from each **test's** type (`get_test_type` in `apps/backend/src/rhesis/backend/tasks/execution/modes.py`). When the two disagree, preflight validates the wrong thing: a set labelled `Single-Turn` holding multi-turn tests skips the execution-model check and can then fail at run time, and a set labelled `Multi-Turn` holding single-turn tests demands an execution model it never uses.

The `generate_test_set` path is safe by construction. `MultiTurnSynthesizer` sets `test_type` on every test and `test_set_type` on the set, so they cannot diverge. The exposure is `create_test_set_bulk` and any direct API client.

PR #2428 fixed the agent-facing side of this: the architect now has criteria for choosing the type, the correct multi-turn `test_configuration` shape, and the required `test_set_type` field documented. That reduces how often the architect gets it wrong, but it is guidance to a model, not enforcement. Any non-agent client can still POST a mismatched payload.

## Policy (decided)

**Mixed test sets are not allowed.** A test set is uniformly `Single-Turn` or uniformly `Multi-Turn`. A payload whose tests do not all resolve to the declared `test_set_type` is rejected, including one that mixes both shapes. Callers wanting both split them into two test sets.

This makes the set-level type authoritative and keeps preflight and execution reasoning about the same thing.

## Goals

- A test set and its tests cannot disagree about their turn type.
[...]

PR #2499, opened 2026-08-16
fix(backend): validate test types against the set type

feiiiiii5:
## Root Cause

Nothing prevented a test set whose `test_set_type` is `Multi-Turn` from containing Single-Turn tests, or vice versa. The effective-type precedence (explicit `test_type`, then content auto-detection, then the parent set type, then the default) lived only inside `bulk_create_tests`, so `create_test_set_bulk` persisted mismatched or mixed payloads silently.

## Fix

1. Extracted the precedence into one shared helper, `app/utils/test_type_resolution.py:resolve_effective_test_type()`. Both the schema validator and `bulk_create_tests` resolve through it now — two copies will drift, and a drift reintroduces exactly this bug.
2. Added a `model_validator` to `TestSetBulkCreate` that rejects any test whose effective type disagrees with `test_set_type`. The 422 names the offending test index, its effective type, the declared type, and the fix (set `test_type` on the test, or send `test_configuration` instead of `prompt` and vice versa). A payload mixing both shapes additionally says to split it into two test sets.
3. SDK `_infer_test_set_type` now raises on a mixed collection (instead of "any multi-turn wins" silently constructing a payload the API rejects). `from_csv` / `from_json` / `from_jsonl` all go through it.
4. SDK `push()` validates tests against the declared set type locally, so callers get the same actionable error without a round-trip 422.
5. Audited the other writers into `bulk_create_tests`: the Garak sync path always builds explicit Single-Turn tests (consistent by construction), and the CSV/JSONL/JSON import path flows through the SDK constructors now covered by (3) and (4).

## Test

- `tests/backend/schemas/test_test_set.py` — 11 cases: mismatched prompt-only vs goal tests both directions, mixed payloads with both declared types, uniform payloads unchanged, explicit `test_type` wins over content, loose casing still enforced, offending index named in the error.
- `tests/backend/utils/test_test_type_resolution.py` — 8 cases pinning the precedence chain in one place.
- `tests/sdk/entities/test_test_set.py` — 7 new cases: mixed CSV/JSON raise with split guidance, uniform untyped multi/single inference, mixed untyped raises, `push()` fails locally on mismatched and mixed collections.

Pre-fix FAIL → post-fix PASS verified: backend schema tests 7 failed / 3 passed on `main`, 18/18 after; SDK tests 6 failed / 1 passed on `main`, 65/65 after. `ruff check` and `ruff format --check` clean.

## Diff Scope

- `apps/backend/.../utils/test_type_resolution.py` (new, 45 lines) — shared precedence helper.
- `apps/backend/.../schemas/test_set.py` — `TestSetBulkCreate.validate_tests_match_set_type`.
- `apps/backend/.../services/test.py` — `bulk_create_tests` now calls the helper (no behavior change).
- `sdk/.../entities/test_set.py` — `_infer_test_set_type` early fail, `_effective_type`, `_validate_tests_match_set_type`, `_push_create` hook.
- Three test files.

## Notes

- Step 6 of #2429 (existing rows): this validator guards new writes only. Existing mismatched or mixed rows are accepted as legacy — they keep working until touched, which the issue lists as an acceptable outcome; the decision is recorded here rather than silently migrating data.
- This is a breaking change for clients sending mixed or mismatched payloads — intended per #2429; the errors carry the documented path (split into two sets).

### DIFF
--- a/apps/backend/src/rhesis/backend/app/schemas/test_set.py
+++ b/apps/backend/src/rhesis/backend/app/schemas/test_set.py
@@ -2,7 +2,7 @@ from datetime import datetime
 from enum import Enum
 from typing import Any, Dict, List, Optional, Union
 
-from pydantic import UUID4, BaseModel, ConfigDict, Field, field_validator
+from pydantic import UUID4, BaseModel, ConfigDict, Field, field_validator, model_validator
 
 from rhesis.backend.app.constants import TestSetType, TestType
 from rhesis.backend.app.schemas import Base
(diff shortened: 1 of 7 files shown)


feiiiiii5:
Closing in favor of a consolidated implementation from the same fork (branch fix/validate-test-set-type-consistency), which additionally enforces the persisted set type for direct bulk writers and will be opened as its own PR momentarily. No maintainer action needed on this one.

Issue #2396, opened 2026-08-07
Deactivated endpoint still shows as active in grid and detail page becomes unreachable


nicolai-rhesis:
## Summary
After setting an endpoint to inactive, the grid still shows it as active, but its detail page can no longer be opened.

## Steps to Reproduce
1. Go to Endpoints and open an endpoint's detail page
2. Set the endpoint to inactive and save
3. You are returned to the endpoints grid
4. Observe the status shown in the grid
5. Click the row to open the detail page again

## Expected Behavior
The grid reflects the new inactive status, and the detail page still opens.

## Actual Behavior
The grid still shows the endpoint as active (stale data), and clicking the row no longer reaches the detail page.

## Environment
- staging, v0.12.0 (`release/v0.12.0`)

## Impact
Two problems in one flow: a stale grid after save, and an inactive endpoint becoming unreachable.

## Acceptance Criteria (testable)
- [ ] Grid status updates immediately after deactivating an endpoint
- [ ] The detail page of an inactive endpoint still opens

## Additional Context
Found during Testing Day, 6 August 2026 ([testing sheet](https://app.notion.com/p/rhesis-ai/Testing-Day-August-2026-6-August-2026-Manual-Testing-Only-3b444bce5c6981aeaf61c4da97bfc7c4)).

feiiiiii5:
Implemented in #2501: saving an endpoint now invalidates the endpoint queries, so the grid shows the new status (Active / Inactive) immediately instead of serving the stale 5-minute list cache. Regression test included (pre-fix FAIL verified).

On the second symptom (detail page unreachable after deactivation): I traced every path on current main — backend GET /endpoints/{id} has no status filter and deactivation does not soft-delete; both frontend detail pages have no status gate and the grid row links by id. No code blocks an inactive endpoint's detail page today, so that half looks fixed since v0.12.0 (or was a side effect of the stale grid). Details in the PR body.

PR #2501, opened 2026-08-16
fix(frontend): refresh endpoints grid after save

feiiiiii5:
## Root Cause

Saving an endpoint (e.g. setting its status to Inactive) updated the endpoint's detail cache but never invalidated the endpoints **list** query. The grid query has a 5-minute `staleTime`, so returning to the grid kept showing the old Active badge until the cache went stale on its own.

## Fix

`saveFields` in `useEndpointDetail` now invalidates `endpointKeys.all()` after a successful save, so both the endpoints grid and the detail page refetch and reflect the persisted status immediately.

## Test

- New `useEndpointDetail.test.tsx` (2 cases):
  - a successful save calls `invalidateQueries({ queryKey: endpointKeys.all() })`
  - the just-saved change is still applied to the detail cache optimistically
- Pre-fix FAIL → post-fix PASS verified (invalidation case fails on `main`).
- Full `src/app/(protected)/endpoints` suite: 11 suites / 59 tests pass. `tsc --noEmit`, ESLint, Prettier clean.

## About the second symptom in #2396 (detail page unreachable)

I traced every code path a deactivated endpoint's detail page goes through on current `main` (backend `GET /endpoints/{id}` via `crud.get_endpoint`/`get_item_detail` — no status filter, no soft-delete on deactivation; frontend `useEndpoint`/`EndpointsClient.getEndpoint` and both detail pages — no status gate, row click link is `id`-based). There is no code that blocks an inactive endpoint's detail page today, so that half of the report appears to have been fixed since v0.12.0, or was a side effect of the stale grid data. The cache-invalidation fix here covers the remaining reproducible symptom.

## Diff Scope

- `apps/frontend/src/app/(protected)/endpoints/[identifier]/components/useEndpointDetail.ts` (4 lines)
- `.../components/__tests__/useEndpointDetail.test.tsx` (new)

### DIFF
--- a/apps/frontend/src/app/(protected)/endpoints/[identifier]/components/__tests__/useEndpointDetail.test.tsx
+++ b/apps/frontend/src/app/(protected)/endpoints/[identifier]/components/__tests__/useEndpointDetail.test.tsx
@@ -1,0 +1,110 @@
+import React from 'react';
+import { renderHook, waitFor } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import { useEndpointDetail } from '../useEndpointDetail';
+import { endpointKeys } from '@/constants/query-keys';
+import type { Endpoint } from '@/utils/api-client/interfaces/endpoint';
+
+jest.mock('next/navigation', () => ({
+  useRouter: () => ({ push: jest.fn(), refresh: jest.fn() }),
+}));
+
+jest.mock('next-auth/react', () => ({
+  useSession: () => ({
+    data: { session_token: 'tok', user: { id: 'u1', name: 'Alice' } },
+    status: 'authenticated',
+  }),
+}));
+
+jest.mock('@/components/common/NotificationContext', () => ({
+  useNotifications: () => ({ show: jest.fn(), close: jest.fn() }),
+}));
+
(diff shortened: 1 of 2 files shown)


EmanueleDeRossi1:
Hi @feiiiiii5 thank you for this fix! This solves the issue. I could not reproduce it with Status (it is just a non-editable field), but this happens with other fields. For instance, if you:

Change Name -> click Save -> The detail page shows the new name. But then when you go back to the grid, you would see that it display still the old Name column. 

This PR fixes that. I did a small audit, and found out this bug is present on other pages as well: Tests, Test Sets, Tasks, and Knowledge. I can follow up with a fix like this one. Thank you again!

feiiiiii5:
Thanks for verifying, and for the audit across the other pages! Glad it solves it. Agreed the same stale-grid pattern exists on Tests / Test Sets / Tasks / Knowledge — happy to leave those to your follow-up fix; keeping this PR scoped to the endpoints grid. Let me know if you'd rather I pick up any of the sibling pages instead.

EmanueleDeRossi1:
Hi @feiiiiii5, don't worry, I can think about it! I would be working on a bigger, platform-wide refactoring, and will fix this issue in the other pages as a side effect. Thank you for the offer though, and thank you for the other PRs you opened these days!

PR #2503, opened 2026-08-17
Fix OWASP metrics accuracy: calibrated rubrics, metric-level stats, and multi-turn context

alexeyhramov123:
## Purpose

This branch fixes a cluster of related gaps in OWASP LLM/Agentic Top 10 metric evaluation accuracy. The rubric text that judges grade against was too vague; aggregate metric stats collapsed per-metric outcomes into test-level pass/fail; multi-turn evaluations passed an empty context to context-dependent metrics; and Polyphemus errors during synthesis were surfaced as unhelpful or misleading messages.

## What Changed

- **Backend — OWASP judge rubric calibration:** Updated `evaluation_prompt`, `evaluation_steps`, `reasoning`, and `evaluation_examples` for all 20 OWASP LLM Top 10 + Agentic Top 10 `CategoricalJudge` metrics in `initial_data.json`. Also corrected `context_required` to `false` for ASI06 (Memory & Context Poisoning) and ASI08 (Cascading Failures). A data migration (`3f5954f6c374`) pushes the updated rubric text to existing orgs using a content-hash guard so user-customized rows are left untouched.
- **Backend — metric stats fix:** `v_metric_stats`'s `effective_success` was falling back to the test's overall pass/fail when a metric's own `is_successful` disagreed — which it did on every mixed-outcome test, dragging passing metrics into "failed" in aggregate. Migration `d52329dc7e4e` drops the fallback; `is_successful` is used directly. The now-dead `effective_metric_success` helper and its test are removed; replaced by `test_automated_metric_success.py` and a new `test_metric_stats_view.py` that exercises the real view.
- **Backend — multi-turn metric context:** `tasks/execution/evaluation.py` and `tasks/execution/batch/evaluation.py` hardcoded `context=[]` for multi-turn runs, so context-dependent metrics (LLM08, ASI06, ASI08) never saw actual retrieved chunks. Added `_collect_conversation_context()` to flatten all turns' context and wired it into both execution paths.
- **Backend — generation error labeling:** `generate_and_save_test_set` was rewriting any `ValueError` mentioning `"model"` into "check your Models settings" guidance — but nearly all LLM errors mention `"model"`, so transient failures (e.g. scale-to-zero timeouts) got mislabeled with wrong guidance. Narrowed the keyword match to specific config-error phrases.
- **Frontend — per-metric outcome filter:** Replaced the test-run detail drawer's metric checkboxes (filtering only on whether a metric ran) with a per-metric outcome selector (`all / evaluated / passed / failed`), so users can isolate exactly which metric outcome they want to inspect. Icon-only pills keep rows compact; a `MetricOutcomeLegend` explains the icons once at the top of the section.
- **Frontend — metric stats outcome fix:** `getEffectiveMetricSuccess()` mirrored the old backend fallback; updated to match the backend fix (use `is_successful` directly, drop the `test` parameter and fallback).
- **SDK — multi-turn batch error handling:** `MultiTurnSynthesizer._generate_batch` crashed with a bare `KeyError: 'tests'` when a model returned `{"error": ...}`. Added the same shape-check and retry logic (up to 3 attempts) that the single-turn path already had; failures surface a `ValueError` with the actual reason.
- **SDK — Polyphemus error surfacing:** Only `HTTPError` was caught, so connection errors and timeouts produced raw urllib3 stack traces. Now catches `RequestException` broadly and converts connection/timeout/5xx failures into a message telling the user the model may be scaling up.
- **SDK — OWASP topic prefixes:** Generated topics now include their risk code, e.g. `"ASI08: Cascading Failures"` instead of `"cascading failures"`, matching the category picker's existing labeling.

## Additional Context

- The metric stats fix and the frontend mirror are paired changes — both needed to agree on what "effective" metric success means.
- The rubric calibration + migration are paired: `initial_data.json` is only read at new-org onboarding, so the migration is required for existing orgs.
- The Polyphemus and multi-turn fixes are independent but were discovered together while testing OWASP generation on a cold-starting instance.

## Testing

- Run `pytest tests/backend/services/stats/` to verify the new metric stats view tests and check the removed fallback tests are gone.
- Run `pytest tests/sdk/synthesizers/test_multi_turn_synthesizer.py tests/sdk/models/providers/test_polyphemus.py` for SDK fixes.
- End-to-end: kick off an OWASP test run against a Polyphemus-backed endpoint that starts cold; confirm topics include risk-code prefixes, errors mention scale-to-zero, and per-metric outcome filters work in the test-run detail drawer.
- Check the metric stats cards on any existing test run with mixed metric outcomes — previously-failing metrics should now show green if their own verdict was passing.


harry-rhesis (commented):
Thanks for the extensive work here, Alexey. The rubric calibration is thorough, the metric stats fix is well reasoned, and the paired backend/frontend changes are cleanly structured. The commit history reads well too.

Two things to address before merging:

**`metric_scope` not migrated for existing orgs.** The `initial_data.json` changes add `"Multi-Turn"` to every LLM Top 10 metric's `metric_scope`, but migration `3f5954f6c374` only pushes rubric text fields (`evaluation_prompt`, `evaluation_steps`, `reasoning`, `evaluation_examples`, `context_required`). It does not update `metric_scope`. The execution path hard-filters metrics by scope (`filter_metrics_by_scope(metrics, MetricScope.MULTI_TURN, ...)` in `executors/metrics.py`), so existing orgs will still have `["Single-Turn"]` on these metrics and they'll be skipped during multi-turn evaluation. The multi-turn context fix (`_collect_conversation_context`) is correct, but existing orgs won't benefit from it for these metrics until `metric_scope` is also migrated.

**Hardcoded `#fff` in `MetricOutcomeRow`.** Two instances of `color: isSelected ? '#fff' : 'primary.main'` in `TestRunDetailFilterDrawer.tsx`. Should use `theme.palette.primary.contrastText` (or another theme token) per the frontend conventions.

Everything else looks solid. Inline comments below with specifics.

apps/backend/src/rhesis/backend/alembic/versions/3f5954f6c374_calibrate_owasp_metric_rubrics.py
```diff
@@ -0,0 +1,204 @@
...
+    return hashlib.sha256("\x00".join(parts).encode("utf-8")).hexdigest()
+
+
+def _new_metric_fields(metric_item: Dict[str, Any]) -> Dict[str, Any]:
+    return {
+        "evaluation_prompt": metric_item["evaluation_prompt"],
+        "evaluation_steps": metric_item.get("evaluation_steps"),
+        "reasoning": metric_item.get("reasoning"),
+        "evaluation_examples": metric_item.get("evaluation_examples"),
+        "context_required": metric_item.get("context_required", False),
```
harry-rhesis, line 139: `_new_metric_fields` updates the five rubric text fields but not `metric_scope`. The `initial_data.json` changes in this same PR add `"Multi-Turn"` to every LLM Top 10 metric's scope, but this migration won't push that change to existing orgs.

Since the execution path filters metrics by scope (`filter_metrics_by_scope(metrics, MetricScope.MULTI_TURN, ...)` in `executors/metrics.py:12`), existing orgs' LLM metrics will still be skipped during multi-turn evaluation. The `_collect_conversation_context` fix is correct but won't fire for these metrics until their scope includes `"Multi-Turn"`.

Suggestion: add `"metric_scope"` to the fields this migration pushes, or add a follow-up migration for it. The hash guard would need to be recalculated (or `metric_scope` could be updated unconditionally since it's additive, not a content edit a user would have customized).

apps/frontend/src/app/(protected)/test-runs/[identifier]/components/TestRunDetailFilterDrawer.tsx
```diff
@@ -193,6 +224,113 @@ function ActivityFilterRow({
...
+                    borderColor: 'primary.main',
+                    borderLeft: isFirst ? '1px solid' : 'none',
+                    borderRight: isLast ? '1px solid' : 'none',
+                    borderRadius: isFirst
+                      ? '999px 0 0 999px'
+                      : isLast
+                        ? '0 999px 999px 0'
+                        : 0,
+                    bgcolor: isSelected ? 'primary.main' : 'transparent',
+                    color: isSelected ? '#fff' : 'primary.main',
```
Comment (harry-rhesis) [inline apps/frontend/src/app/(protected)/test-runs/[identifier]/components/TestRunDetailFilterDrawer.tsx]: Hardcoded `'#fff'` here (and in the existing `CompactSegmentedPills` at line 170, which predates this PR). Per the frontend conventions, this should use a theme token like `theme.palette.primary.contrastText`.

PR #2510, opened 2026-08-17
Add basic Grafana alerts with Discord and email notifications

asad-miah:
## Purpose

We had no alerting configured in Grafana at all. This adds a starter set of basic, generic k8s health alerts across dev, stg, and prd, notifying both Discord and email, so failures stop going unnoticed until someone happens to check dashboards manually.

## What Changed

- Add a `GrafanaAlertRuleGroup` with 6 starter alerts: pod crash-looping, pod not ready, container OOMKilled, Prometheus scrape target down, TLS cert expiring soon (all cert-manager certs, not just Grafana's own), and PVC almost full.
- Add a shared contact point (`default-notifications`) with Discord and email receivers, and a notification policy routing all alerts to it.
- Discord webhook URL is per-cluster (dev/stg/prd), synced from GCP Secret Manager via a dedicated `ExternalSecret`, since each env can point at its own channel.
- Email reuses the app's existing SMTP credentials (`SMTP_HOST`/`SMTP_USER`/`SMTP_PASSWORD`), synced separately into the `monitoring` namespace. The sender address is a plain literal rather than reusing the app's `FROM_EMAIL`, since that value is formatted for the app's email library (`"Display Name" <email>`) and Grafana's `from_address` field rejects anything but a bare address.
- Alert folder, contact point, notification policy, and rule group live in `base/grafana-resources` since they're identical in every env; only the Discord webhook and SMTP secrets are per-cluster.
- Pinned the Prometheus datasource to a stable `uid: prometheus` so alert rules can reference it deterministically instead of an operator-generated UID.

## Additional Context

Deliberately out of scope for this PR (needs new ServiceMonitors first, tracked as follow-up work): Postgres/Valkey exporter metrics, ingress-nginx request/error-rate metrics, ArgoCD sync/health drift, and External Secrets Operator sync failures.

Manual prerequisite before this can sync successfully in stg/prd: create `stg-grafana-discord-webhook` and `prd-grafana-discord-webhook` in GCP Secret Manager (documented in `kubernetes/monitoring/PREREQUISITES.md`). No new secret is needed for email since it reuses existing app SMTP credentials.

## Testing

Tested end-to-end on dev by temporarily pointing the `dev-grafana-resources` ArgoCD Application at this branch (`argocd app set dev-grafana-resources --revision feature/grafana-dev-alerts` with `dev-base` auto-sync paused, then reverted after testing):
- Confirmed all 6 alert rules, the contact point, notification policy, and folder synced successfully (`AlertGroupSynchronized`/`ContactPointSynchronized` conditions `True`).
- Fired a test notification from the Discord integration and confirmed delivery.
- Fixed a crash caused by an invalid `use_discord_username` type (string vs bool) and a crash caused by an invalid SMTP `from_address` format, both caught during this manual test.
- Verified the contact point now has both `discord` and `email` receivers via Grafana's own provisioning API.
- Verified `kubectl kustomize` renders cleanly for all three environments (dev/stg/prd).


kubernetes/clusters/dev/external-secrets/rhesis-app-secrets.yaml
```diff
@@ -334,6 +334,10 @@ spec:
       remoteRef:
         key: grafana-datasource-password
 
+    - secretKey: GRAFANA_DISCORD_WEBHOOK_URL
```
harry-rhesis: `GRAFANA_DISCORD_WEBHOOK_URL` is synced into the app namespace here, but nothing consumes it. The webhook is already synced into the `monitoring` namespace via the dedicated `discord-webhook-secret.yaml`, which is what the GrafanaContactPoint actually reads.

This widens the blast radius of the secret (readable by any pod mounting the app secret) with no benefit. Same applies to the stg and prd copies. I'd remove all three.

kubernetes/base/grafana-resources/basic-alerts.yaml
```diff
@@ -0,0 +1,242 @@
...
+      execErrState: Error
+      data:
+        - refId: A
+          datasourceUid: prometheus
+          relativeTimeRange:
+            from: 600
+            to: 0
+          model:
+            datasource: { type: prometheus, uid: prometheus }
+            expr: up
```
harry-rhesis: Bare `up` with no label filter matches every scrape target in Prometheus, including internal ones (etcd, kube-scheduler, coredns, node-exporter, etc.) that transiently fail or aren't reachable depending on cluster setup. Labeled `severity: critical`, this will likely produce noise from targets that aren't operationally relevant.

Consider scoping to namespaces you care about:
```yaml
expr: up{namespace=~"default|monitoring"}
```
Or downgrade to `severity: warning` and increase `for:` to 10m to filter transient blips.

kubernetes/base/grafana-resources/basic-alerts.yaml
```diff
@@ -0,0 +1,242 @@
...
+            refId: B
+      labels:
+        severity: critical
+      annotations:
+        summary: "Pod {{ $labels.namespace }}/{{ $labels.pod }} has not been Ready for 10m"
+
+    - uid: rhesis-pod-oomkilled
+      title: Container OOMKilled
+      condition: B
+      for: 0m
```
harry-rhesis, line 103: `for: 0m` means this fires immediately. `kube_pod_container_status_last_terminated_reason{reason="OOMKilled"}` is a gauge that stays 1 until the container's termination status is cleared by a new restart. With a 5m evaluation interval, this will re-fire every interval for the same OOM event, generating repeated notifications even if the issue was already addressed.

Consider bumping to `for: 5m` to match the evaluation interval. Or, if the intent is truly instant notification, add a note explaining the trade-off.

kubernetes/base/grafana-resources/basic-alerts.yaml
```diff
@@ -0,0 +1,242 @@
...
+            type: threshold
+            expression: A
+            conditions:
+              - evaluator: { type: gt, params: [3] }
+                operator: { type: and }
+                query: { params: [] }
+                reducer: { type: last, params: [] }
+            refId: B
+      labels:
+        severity: critical
```
harry-rhesis, line 56: Five of six alerts are `severity: critical`. When almost everything is critical, the label stops being useful for triage.

Suggestion: reserve `critical` for crash-looping and OOMKilled (actively failing). Downgrade `scrape-target-down` (often transient), `pod-not-ready` (10m not-ready is often a rolling deploy), and `pvc-almost-full` (10% free is a "you should look" signal, not a pager) to `warning`. This also sets you up to route `critical` to Discord and `warning` to email-only later.

kubernetes/base/grafana-resources/contact-point.yaml
```diff
@@ -0,0 +1,25 @@
...
+        use_discord_username: true
+      valuesFrom:
+        - targetPath: url
+          valueFrom:
+            secretKeyRef:
+              name: grafana-discord-webhook
+              key: url
+    - type: email
+      settings:
+        addresses: hello@rhesis.ai
```
harry-rhesis: `hello@rhesis.ai` is hardcoded here and also in `grafana-instance.yaml` as `GF_SMTP_FROM_ADDRESS`. If this address changes, two files in different directories need updating. Not blocking, but worth being aware of.

harry-rhesis (commented):
Nice work, Asad. This is a solid starter alert set and the base/overlay structure is clean. A few things to consider below.

**What looks good:**
- Sync-wave ordering (ExternalSecrets at 0, folder + contact point at 1, rules + policy at 2) avoids race conditions.
- Pinning `uid: prometheus` on the datasource so alert rules can reference it deterministically is the right call.
- Secrets handled properly via GCP Secret Manager / ExternalSecrets with `creationPolicy: Owner`, nothing inline.
- Good docs in PREREQUISITES.md and clear inline comments explaining non-obvious decisions (like why `FROM_EMAIL` can't be reused).
- Thorough testing on dev, and catching the bool/from_address bugs during that test.

kubernetes/clusters/dev/external-secrets/rhesis-app-secrets.yaml
asad-miah: Good catch — confirmed nothing reads `GRAFANA_DISCORD_WEBHOOK_URL` from `rhesis-app-secrets` in any namespace. Removed all three copies (dev/stg/prd) in 7fe31167e. The webhook still syncs from the same GSM key via `discord-webhook-secret.yaml` directly into `monitoring`, which is what `GrafanaContactPoint` actually reads.

kubernetes/base/grafana-resources/basic-alerts.yaml
asad-miah: Confirmed this live on dev — coredns currently shows 2/2 targets down (GKE's managed CoreDNS doesn't expose its metrics port the way kube-prometheus-stack's default ServiceMonitor expects; DNS itself is fine). This would have paged `critical` on the very first evaluation after merge. Fixed in 7fe31167e: excluded `job="coredns"`, bumped `for` to 10m, and downgraded to `warning`.

kubernetes/base/grafana-resources/basic-alerts.yaml
asad-miah: You're right that this doesn't resolve on its own. Bumping `for:` alone wouldn't actually fix it though — `kube_pod_container_status_last_terminated_reason` stays 1 indefinitely, so it'd still re-notify every `repeat_interval` (4h) for a stale event. Went with joining against `increase(kube_pod_container_status_restarts_total[10m])` instead, so the alert self-clears ~10-15m after the actual OOM event rather than never. Fixed in 7fe31167e.

kubernetes/base/grafana-resources/basic-alerts.yaml
asad-miah: Agreed, done in 7fe31167e — kept `critical` on crash-loop and OOMKilled only, downgraded pod-not-ready, scrape-target-down, and PVC-almost-full to `warning`.

kubernetes/base/grafana-resources/contact-point.yaml
asad-miah: Fixed properly in 119003df4 rather than left as a known tradeoff — added a `grafana-alert-email` ConfigMap in `base/grafana-resources/kustomization.yaml` and pointed both `grafana-instance.yaml` (`GF_SMTP_FROM_ADDRESS`) and `contact-point.yaml` (email receiver) at it via kustomize `replacements`, same pattern already used for `grafana-hostname`/`grafana-root-url`. Now it's defined once.

asad-miah:
Addressed both leftovers from the approval review in 8963d523f:

- Fixed the stale "dev only" header comment in `basic-alerts.yaml` — leftover from before the file moved to `base/`.
- `pod-not-ready` now excludes pods in `Succeeded`/`Failed` phase (`unless on (namespace, pod) (kube_pod_status_phase{phase=~"Succeeded|Failed"} == 1)`), so a completed Job/CronJob pod lingering past its TTL won't false-positive as not-ready. No live evidence of this firing yet, but this cluster does run one-off Jobs (e.g. the license-issuer workflow via arc-runner-dev), so worth closing proactively.

PR #2512, opened 2026-08-18
Fix OWASP migrations before changing behaviors->requirement

EmanueleDeRossi1:
## Summary
- Fresh database migrations fail with `UndefinedTable: relation "requirement" does not exist`
- Two migrations (`38ed899b9f41`, `b857edcac3c0`) import the live `Requirement` ORM model instead of using raw SQL. #2487 renamed the `behavior` table to `requirement` and updated that model in place, but on a from-scratch DB these two migrations run *before* the rename  migration, when the table is still called `behavior` — so they crash querying a table that doesn't exist yet
- This never showed up on existing dev/prod databases since they already had the table from before the rename — it only breaks a full migration replay from an empty database
- Both migrations now skip their body when the `requirement` table isn't there yet, which is safe since fresh installs get OWASP behaviors/metrics seeded another way, through onboarding's `load_initial_data()`

## Test plan
- [ ] `alembic downgrade base && alembic upgrade head` on a fresh database completes without error
- [ ] Existing (already-migrated) dases upgrade to head with no behavior change

### DIFF
--- a/apps/backend/src/rhesis/backend/alembic/versions/38ed899b9f41_add_owasp_behaviors_and_metrics_to_.py
+++ b/apps/backend/src/rhesis/backend/alembic/versions/38ed899b9f41_add_owasp_behaviors_and_metrics_to_.py
@@ -29,6 +29,7 @@ Create Date: 2026-07-02
 from typing import Sequence, Union
 
 from alembic import op
+from sqlalchemy import inspect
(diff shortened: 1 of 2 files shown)


harry-rhesis:
@EmanueleDeRossi1 nice catch, thanks!

PR #2527, opened 2026-08-20
OWASP test generation: parity fixes and Architect integration

alexeyhramov123:
## Summary
Fixes 4 reported OWASP generation issues, plus 3 more found during a consistency audit against the sibling Garak feature:

**Originally reported:**
- Architect chat had no tool to generate OWASP or Garak test sets — registered both (`mcp_tools.yaml`).
- The test set table didn't pick up a completed OWASP generation without a manual reload — added cache invalidation on generation start, mirroring Garak.
- OWASP generation completion wasn't surfaced as an in-app notification, unlike Garak/plain generation — added the missing decorator.
- Generated test topics could drift from the canonical casing (e.g. "improper output handling" instead of "LLM05: Improper Output Handling") because the LLM's echoed `topic` field was trusted instead of the computed one — now overwritten deterministically for both single- and multi-turn generation.

**Found auditing OWASP against its Garak sibling for consistency:**
- OWASP generation wasn't gated by the test-generation quota, unlike every other generation endpoint — real quota-bypass bug, now gated.
- Garak's dynamic-probe generation had no `model_id` override, unlike OWASP and the plain generator — added end-to-end (schema, router, UI model selector).
- `GarakImportDrawer.tsx` actually hosted both Garak and OWASP flows despite its name; a dead standalone `OwaspGenerateDrawer.tsx` was never imported anywhere — renamed to `SecurityTestDrawer` and deleted the dead file.

Also added test coverage that was missing on the Garak side: OWASP router dispatch tests, OWASP service tests, and Garak tasks in the notification-decorator ordering guard.

## Test plan
- [x] Backend: full pytest suite green (`apps/backend`), including new dispatch/quota/service tests
- [x] SDK: `make test` green, including new topic-casing coverage
- [x] Frontend: jest, type-check, ESLint, Prettier, hardcoded-styles scan all green
- [x] Both community/EE-boundary checks green
- [x] `mcp_tools.yaml` schema validation green


harry-rhesis:
@alexeyhramov123 terrific work, thank you so much!

PR #2560, opened 2026-08-21
Add VolumeSnapshotClass for prd PVC snapshots

### LINKED ISSUE #1691: **Severity: MEDIUM — Reliability (production)**
**Severity: MEDIUM — Reliability (production)**

SingleBinary Loki with `replicas: 1`, `replication_factor: 1`, and filesystem storage is fine for dev/stg, but in production this is a single point of failure with no data durability beyond the PVC. If the node fails before data is flushed, recent logs are lost.

**Suggestion:** For a follow-up, consider switching `storage.type` to `gcs` with a dedicated bucket for prd — that gives durability independent of PVC lifecycle. At minimum, ensure PVC snapshots are enabled via a VolumeSnapshotClass.

_Originally posted by @harry-rhesis in https://github.com/rhesis-ai/rhesis/pull/1592#discussion_r3080038934_

asad-miah:
## Purpose

Loki runs as SingleBinary in prd with `replicas: 1`, `replication_factor: 1`, and filesystem storage, so it has no data durability beyond its PVC — if the node fails before data is flushed, recent logs are lost. This adds the minimum safeguard called out in the linked issue: a VolumeSnapshotClass so PVC snapshots are possible in prd.

## What Changed

- Added `kubernetes/clusters/prd/volumesnapshotclass.yaml`, a `VolumeSnapshotClass` (`pd-retain`) using the GKE persistent-disk CSI driver (`pd.csi.storage.gke.io`), marked as the cluster default, with `deletionPolicy: Retain` so deleting a snapshot object never deletes the underlying disk snapshot.
- Wired it into `kubernetes/clusters/prd/kustomization.yaml` so ArgoCD syncs it with the rest of the prd stack.

## Additional Context

- Closes #1691.
- Scoped to prd only, matching the issue's severity ("Reliability (production)") and the review comment it originated from. The bigger recommendation in that same comment — switching Loki's `storage.type` to `gcs` — is called out there as a separate follow-up, not part of this change.
- This only makes PVC snapshots *possible*; it doesn't itself schedule automatic snapshots. Actually snapshotting the Loki PVC on a schedule (e.g. via a CronJob or a backup tool) would be a further follow-up.

## Testing

Rendered the full prd kustomization locally to confirm it builds cleanly and the new resource appears as expected:

```bash
kubectl kustomize kubernetes/clusters/prd
```

### DIFF
--- a/kubernetes/clusters/prd/kustomization.yaml
+++ b/kubernetes/clusters/prd/kustomization.yaml
@@ -22,6 +22,7 @@ resources:
   # Monitoring stack (sync-wave on each wrapper defers grafana-resources until CRDs exist)
   - grafana-operator.yaml
   - kube-prometheus-stack.yaml
+  - volumesnapshotclass.yaml
   - loki.yaml
   - alloy.yaml
   - grafana-resources.yaml
--- a/kubernetes/clusters/prd/volumesnapshotclass.yaml
+++ b/kubernetes/clusters/prd/volumesnapshotclass.yaml
@@ -1,0 +1,11 @@
+# Makes PVC snapshots possible in prd (e.g. for Loki's filesystem storage, which has no
+# durability beyond its PVC). Retain so a deleted VolumeSnapshot never takes the underlying
+# disk snapshot with it.
+apiVersion: snapshot.storage.k8s.io/v1
+kind: VolumeSnapshotClass
+metadata:
+  name: pd-retain
+  annotations:
+    snapshot.storage.k8s.io/is-default-class: "true"
+driver: pd.csi.storage.gke.io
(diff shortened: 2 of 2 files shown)


harry-rhesis (approved):
Thanks, Asad. Clean, minimal, correct. The `deletionPolicy: Retain` is the right call for production, and removing the default-class annotation (addressing peqy's comment) avoids silently opting every other PVC into Retain-forever snapshots.

Good separation of concerns from the broader Loki durability work too. The PR description is clear that GCS storage is a separate follow-up and this only enables manual/scheduled snapshots.

No issues. Looks good.

PR #2604, opened 2026-08-26
feat(connector): persist execution traces instead of logging only

### LINKED ISSUE #2502: connector: execution traces are acknowledged but never persisted (existing TODO)
## Summary

The SDK connector endpoint for execution traces (`POST` trace receiver in `apps/backend/src/rhesis/backend/app/routers/connector.py`) validates project membership, logs the trace, and returns `{"status": "received"}` — but never persists it:

```python
# TODO: Store trace in database for analytics
# For now, traces are logged

return TraceResponse(status="received")
```

From the SDK caller's perspective the contract looks durable ("received"), but the trace only exists in server logs and is lost on rotation/restart. This reads as silent data loss for an observability feature: a user wiring `@connector`-style tracing for analytics gets a success response and then finds nothing to query.

Note this is distinct from the OTel telemetry pipeline: `models/trace.py` (`Trace` entity) is consumed by `services/telemetry/*` (enrichment/tree-builder) for span ingestion, while the connector's `ExecutionTrace` schema (function_name/inputs/output/duration_ms/status/error/project_id/environment) is a separate contract with no persistence path. Related-but-distinct open issues: #1794 (trace metrics), #1256 (trace cost), #1084 (orphaned span linking).

## Options

a) **Persist**: map `ExecutionTrace` onto the existing `Trace` entity (or a dedicated `execution_traces` table + Alembic migration) and store it inside the already-opened tenant db session; return the created id.
b) **Honest contract**: if persistence is intentionally deferred, make that visible — return `{"status": "received-not-stored"}` (or a documented flag) and downgrade the log to a warning-level note, so callers don't assume durability.
c) **Feature-flag**: persist when an analytics flag/plan is enabled, otherwise the honest response from (b).

Happy to send a PR once the intended direction is confirmed — filing first because the schema/tenancy choice (a) is a design call.

feiiiiii5:
Fixes #2502

## Motivation

`POST /connector/trace` validated project membership and returned `{"status": "received"}`, but the trace was never stored — the handler still had the original TODO. From the SDK caller's perspective the contract looked durable, while the data only existed in server logs and was lost on rotation/restart (silent data loss for an observability feature).

## Changes

Implements option **(a)** from the issue, with a dedicated table rather than mapping onto the OTel `Trace` entity (which requires span identifiers connector traces don't have):

- **New model** `ExecutionTrace` → `execution_trace` table: project FK (CASCADE), organization_id, environment, function_name, inputs (JSONB), output, duration_ms, status, error, executed_at — with indexes on the filterable columns.
- **Alembic migration** `b7e1c9d4a2f3` (chains on `f2b3c4d5e6a7`, single head verified).
- **Router**: inside the already-open tenant session, persist after the membership check and return the created id as an optional `trace_id` field on `TraceResponse` (backwards compatible — existing callers reading only `status` are unaffected). Persistence failures surface as request errors instead of being swallowed, so callers never get "received" without storage.
- **Tests**: two regression tests asserting the persisted row via `test_db` (fields incl. tz-aware `executed_at`, FK-valid project). The three legacy trace tests now use a seeded real project since the endpoint resolves `project_id` for storage.

## Testing

- `ruff check` clean on all touched files.
- `alembic heads` confirms a single head (`b7e1c9d4a2f3`) — migration chain valid.
- Model/schema import checks pass; full route tests require the Postgres test DB (`SQLALCHEMY_DATABASE_TEST_URL`), which I don't have locally — expecting CI to exercise them.

## Design notes

If persistence is meant to be plan/flag-gated (option c), the natural seam is wrapping the `db.add` block — happy to adapt.


apps/backend/src/rhesis/backend/alembic/versions/b7e1c9d4a2f3_add_execution_trace.py
```diff
@@ -0,0 +1,61 @@
...
+
+# revision identifiers, used by Alembic.
+revision: str = "b7e1c9d4a2f3"
+down_revision: Union[str, None] = "f2b3c4d5e6a7"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.create_table(
```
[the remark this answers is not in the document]
feiiiiii5: Addressed in 990e3a50a. The migration now creates the full Base-backed table — `id` (UUID PK with `gen_random_uuid()` server default), `nano_id`, `created_at`, `updated_at`, and `deleted_at` (timestamps timezone-aware to match `Base`), styled after the `c4d8e2f1a3b5_add_architect_tables` migration — so the DDL matches the ORM model instead of declaring `PrimaryKeyConstraint("id")` on a column it never created. I kept the model on `Base` since soft-delete/audit columns are what every other table here carries.

apps/backend/src/rhesis/backend/app/routers/connector.py
```diff
@@ -347,6 +349,21 @@ def receive_trace(
     with get_db_with_tenant_variables(organization_id, user_id, "") as db:
         _assert_project_membership(db, trace.project_id, current_user)
 
+        record = ExecutionTraceModel(
```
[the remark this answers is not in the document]
feiiiiii5: Addressed in 990e3a50a. `ExecutionTrace.project_id` is now typed `UUID` in the request schema, so a malformed id is rejected with a 422 by FastAPI validation before the handler runs. The handler passes `str(trace.project_id)` to `_assert_project_membership` (which keeps its own 400 path) and uses the validated `trace.project_id` directly for the model, dropping the now-redundant `uuid.UUID(...)` cast that was the 500 risk.

tests/backend/routes/test_connector.py
```diff
@@ -23,6 +26,21 @@
 _MEMBERSHIP_GUARD = "rhesis.backend.app.routers.connector._assert_project_membership"
 
 
+@pytest.fixture
```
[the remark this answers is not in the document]
feiiiiii5: Good catch — you were right to be suspicious: this PR has no test CI attached (the only checks are DCO and your review), so the 403 path had never actually executed. Addressed in 990e3a50a by taking the first option: `seeded_project` now also creates the authenticated user's `ProjectMembership` row, so the persistence tests exercise the real `_assert_project_membership` guard instead of needing a patch. The WebSocket/trigger tests keep patching the guard since those exercise connection-manager behaviour, not authorization.

harry-rhesis:
HI @feiiiiii5! Good to see you again :) Would you mind fixing the failing checks and rebasing your PR, so we can proceed with review? Thank you so much!

feiiiiii5:
All three red checks from Sep 3 are addressed; CI is green on this head.

**1. `migrations` (+ downstream `lint`): stale `down_revision`.** The branch pointed at `f2b3c4d5e6a7`, but main has since landed `ff71b040aebf -> c7e2f4a91b83 -> 84d6c1464b96 -> 105b0268e6a4 -> 42f4cf7014a0 -> b258958270d0` (verified linear, single head). Retargeted to `b258958270d0`; no merge migration needed. I also checked `105b0268e6a4` (trace execution/verdict columns) for overlap with this PR's `execution_trace` table — it only ALTERs `trace`, no conflict.

**2. `Backend tests` (5 connector failures): test session vs endpoint session.** `receive_trace` opens its own session via `get_db_with_tenant_variables` (direct call, not DI), so it could not see the `test_db`-only flushed `seeded_project` rows — 403 from the live membership guard where unmocked, FK violation on insert where mocked. Tests now share `test_db` through that same seam (`mock_db_context` precedent in the same file); everything stays inside the outer test transaction.

**3. `Backend tests` (4 RLS coverage failures): `execution_trace` had no RLS.** The migration now declares `ENABLE + FORCE ROW LEVEL SECURITY`, a PERMISSIVE `tenant_isolation` and a RESTRICTIVE `project_isolation`, bodies copied verbatim from `449d5364b4ff_create_notification_table` (including the `DROP IF EXISTS` guards and the note on the auto-apply trigger).

One production-relevant change bundled here: `receive_trace` now runs under project scope (`str(trace.project_id)`) instead of org-level `""`. `project_isolation` is fail-closed, so inserting this project-scoped row with an empty project scope would be denied by RLS outside tests.

feiiiiii5:
Hi @harry-rhesis — both done:

- The three failing checks are fixed (details in my comment above) and were green on the previous head.
- Branch is now updated onto current main (`e6e52c5`). I used a merge-update rather than a rebase; happy to switch to a rebased branch if you prefer linear history — just say the word.

CI is re-running on the merge; will watch it.

feiiiiii5:
Hi @harry-rhesis — CI is fully green on the updated branch (Backend 7m57s, migrations, lint, Java, DCO, peqy all pass). Ready for review whenever you are.

tests/backend/routes/test_connector.py
```diff
@@ -24,6 +27,49 @@
...
+            organization_id=uuid.UUID(test_org_id),
+        )
+    )
+    test_db.flush()
+    yield project
+    test_db.rollback()
+
+
+@pytest.fixture
+def shared_tenant_session(test_db):
```
harry-rhesis, line 55: This fixture solves a real problem, and the patch target is correct: the router imports `get_db_with_tenant_variables` into its own namespace, so patching it on the router module is the right seam.

The side effect is that replacing the entire context manager also skips `_apply_scope_variables`, so the RLS GUCs (`app.current_organization`, `app.current_project`) are never set on the session the handler actually uses. That means the `project_isolation` RESTRICTIVE policy added in the migration, and the reason for passing `str(trace.project_id)` instead of `""`, are not covered by any test here. If the project scope were wrong, these tests would still pass.

Might be worth one test that goes through the real `get_db_with_tenant_variables` against committed rows, so the fail-closed policy gets exercised at least once. Everything else can keep using this fixture.

apps/backend/src/rhesis/backend/app/routers/connector.py
```diff
@@ -344,8 +346,26 @@ def receive_trace(
...
 
-    with get_db_with_tenant_variables(organization_id, user_id, "") as db:
-        _assert_project_membership(db, trace.project_id, current_user)
+    # Project scope (not org-level "") is required here: project_isolation is a
+    # fail-closed RESTRICTIVE policy, so inserting this project-scoped row with
+    # an empty project scope would be denied by RLS in production.
+    with get_db_with_tenant_variables(organization_id, user_id, str(trace.project_id)) as db:
+        _assert_project_membership(db, str(trace.project_id), current_user)
+
+        record = ExecutionTraceModel(
```
harry-rhesis, line 355: `apps/backend/AGENTS.md` says layering is routers to services to crud, with no SQL in routers. The record construction plus `db.add` and `db.commit` is persistence logic living in the router.

For context, 13 other routers already commit directly, so this matches current practice rather than departing from it. Raising it only because a brand new entity is the cheapest moment to put this in `crud/execution_trace.py`, before a second caller (batch ingest, a retention sweep) needs the same construction.

apps/backend/src/rhesis/backend/app/routers/connector.py
```diff
@@ -344,8 +346,26 @@ def receive_trace(
...
+            function_name=trace.function_name,
+            inputs=trace.inputs,
+            output=trace.output,
+            duration_ms=trace.duration_ms,
+            status=trace.status,
+            error=trace.error,
+            executed_at=datetime.fromtimestamp(trace.timestamp, tz=timezone.utc),
+        )
+        db.add(record)
+        db.commit()
```
harry-rhesis: `get_db_with_tenant_variables` already ends its yield with `if db.in_transaction(): db.commit()`, specifically so deferred writes land while the tenant GUCs are still valid. With this explicit commit in place, the context manager's commit becomes a no-op.

Either one works, but only one is needed, and leaning on the context manager is the more common pattern here.

apps/backend/src/rhesis/backend/app/routers/connector.py
```diff
@@ -366,7 +386,4 @@ def receive_trace(
 
     logger.info("=" * 80)
 
-    # TODO: Store trace in database for analytics
-    # For now, traces are logged
-
-    return TraceResponse(status="received")
+    return TraceResponse(status="received", trace_id=str(record.id))
```
harry-rhesis: `record.id` is read after the `with` block has already closed the session. This works today because `SessionLocal` sets `expire_on_commit=False` (`database.py:259`), so the value stays in the instance dict instead of triggering a refresh against a closed session.

That is a quiet dependency on a session setting made elsewhere. Capturing the id inside the block would make this independent of it:

```python
        db.add(record)
        db.commit()
        trace_id = str(record.id)
```

apps/backend/src/rhesis/backend/app/schemas/connector.py
```diff
@@ -40,11 +41,14 @@ class ExecutionTrace(BaseModel):
     status: str = Field(..., description="Execution status: success or error")
     error: Optional[str] = Field(None, description="Error message if failed")
     timestamp: float = Field(..., description="Unix timestamp of execution")
-    project_id: str = Field(..., description="Project identifier")
+    project_id: UUID = Field(..., description="Project identifier")
```
harry-rhesis, line 44: Tightening this to `UUID` is the right call, and it moves validation to the edge where it belongs.

Just noting the one observable change: `_assert_project_membership` already raised 400 with "Invalid project_id: must be a UUID" for a non-UUID value, so those requests were rejected before as well. They now get a 422 from Pydantic instead. No caller that previously succeeded starts failing, but the status code changes, so anything asserting on 400 would need updating.

apps/backend/src/rhesis/backend/app/models/__init__.py
```diff
@@ -8,6 +8,7 @@
 from .comment import Comment
 from .embedding import Embedding
 from .endpoint import Endpoint
+from .execution_trace import ExecutionTrace
```
harry-rhesis: Two small things on this line.

The root `AGENTS.md` asks for absolute imports, including inside a package's own `__init__.py`. The rest of this file is still relative, and the rule is to convert as files get touched:

```python
from rhesis.backend.app.models.execution_trace import ExecutionTrace
```

Separately, `ExecutionTrace` is not added to the `__all__` list below. The import is what registers the model with `Base.metadata`, so nothing breaks, but every other model is listed there.

harry-rhesis (commented):
Thanks for putting this together, and for the care that clearly went into it. Persisting the connector traces instead of only logging them is a solid improvement, and a few details stood out: the explicit RLS policies in the migration with the note about `auto_apply_rls_policies` not firing reliably for a brand new CREATE TABLE, the switch from an empty project scope to `str(trace.project_id)` with a comment explaining why the fail-closed RESTRICTIVE policy needs it, and the `shared_tenant_session` fixture that solves the split-session problem the new tests would otherwise run into.

I also verified the migration chain: `b258958270d0` is the single current head of `main` (254 revisions, nothing else claims it as a parent), so this extends the chain cleanly with no fork.

Nothing here is blocking. One item is worth a look before merge, the rest are small.

- **Test coverage:** mocking the whole tenant session also skips the RLS setup, so the policy this change exists for is never exercised.
- **Minor:** persistence sits in the router rather than a service or crud module, one of the two commits is redundant, and `record.id` is read after the session context closes.
- **Nits:** relative import and a missing `__all__` entry in `models/__init__.py`.

One heads up that is not a defect: moving `project_id` from `str` to `UUID` changes the rejection status for a non-UUID value from 400 to 422. `_assert_project_membership` already rejected those, so nothing that previously succeeded starts failing, but the status code moves.

feiiiiii5:
Thanks for the detailed pass — all six addressed in `b4b7826` (CI running, will confirm green here):

1. **RLS hole in the fixture (scope never set): agreed, real gap.** Added `test_receive_trace_persists_through_real_tenant_session`: commits project + membership via `real_commit_test_db`, posts through `real_commit_client` with no session patching, so the endpoint's own `get_db_with_tenant_variables` sets the GUCs on a separate connection and `project_isolation` actually enforces. BEFORE: every persist test ran inside the shared session with no GUCs, so a wrong scope would still pass. AFTER: a wrong scope fails closed. explicit cleanup in `finally` (traces, membership, project).
2. **Router vs `crud/`:** fair point and the timing argument is right. I kept it in the router for now to match the 13 existing routers that commit directly and to keep this diff reviewable — happy to move construction into `crud/execution_trace.py` as a follow-up commit here if you prefer. Which way?
3. **Double commit:** fixed — removed the router `db.commit()`. Now `db.flush()` + capture; the CM owns the commit (verified it commits `if db.in_transaction()` at exit).
4. **`record.id` after close:** fixed — `trace_id = str(record.id)` captured inside the block right after flush. Needed the flush because the PK is `server_default gen_random_uuid()`, so the id only materializes on flush; no more reliance on `expire_on_commit=False`.
5. **UUID 400 → 422:** noted, no behavior loss. The only test asserting 400 on a non-UUID (`test_trigger_test_invalid_project_id`) hits `/connector/trigger`, a different endpoint/schema untouched by this change; nothing asserts 400 on the trace path.
6. **Import + `__all__`:** fixed — absolute import, and `ExecutionTrace` added to `__all__` (next to `Trace`).

feiiiiii5:
Small correction to my comment above: on `b4b7826` the CI runs are queued as `action_required` (fork workflow approval) rather than running, so they will go green once the workflows are approved for this fork. The code side of the six review points is complete; nothing else is pending on my end.

harry-rhesis:
Thanks for the thorough follow-ups, @feiiiiii5. Everything from the review looks addressed: the flush-and-capture pattern, the project scope on the tenant session, the real-commit RLS test, and the `__all__` entry.

On the **router vs crud/** question: keeping it in the router for now is fine. A follow-up is the right time for that move, not this diff.

One small thing I noticed: the import in `models/__init__.py` is still relative (`from .execution_trace import ExecutionTrace`). The rest of the file is relative too, so it's consistent today, but since you mentioned switching it to absolute, just flagging it's still the dotted-relative form. Not blocking.

CI is green except for the backend tests still pending (fork workflow approval). Once those pass, this looks good to merge.

harry-rhesis (approved):
Everything looks good. Thanks for the careful work on this, @feiiiiii5, and for addressing every review point so thoroughly. Nice contribution.

PR #2608, opened 2026-08-26
Retire useGridQuery and useGridState in favor of list descriptors

EmanueleDeRossi1:
## Purpose

Annotations, Jobs, Traces, and the embedded test-set tests grid were the last grids on the react-query hooks (`useGridQuery`/`useGridState`) instead of list descriptors. This converts all four to `useList`, so every grid in the app fetches through the one hook, and deletes the old hooks and everything that only existed to serve them.

## What Changed

- `usePaginatedList` gains silent live polling (`pollMs`): refreshes triggered by the poll timer skip the loading flag, so Jobs keeps its 3s refresh while jobs are running without the grid flashing its loading overlay.
- New descriptors: `annotationsList` (plain REST params via `extraParams`, gated on the same `TestResult.READ`/`Telemetry.READ` pair as the page), `jobsList` (drawer filters as specs, the Active pill's status-list clause via `extraFilters`), `tracesList(scopedProjectId)` (a factory closing over the active project, adapting the telemetry response to the standard paginated shape), and `testSetTestsList(testSetId)` (reuses the main tests page's exported `TESTS_FILTERS` over the nested endpoint, fixed `topic asc` sort).
- `getJobs` returns the standard `PaginatedResponse` instead of a flattened `{data, totalCount}`; `buildTraceQueryParams` drops its `limit`/`offset` params since pagination is the hook's job; the traces refresh FAB bumps a `refreshTrigger` prop instead of invalidating react-query keys.
- Deleted: `useGridQuery`, `useGridState` (and its tests), `DEFAULT_GRID_SORT`, `combineJobFiltersToOData`, the `annotationKeys`/`traceKeys` query keys, and three cross-page `annotationKeys.all()` invalidations that became no-ops once the annotations grid stopped using react-query.

## Additional Context

- Stacked on #2607, which is stacked on #2606. Net -481 lines.
- Also fixes the tests descriptor's stale "Not SSR-prefetched" comment left over from the SSR PR.

## Testing

- `npx tsc --noEmit`, ESLint, Prettier, and `next build` are clean.
- Full jest suite passes (227 suites, 2110 tests); trace-filter-params tests updated for the new signature.
- Manual: Jobs live-polls while a job runs without flashing; Annotations/Traces filter and paginate as before; the test set detail grid filters tests and reports its unfiltered total.


EmanueleDeRossi1:
All three of peqy's comments do not apply:

1- Session, features, and permissions are all SSR-seeded, so sessionStatus/enabled don't flip on hydration and no refetch fires. (That code is also from #2607, not this PR.)
2. GridStateGate only checks `data == null`, never the shape — and `loading ? null : {}` is the same idiom all the other grids use.
3. Client factories carry no credentials (auth comes from the BFF proxy); the new factory only adds the project-scope header, which is intentional. Traces isn't SSR-prefetched.

PR #2614, opened 2026-08-26
Unify list grids behind a single EntityGrid component

EmanueleDeRossi1:
## Purpose

Every list grid hand-assembled the same ~10 pieces (fetch hook, toolbar context, filter drawer state, error alert, bulk delete, actions column, empty state), and Explorer fetched differently altogether. This PR centralizes all of it in one `EntityGrid` component so grids look and behave the same, every grid gets a trailing Actions column with delete, and the duplicated code is removed. Net −3,000 lines.

## What Changed

- New `EntityGrid` component composing fetch, toolbar, filters, selection/bulk delete, actions column, and empty state; `ListDescriptor` gains a `delete` spec.
- All 12 list grids migrated (tasks, endpoints, sources, jobs, annotations, experiments, test runs, test sets, tests, tokens, explorer, team members). Reviewable one commit per grid.
- Backend: explorer list endpoint gains `$filter` + `X-Total-Count`; explorer and tokens grids move to real server-side pagination/filtering (removes explorer's silent 100-row cap).
- Removed ~600 lines of dead BaseDataGrid props and the 12 bespoke per-grid toolbar contexts.

## Additional Context

- Visible changes: test runs' run-kind toggle reappears (was silently broken), explorer selection moves behind the standard toggle with export/delete as page actions, sources/experiments load errors become the standard dismissible alert, team members is now a DataGrid instead of a plain table.
- Grids no longer unmount during refetches, fixing search-box focus loss while typing.

## Testing

All 232 frontend suites pass (2,194 tests, incl. new EntityGrid suite), `tsc` and ESLint clean; backend explorer route tests pass (93, incl. new `$filter`/count-header cases). Smoke-test the explorer, tokens, and team members pages via `./rh dev`.

### DIFF
--- a/apps/backend/src/rhesis/backend/app/crud/explorer.py
+++ b/apps/backend/src/rhesis/backend/app/crud/explorer.py
@@ -40,6 +40,15 @@ logger = logging.getLogger(__name__)
 # --- Test sets ---------------------------------------------------------------------
 
 
+def only_explorer_test_sets(query):
+    """Row selection for the Explorer list: test sets flagged via ``explorer_row``.
+
+    Shared with the route's ``with_count_header`` so X-Total-Count counts exactly
+    the rows the list returns.
+    """
+    return query.filter(models.TestSet.explorer_row.is_(True))
+
+
 def get_explorer_test_sets(
     db: Session,
     organization_id: str,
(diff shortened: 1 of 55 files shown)


EmanueleDeRossi1:
Fixed: `useBulkDelete` now normalizes ids with `.map(String)` instead of the as `string[]` cast. Leaving the rest as-is: the `serverSort` and `response` param notes are already resolved/confirmed, and the tokens client-clock filter is deferred to a follow-up since it needs a server-side filter key.

PR #2615, opened 2026-08-26
Server-prefetch the remaining list and detail pages

EmanueleDeRossi1:
## Purpose

Finish the server-prefetch rollout started in #2609. Every page that still fetched on mount showed a spinner or a grid skeleton that then swapped to an empty state (e.g. Tools, Traces) — the first paint should already have the data.

## What Changed

- List pages: Tools, Models, Projects, Tokens, Architect, Explorer, Traces and the Organization Settings team tab (when `?tab=team`) prefetch page 1 on the server and pass `initialData` to their client component.
- Detail pages: Job, Task, Experiment, Metric, Endpoint and project-scoped Endpoint fetch the entity on the server; 404 goes to the shared not-found page.
- Insights resolves the endpoint with the same cookie/project rule as the client, hydrates the endpoint list into react-query, and prefetches default-filter insights. Usage seeds `UsageProvider` via a new `initialUsage` prop.
- Traces: `createServerApiFactory` already reads the active-project cookie, so the old "can't prefetch, project scope is client state" exception is gone. The wrapper decides the empty state on first render.
- Removed the `isAuthenticated(status) -> setLoading(false)` trap that flashed an empty layout while the session was still resolving, plus dead loading branches and the `useIsCustomMetric` hook.

## Additional Context

- Follows the pattern from #2606–#2609. Remaining client-only pages are forms, Playground and Onboarding, which have nothing to prefetch.
- Small behaviour shifts: non-rhesis/custom metrics redirect to `/metrics` on the server instead of a client toast; a 404 on first load of a detail page lands on the shared not-found page.

## Testing

- `tsc`, eslint and the full frontend jest suite (231 suites) pass.
- Manual: open Tools and Traces on an empty project — the empty state should render directly, no spinner/skeleton. Check Insights, the metric detail/tuning page and a job detail page load with content on first paint.


EmanueleDeRossi1:
About the [Critical] point by peqy: this isn't a bypass. The backend checks `job:read` / `task:read` on these routes, and the server fetch uses the user's own token, so a user without the capability gets a 403 and no data ends up in the RSC payload. `hasServerCapability` is only a UX shortcut, not the security boundary.

All other detail pages (endpoints, test-sets, tests, metrics, …) work the same way, so I'll keep jobs/tasks consistent with them.

PR #2635, opened 2026-08-27
Fix Tests tab refresh/sort and add remove-from-test-set

EmanueleDeRossi1:
## Purpose

Two user-reported bugs on a test set's Tests tab: a test assigned via the Assign drawer doesn't show up in the grid until you leave and re-enter the tab, and there was no way to remove a test from a test set once it was added.

## What Changed

- `GET /{test_set_identifier}/tests` now accepts `sort_by`/`sort_order` (was `order_by`/`order`), matching every other list endpoint and what the frontend actually sends. The old mismatched names meant sort requests were silently dropped and the endpoint always fell back to `created_at desc` — this was also the root cause of tests appearing unsorted.
- Migrated the Tests tab grid (`TestSetTestsGrid`) from a bespoke `BaseDataGrid` setup to the shared `EntityGrid` component, the same pattern already used by the main Tests page and the Annotations grid.
- Wired a `refreshTrigger` from the Assign flow down to the grid so a successful assign now triggers a real refetch instead of a no-op cache invalidation that nothing was subscribed to.
- Fixed the grid's column field names (`topic.name`, `requirement.name`, `category.name`) to match what the backend's sort whitelist actually accepts, and enabled real interactive server-side sorting (previously sorting was a client-only no-op against server-paginated data).
- Added a "Remove from test set" row action, backed by the existing (previously unused) `POST /test-sets/{id}/disassociate` endpoint — added the missing frontend client method and request/response types for it.

## Additional Context

None.

## Testing

- `tsc --noEmit`, ESLint, and Prettier all clean on the changed files.
- Full frontend Jest suite passes (248 suites / 2426 tests).
- Manually verified against a running dev instance: assigning a test now shows up in the grid immediately, column headers sort interactively against the server, and the new remove action successfully disassociates a test and updates the grid/count.

### DIFF
--- a/apps/backend/src/rhesis/backend/app/routers/test_set.py
+++ b/apps/backend/src/rhesis/backend/app/routers/test_set.py
@@ -510,8 +510,8 @@ def get_test_set_tests(
     response: Response,
     skip: int = 0,
     limit: int = 10,
-    order_by: str = "created_at",
-    order: str = "desc",
+    sort_by: str = "created_at",
+    sort_order: str = "desc",
     filter: str | None = Query(None, alias="$filter", description="OData filter expression"),
     db: Session = Depends(get_tenant_db_session),
     current_user: User = Depends(require_current_user_or_token),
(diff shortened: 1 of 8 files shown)


EmanueleDeRossi1:
Both suggestions don't apply: the param rename has no real caller (SDK and docs never used `order_by`/`order`), and `total_tests` is the size of the removal request, not the remaining test count. Using it would show `1` instead of the real total.

PR #2657, opened 2026-08-30
fix(features): document the no-tenant-context route and cover the no-org path

### LINKED ISSUE #2653: Align get_current_organization_optional with tenant-context org resolution
## Summary

`get_current_organization_optional` resolves the caller's organization from `current_user.organization_id` using a non-tenant-scoped session, while `get_current_organization` resolves it from `get_tenant_context`. Two dependencies answer the same question by different routes, and `GET /features` switched from one to the other.

## Background / Context

Found while fixing nine `GET /features` test failures on `release/v0.14.0` (branch `fix/features-tests-mocked-org`). The endpoint moved to `get_current_organization_optional` so it can degrade gracefully for users mid-onboarding instead of returning 403. `tests/backend/routes/test_features.py` stubbed the tenant pair but not `get_db_session`, so a `Mock` organization id reached Postgres and psycopg2 rejected the UUID cast. The fixture is fixed on that branch, but the divergence is what let the drift go unnoticed.

`get_current_organization`'s docstring states the invariant the other one does not follow:

> the organization id always comes from the authenticated tenant context, never user-supplied input

There is a second, sharper inconsistency. `test_license_info_reflects_org` asserts that the license provider is "always called with the org, never None", while the endpoint now passes whatever the optional dependency returns, which is `None` for a user with no org yet. `FeatureRegistry.license_info(org=None)` and `QuotaRegistry.get_limits(None)` are both reachable in that state, and nothing covers it.

Note that `apps/backend/src/rhesis/backend/app/routers/features.py` currently differs between `main` and `release/v0.14.0`, so this needs attention when the release branch merges back, in whichever direction that resolves.

## Goals

- One documented way to resolve the caller's organization, whether or not an org exists yet.
- Settle whether `license_info` / `get_limits` may receive `None`, and make the test and the endpoint agree.

## Deliverables

- Either `get_current_organization_optional` reworked to take the org id from the tenant context, or a comment at the definition explaining why it cannot (for example, the tenant-scoped session being unusable before an org exists).
- A test covering `GET /features` for a user with no organization, which is currently untested.

- [ ] Both dependencies agree on where the organization id comes from, or the difference is documented where they are defined.
- [ ] The no-org path has a test, and the `never None` expectation is either enforced or dropped deliberately.
- [ ] Backend suite green on `release/v0.14.0`.

feiiiiii5:
## Root Cause

`get_current_organization` and `get_current_organization_optional` answer the same question ("which org is the caller in?") by different routes: the strict one goes through `get_tenant_context` (403-ing org-less users by design), the optional one reads `current_user.organization_id` on a plain session — and nothing at the definitions explained why, or whether the org id could be user-supplied. The divergence surfaced in `GET /features` (issue #2653): the endpoint degrades gracefully for mid-onboarding users, but `test_license_info_reflects_org` asserts the license provider is "always called with the org, never None", and the no-org path had no test at all.

## Fix

Take the **document-the-difference** route (the optional variant cannot route through the tenant context — mid-onboarding users are exactly its constituency, and `get_tenant_context` raises 403 when `organization_id` is missing):

- `get_current_organization_optional` now documents why the tenant-context route is unavailable, and why the security invariant on `get_current_organization` still holds in substance: the org id comes from the same authenticated user record (both dependencies deduplicate `require_current_user_or_token`), never from user-supplied input; the plain session is safe because the lookup is a primary-key read, not a tenant-filtered query.
- Documents that downstream consumers accept `None` by design (default providers answer it with the community/free-tier state).

## Test

- [x] New `test_no_org_degrades_to_community`: a user with `organization_id=None` calls `GET /features` → 200, `license == {"edition": "community", "licensed": False}`, `enabled == []`, `warnings == {}`, `limits == limits_to_wire(FREE_TIER_LIMITS)`; `get_tenant_context` is intentionally left un-overridden to prove the route never reaches it. (No local Docker — validated by CI on this PR with the real Postgres service.)
- [x] `test_license_info_reflects_org` (org-bearing path) unchanged and still green — the "never None" invariant stays enforced for the with-org path; the no-org path is now explicitly the designed `None` behavior.

## Diff scope

2 files, +55/−5: dependency docstring + one test. No runtime behavior change.

## AI Disclosure

AI-assisted implementation and test authoring; the design decision (document-the-difference rather than rework, per the issue's own option framing), the None-behavior analysis of `DefaultLicenseProvider`/`DefaultQuotaProvider`, and the final diff were human-reviewed before submission.

## Not PR-related failures

None expected; the route tests run against the CI Postgres service (no local Docker available for a pre-run).

Fixes #2653

### DIFF
--- a/apps/backend/src/rhesis/backend/app/dependencies.py
+++ b/apps/backend/src/rhesis/backend/app/dependencies.py
@@ -337,8 +337,18 @@ def get_current_organization_optional(
     degrade gracefully for users mid-onboarding (no org created yet)
     instead of 403-ing.
 
-    Uses a plain (non-tenant-scoped) session since we only read the org
-    by primary key, not run tenant-filtered queries.
+    Why this cannot resolve the org through the tenant context (see
+    :func:`get_tenant_context`): users mid-onboarding are exactly the
+    constituency here, and the tenant context raises 403 when
+    ``organization_id`` is missing -- routing through it would reintroduce
+    the 403 this dependency exists to avoid. The org id still comes from the
+    same authenticated user record (both dependencies deduplicate
+    ``require_current_user_or_token``), never from user-supplied input, so
+    the security invariant documented on :func:`get_current_organization`
+    holds in substance. The plain (non-tenant-scoped) session is safe
+    because we only read the row by primary key, not run tenant-filtered
+    queries. Downstream consumers accept ``None`` by design: the default
+    providers return the community/free-tier state for an org-less user.
     """
     if not current_user.organization_id:
         return None
--- a/tests/backend/routes/test_features.py
+++ b/tests/backend/routes/test_features.py
@@ -22,7 +22,7 @@ from rhesis.backend.app.features import (
 )
 from rhesis.backend.app.main import app
 from rhesis.backend.app.models.organization import Organization
-from rhesis.backend.app.quota import QuotaResource
+from rhesis.backend.app.quota import FREE_TIER_LIMITS, QuotaResource, limits_to_wire
 
 _TEST_ORG_ID = UUID("00000000-0000-0000-0000-000000000000")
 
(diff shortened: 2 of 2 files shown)


harry-rhesis (commented):
Thanks for the clean PR, @feiiiiii5. Two files, clear scope, good test coverage of the no-org path. Two suggestions:

**1. Dead `org_stub` setup (conciseness)**

`org_stub` and `db_stub.get.return_value = org_stub` in `test_no_org_degrades_to_community` are never reached. `organization_id=None` causes `get_current_organization_optional` to return `None` before `db.get()` is called. Removing those two lines makes the test's actual coverage (the early-return path) immediately obvious.

**2. Make the "never reaches tenant context" invariant explicit (completeness)**

The comment says `get_tenant_context` is intentionally not overridden, but the enforcement is implicit (the real dependency would fail without a DB). If `get_tenant_context` is ever refactored to not need a session, the test would silently stop catching accidental use of that path. Consider overriding it with something like:

```python
app.dependency_overrides[get_tenant_context] = Mock(
    side_effect=AssertionError("route must not resolve tenant context")
)
```

That turns the design invariant into a machine-enforced guard.

harry-rhesis:
Thanks for the solid work on this, @feiiiiii5. I've retargeted the PR to `main` (was `release/v0.14.0`). There are merge conflicts now, so when you get a chance could you rebase your branch onto `main` and force-push? That should clear them up.

feiiiiii5:
Rebased onto current main and force-pushed (branch now contains only the 2-file fix, release history dropped). Also addressed the earlier review on test_no_org_degrades_to_community: removed the unreachable org_stub setup and made the never-reaches-tenant-context invariant explicit (override raises + assert db.get not called). Let me know if you would like anything else adjusted.

feiiiiii5:
DCO fixed: all three commits now carry sign-off (DCO check green). This push also keeps the is_paid license assertion fix; tree is otherwise unchanged since the rebase. Ready for your review whenever suits.

harry-rhesis:
Great work addressing the review feedback, @feiiiiii5. The explicit tenant context guard and the `assert_not_called()` on `db_stub.get` are nice touches. Clean and thorough. Merging now.

PR #2666, opened 2026-08-31
fix(frontend): distinguish editable fields from read-only in dark mode

harry-rhesis:
## Summary

- In dark mode, `EditableField` text fields looked identical in read and edit mode because the global theme applies the same `fieldSurface` background to all outlined inputs, and the border color (`#30363d`) has near-zero contrast against that fill (`#2a2e36`).
- Added `editableOutlinedFieldSx` that sets `backgroundColor: transparent` on the input root when in edit mode, so the card surface shows through and the outlined border becomes visible.
- Read mode keeps the filled background with no border (unchanged).

## Test plan

- [ ] Open Settings > Personal Information in dark mode
- [ ] Confirm read mode fields show a subtle fill with no border
- [ ] Click Edit and confirm fields now show a visible outlined border against the darker card surface
- [ ] Verify light mode still looks correct (no visual regression)
- [ ] Check the change-password drawer fields are unaffected

### DIFF
--- a/apps/frontend/src/components/common/EditableField.tsx
+++ b/apps/frontend/src/components/common/EditableField.tsx
@@ -2,7 +2,10 @@
 
 import React from 'react';
 import { TextField, type TextFieldProps } from '@mui/material';
-import { readOnlyOutlinedFieldSx } from '@/components/common/drawerFormFieldSx';
+import {
+  editableOutlinedFieldSx,
+  readOnlyOutlinedFieldSx,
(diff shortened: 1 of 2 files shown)


harry-rhesis:
No need to scope it to dark mode. The global theme only sets `backgroundColor: fieldSurface` on `MuiOutlinedInput` inside a `mode === 'dark'` guard (`theme.ts:610-613`), so in light mode there's no background to override. The `transparent` value is effectively a no-op there.

PR #2679, opened 2026-09-01
Give server-rendered navigations immediate feedback

harry-rhesis:
## Problem

Since large parts of the app moved to server-side rendering, clicking a link gives no feedback
until the server finishes rendering. The click lands, nothing happens, then the new page appears.
There were no `loading.tsx` files anywhere and no navigation indicator, so every navigation felt
like a dead click.

During client-side navigation Next.js does not re-run layouts, only `page.tsx`, so the wait is
whatever the page's own server fetches take.

## What changed

**Navigation progress bar.** `nextjs-toploader` renders a thin primary-coloured bar as soon as a
navigation starts. This covers every route with no per-page work and is the main feedback
mechanism.

**Route loading skeletons.** `loading.tsx` wraps `page.tsx` in a Suspense boundary so a placeholder
paints while the page's server fetches run. Two shared skeletons cover the dominant page shapes.
Their geometry is copied from the real components (`PageLayout`, `GridToolbar`, `BaseDataGrid`,
`DetailTabNav`, `Fab`) so the swap to content does not shift layout: 56px FABs at `FAB_GROUP_GAP`,
30px toolbar and column insets, 52px rows, and pill shapes in the columns that hold chips.

The catch-all under `(protected)` serves every list page. Detail routes override it with the detail
shape, `architect` uses a centred spinner because it is full-bleed, and `test-sets` bumps the FAB
count to 4.

**Skeletons are delayed by 500ms.** A skeleton predicts the shape of content that has not arrived,
and the prediction is worst on a page whose list is empty: six rows of fake table dissolve into an
onboarding card. Empty pages are also the fastest, since there is no data to fetch, so they are
exactly the pages that never needed a skeleton. Every fallback is held invisible for 500ms and then
fades in, and the progress bar covers that window.

This is the delayed-indicator pattern that TanStack Router ships as `defaultPendingMs` and Ant
Design as `Spin`'s `delay` prop. 500ms sits between the ~300ms floor below which a delay is
imperceptible and the 1s mark where users notice one. The pattern's other half, a minimum display
duration, is deliberately absent: React owns when it unmounts a `loading.tsx` fallback, so
honouring a minimum would need hand-rolled Suspense boundaries and client timers on every route.
The fade covers that case instead, since a fallback unmounted mid-fade never reaches full opacity.

**EE components use MUI deep imports.** Barrel imports pull the full CJS barrel, which eagerly
evaluates `useMediaQuery`, and Next's barrel optimiser does not reach `transpilePackages` sources.
Not load-bearing (the underlying crash is fixed by `patches/@mui+material+7.3.5.patch`), just the
packaging MUI recommends. Drops out cleanly if you would rather keep this PR narrower.

**Installs now fail when a patch does not apply.** `patch-package` only exits non-zero in CI;
locally it warns and exits 0. That is how `patches/@mui+material+7.3.5.patch` went missing during
this branch, taking the whole app down with
`unstable_createUseMediaQuery is not a function` while the install that broke it looked clean. The
patch defers a client-only call out of module evaluation so `@mui/material` can be evaluated in the
RSC server graph. `--error-on-fail` applies the CI behaviour everywhere; verified against a
deliberately unappliable patch, exit 0 before and exit 1 after.

## Deliberately not included: streaming the protected layout

An earlier version of this branch moved the protected layout's features, permissions and terms
fetches into a Suspense-wrapped server component, with a fallback that mounted
`ProtectedLayoutClient` seeded with `null`. It has been dropped after review.

The fallback duplicated the entire provider tree *and* `{children}`. Those providers all fetch when
their seed is null (`FeaturesProvider` only passes `initialData` for a non-null seed,
`NotificationsProvider` fetches its summary on mount, `WebSocketProvider` mints a token per connect,
`TermsAcceptanceGate` fetches when unseeded), so if the client hydrates the fallback before the
boundary resolves, all of that runs twice, and the page subtree sits in both branches.

Whether it double-runs is timing dependent, and the bad case is a slow server, which is exactly the
case the change was meant to improve. Next.js also does not re-run layouts on client-side
navigation, so the change only ever affected cold loads, not the sidebar-click case this PR is
about. Narrow benefit, real cost.

Doing it properly means passing the promises into a stably mounted `ProtectedLayoutClient` and
unwrapping them with React 19's `use()`, so the providers mount once and stream their data in. That
changes the provider seeding contract and belongs in its own PR.

## Testing

- `next build --webpack` passes on the rebased branch
- `tsc --noEmit`, `eslint` and `prettier --check` pass
- Exercised in the browser against a local dev server, including the delay behaviour

Reviewers may want to look at the delay specifically, since its whole value is timing. Throttling to
Slow 3G and navigating between two list pages shows the skeleton fading in rather than appearing
hard; navigating to an entity with no rows should show the progress bar and nothing else.

## Notes

Three papercuts found along the way and fixed here rather than left behind:

- `postinstall` now runs `patch-package --error-on-fail` (see above).
- The `experimental.optimizePackageImports` entry for the MUI packages was a no-op, since both are
  already in Next's default list and Next unions the two. Removed, with a note on why there is
  deliberately no entry.
- `scripts/generate-templates.js` wrote its output unformatted, so every `prebuild` de-formatted the
  checked-in `test-templates.generated.ts` with a ~250 line content-identical diff. It now runs the
  output through Prettier and is verified idempotent.

### DIFF
--- a/apps/frontend/next.config.mjs
+++ b/apps/frontend/next.config.mjs
@@ -56,20 +56,17 @@ const nextConfig = {
   // Source maps: enable in prod for debugging, faster option in dev
   productionBrowserSourceMaps: isProd,
 
-  // Next.js already modularizes the app's own barrel imports for the heavy
-  // libraries (@mui/material, @mui/icons-material, lucide-react, date-fns,
-  // recharts, …) via its default optimizePackageImports list. But that default
-  // does NOT reach the barrel imports inside our transpilePackages workspace
-  // package @rhesis/ee-frontend, so its `import { X } from '@mui/material'`
-  // still pulls in the full CJS barrel. Under Turbopack that barrel eagerly
-  // evaluates @mui/material/useMediaQuery, whose CJS→ESM interop of
-  // `unstable_createUseMediaQuery` breaks in MUI v7.3.x — crashing the whole
-  // app at module-eval time (see mui/material-ui#46688). Listing the packages
-  // explicitly forces the barrel-to-deep-import rewrite for the transpiled EE
-  // sources too, sidestepping the eager useMediaQuery evaluation.
-  experimental: {
-    optimizePackageImports: ['@mui/material', '@mui/icons-material'],
-  },
+  // No experimental.optimizePackageImports entry here on purpose: @mui/material
+  // and @mui/icons-material are already in Next's default list, and Next unions
+  // the user's list with that default (`new Set([...userProvided, ...defaults])`
+  // in next/dist/server/config.js), so naming them changes nothing. The list is
+  // keyed on package name, so it never "reaches into" transpilePackages either.
+  //
+  // The crash this used to claim to fix (mui/material-ui#46688,
+  // `unstable_createUseMediaQuery is not a function`) is not an interop problem:
+  // @mui/material/useMediaQuery calls a 'use client' export at module-evaluation
+  // time, which fails whenever the module lands in the RSC server graph. That is
+  // fixed by patches/@mui+material+7.3.5.patch, which defers the call.
 
   // Environment variables available to the client
   // NEXT_PUBLIC_ prefix guarantees availability in client components
--- a/apps/frontend/package-lock.json
+++ b/apps/frontend/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "rhesis-app",
-  "version": "0.13.0",
+  "version": "0.14.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "rhesis-app",
-      "version": "0.13.0",
+      "version": "0.14.0",
       "hasInstallScript": true,
       "dependencies": {
         "@emotion/react": "^11",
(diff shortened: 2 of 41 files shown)


harry-rhesis:
Checked both points against the installed package. Neither reproduces.

**On the critical (`ERR_REQUIRE_ESM`):** Prettier 3 is not ESM-only. It ships a CJS entry and declares it in `exports`. From `node_modules/prettier/package.json` at the locked 3.7.3:

```json
"main": "./index.cjs",
"type": "commonjs",
"exports": { ".": { "require": "./index.cjs", "default": "./index.mjs" } }
```

So `require('prettier')` resolves to `index.cjs`:

```
$ node -e "const p=require('prettier'); console.log(typeof p.format, typeof p.resolveConfig)"
function function
```

`npm run generate-templates` and `next build --webpack` (which runs `prebuild`) both pass on this branch, which would not be possible if the require threw. The v3 change that does matter is that `format()` and `resolveConfig()` return promises, which is why `main()` is async and awaits both.

**On the `resolveConfig()` null spread:** object spread ignores `null` and `undefined`, so this is a no-op rather than a throw:

```
$ node -e "console.log(JSON.stringify({...null, a: 1}))"
{"a":1}
```

A missing config would simply mean Prettier defaults get used. In practice the repo has a config, and the generator's output is byte-for-byte identical to the checked-in file, so the resolved config is being applied.

Leaving both as-is. Happy to add `?? {}` for explicitness if a reviewer prefers it, but it does not change behaviour.

PR #2682, opened 2026-09-01
Match loading skeletons to each page's real shape

harry-rhesis:
## Problem

#2679 added route loading skeletons, but only one shape: a DataGrid. Most protected pages are
tables, so the catch-all under `(protected)` fit them. It does not fit the pages that are not
tables, and those got a fake table that dissolved into something else entirely.

Three shapes were missing.

## Card directories

metrics, requirements, models, tools and projects render a grid of the shared `EntityCard` behind
the borderless `directoryToolbarSx` toolbar, not a DataGrid.

All five declare an identical grid (`xs: 1fr`, `sm: 1fr 1fr`, `md: repeat(3, 1fr)` at
`theme.spacing(3)`), so one `PageCardGridSkeleton` covers them. Geometry mirrors `EntityCard`
itself: the 30px inset, the 20px section gap, the description clamped to 3 lines at 22px, the avatar
meta row and the chip section. Six cards by default, two rows at desktop width, overridable per
route.

## Insights dashboard

Insights is not a list at all: a compact filter bar, a pass-rate summary bar, a three-column
requirement breakdown, then a stack of per-requirement rows. `PageDashboardSkeleton` mirrors
`InsightsSummaryBar` and `RequirementInsightsView`.

This is the loosest fit of the set. The per-requirement rows are approximated as plain bars rather
than reproducing `RequirementInsightsRow`'s internals. If that reads as too vague in review, the
commit is self-contained and the route can fall back to a spinner instead.

## Section pages

settings, organizations/settings and organizations/usage stack `SectionCard`s. They are close enough
to a detail page that a fifth component was not warranted: `PageDetailSkeleton` gains a `showTabs`
prop and is reused. settings has no tab bar, so it also drops the `DetailTabPanel` top inset the bar
would sit above. All three pass `actionCount={0}` and `breadcrumbCount={0}`, matching headers that
carry neither.

## Chat page

playground is a chat interface, so its body has no predictable shape and a spinner is the honest
placeholder. Unlike architect it renders a `PageLayout` header with two FABs, so the header is drawn
and only the chat area falls back to the spinner.

## Shared building blocks

Adding two shapes meant four skeletons repeating the same pieces, so the last two commits pull them
out. Each is now declared once:

| Piece | Was | Now |
| --- | --- | --- |
| `PageLayout` header | 4 copies | `SkeletonPageHeader` |
| Filter / search / tabs row | 3 near-identical rows, 3 copies of the same control sizes | `SkeletonToolbar` |
| Card surface (radius, elevation, border) | 3 copies | `skeletonSurfaceSx` |
| Border helper | 4 copies under 3 names | `skeletonBorder` |
| Full-width status wrapper | 4 copies | `DelayedReveal` |

The three toolbars did differ in two real ways, now props rather than duplication: the gap (spacing
units inside the bordered grid card, a literal 20px for `directoryToolbarSx`) and the right slot
(`actions` for the grid's Select/Columns/Density/Export, `chips` for the insights filters, `none` for
card directories).

Worth being straight about the size: this does **not** shrink the code. 1131 lines before, 1134
after, because the two new modules offset what came out of the skeletons. The value is that the
control dimensions (36, 240, 38, 248) and the surface can no longer drift between shapes, not fewer
lines.

## Pages that needed nothing

Worth recording so the next person does not re-check them:

- `organizations`, `organizations/new`, `organizations/[identifier]` and `organizations/team` are
  pure `redirect()` calls. A server redirect resolves well inside the 500ms hold, so no skeleton is
  ever painted.
- `dashboard` and `test-results` are also redirects, to insights.
- `generation` is a client-side redirect that already renders its own spinner.
- `explorer` renders `ExplorerGrid`, a real DataGrid, so the table catch-all is already correct, and
  its default FAB count already matches.

## Testing

- `next build --webpack` passes
- `tsc --noEmit`, `eslint`, `prettier --check` and the hardcoded-styles check pass across all 15
  touched files
- 269 existing tests pass across the metrics, requirements, models, tools, insights and projects
  suites

Each route passes its real header FAB count (metrics 1, requirements 1, models 2, tools 1,
projects 1, insights 2, playground 2, the three section pages 0) rather than the generic default of
2, so the header does not shift when content arrives.

Reviewers may want to look at `/metrics` and `/insights` directly. The 500ms hold means throttling
is needed to see a skeleton at all.

### DIFF
--- a/apps/frontend/src/app/(protected)/insights/loading.tsx
+++ b/apps/frontend/src/app/(protected)/insights/loading.tsx
@@ -1,0 +1,7 @@
+import PageDashboardSkeleton from '@/components/loading/PageDashboardSkeleton';
+
+// Insights is a dashboard: filter bar, pass-rate summary, requirement
+// breakdown and per-requirement rows. Neither list shape fits it.
+export default function InsightsLoading() {
+  return <PageDashboardSkeleton actionCount={2} />;
+}
--- a/apps/frontend/src/app/(protected)/metrics/loading.tsx
+++ b/apps/frontend/src/app/(protected)/metrics/loading.tsx
@@ -1,0 +1,7 @@
+import PageCardGridSkeleton from '@/components/loading/PageCardGridSkeleton';
+
+// metrics renders a grid of EntityCards, not a DataGrid, so the catch-all
+// table skeleton would be the wrong shape here.
+export default function MetricsLoading() {
+  return <PageCardGridSkeleton actionCount={1} />;
+}
--- a/apps/frontend/src/app/(protected)/models/loading.tsx
+++ b/apps/frontend/src/app/(protected)/models/loading.tsx
@@ -1,0 +1,7 @@
+import PageCardGridSkeleton from '@/components/loading/PageCardGridSkeleton';
+
+// models renders a grid of EntityCards, not a DataGrid, so the catch-all
+// table skeleton would be the wrong shape here.
+export default function ModelsLoading() {
+  return <PageCardGridSkeleton actionCount={2} />;
+}
--- a/apps/frontend/src/app/(protected)/organizations/settings/loading.tsx
+++ b/apps/frontend/src/app/(protected)/organizations/settings/loading.tsx
@@ -1,6 +1,7 @@
 import PageDetailSkeleton from '@/components/loading/PageDetailSkeleton';
 
-// Organization settings is a tabbed section page with no header FABs.
+// Organization settings is a tabbed section page with a 2-item breadcrumb
+// trail and no header FABs.
 export default function OrganizationSettingsLoading() {
-  return <PageDetailSkeleton actionCount={0} breadcrumbCount={0} />;
+  return <PageDetailSkeleton actionCount={0} breadcrumbCount={2} />;
 }
--- a/apps/frontend/src/app/(protected)/organizations/settings/loading.tsx
+++ b/apps/frontend/src/app/(protected)/organizations/settings/loading.tsx
@@ -1,0 +1,6 @@
+import PageDetailSkeleton from '@/components/loading/PageDetailSkeleton';
+
+// Organization settings is a tabbed section page with no header FABs.
+export default function OrganizationSettingsLoading() {
+  return <PageDetailSkeleton actionCount={0} breadcrumbCount={0} />;
+}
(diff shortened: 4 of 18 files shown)


harry-rhesis:
Good catch, all three confirmed and fixed in 2734e50f3.

The real headers do render trails:

```
settings               const breadcrumbs = [{ label: 'Settings', href: '/settings' }]
organizations/settings [{ label: organizationName, ... }, { label: 'Organization Settings', ... }]
organizations/usage    [{ label: organizationName, ... }, { label: 'Usage', ... }]
```

So the skeleton was omitting the breadcrumb row and the header dropped when content arrived, which is exactly the shift these skeletons are meant to avoid. Now `breadcrumbCount={1}`, `{2}` and `{2}` respectively. Passing a non-zero count also restores the 20px stack gap that `PageLayout` only applies when a trail is present, so the spacing matches too.

I had read `breadcrumbs` in each `pageHeader` object and assumed it was empty without opening it.

I also audited every other route with a `loading.tsx` for the same mismatch, and there are none. Every list, card and dashboard page passes `breadcrumbs={[]}`, so 0 is correct for them, and every `[identifier]` route passes a 2-item trail, which matches the `PageDetailSkeleton` default.

harry-rhesis:
Both confirmed and fixed in 468f967d8. The second one turned out to be worse than reported.

**Tools tabs** — correct as stated. `ToolsPageClient` passes no `middleContent` to `GridToolbar`, so it has no centred pill tabs and the skeleton was drawing tabs that vanish. Added a `showTabs` prop on `PageCardGridSkeleton` and tools passes `false`. Checked the other four directories: metrics, requirements, models and projects all pass `middleContent`, so only tools needed it.

**Insights alignment** — right symptom, but the fix would have aligned shapes that shouldn't exist. Following it into `TestResultsFilters`:

- it never passes `rightContent`, so there is no right-hand slot at all
- its `middleContent` is conditional, `undefined` unless filters are active or the expand toggle applies

So the three filter chips in the dashboard skeleton were UI I had invented. Adding a spacer would have neatly right-aligned a row of placeholders with no counterpart in the real bar. They're removed, along with `SkeletonToolbar`'s `chips` variant and its `chipWidths` prop. The insights toolbar skeleton is now the filter button plus search pill, which is what the bar actually renders on load.

I did take the underlying point about the spacer and generalised it: `GridToolbar` inserts one whenever `middleContent` is absent, so it is now unconditional rather than gated on `rightSlot="actions"`.

`tsc`, `eslint`, `prettier` and the hardcoded-styles check all pass.

PR #2685, opened 2026-09-01
Enforce tracing-span limits through retention and an ingest backstop

harry-rhesis:
## What this does

Makes `tracing_spans` an enforced resource. The published limit was already metered and notified on, but nothing acted on it, because a 402 on span ingest is silently swallowed by the exporter's `BatchSpanProcessor`. So enforcement takes two other shapes:

- **Tier-based retention.** A nightly sweep hard-deletes trace rows past the org's `retention_days`, resolved from the tier catalog plus any per-org override on the licence token.
- **A backstop on ingest.** Rejection at 10x the tier limit, for runaway or abusive exporters only. It fails open: a billing-side problem must never break ingestion.

Both are off by default. `TRACE_RETENTION_ENABLED` gates the sweep, and it starts in dry-run so the blast radius shows up in logs before anything is deleted.

## Notable details

**The sweep must not cross org boundaries.** It runs under `bypass_tenant_filter()`, which is required rather than incidental: the session's scope has no project, so the ORM auto-filter would otherwise add `project_id IS NULL` and match no traces at all. That makes the explicit `organization_id` predicate load-bearing, and there's a regression test that sweeping one org leaves another's traces alone.

**Deletes are batched** at 5000 rows per transaction. One unbounded `DELETE` would take an org's whole backlog at once on first live run, holding row locks and writing a single large WAL burst.

**The backstop's fail-open is real.** It takes `tenant_context` and looks the org up inside its own `try`, rather than depending on `get_current_organization` — FastAPI resolves sub-dependencies before the function body, so that dependency's own `db.get` could 500 and a missing org row 403s, both from outside the `try`.

**Retention overrides are minted, not just read.** `custom_retention_days` flows JWT → `Entitlements` → `provider.info()` → `ConfigQuotaProvider`, and `--retention-days` on the licence CLI writes it. `retention_days` from the tier YAML is now validated like its siblings; a string `"90"` from a mounted ConfigMap would otherwise reach `timedelta(days="90")` and be swallowed by the sweep's per-org `except`.

## Also in here

Two fixes found while testing this, both user-visible:

- The free tier rendered as **"Community"** in the sidebar, org menu and usage page, while the pricing page advertises those exact limits as the **Free** plan. `plan.name` is rendered verbatim by clients, so the mapping belongs in `build_plan`.
- `@mui/material` was declared `"^7"` while `postinstall` runs `patch-package --error-on-fail` against a patch named for 7.3.5. That patch is load-bearing (it stops `useMediaQuery` crashing at module-evaluation in the RSC graph), so the first `npm install` after any 7.x release could hard-fail the install. Pinned, lockfile regenerated so `npm ci` stays in sync.

## Testing

337 backend tests pass across the licensing, quota, gates and retention suites. New coverage:

- backstop boundary: unlimited never backstopped, off-by-one at 10x, over-limit-but-under-backstop allowed, verdict carries the backstop ceiling not the tier limit
- retention escalation loopholes: unparseable overrides fall back to the tier, cross-org isolation, revoked / expired / unlicensed orgs
- sweep: org isolation, per-org error isolation, dry-run counts without deleting, unlimited retention skips, global override, batching across multiple passes, and a failed sweep reporting `orgs_failed` rather than looking clean
- mint/verify round trip for `custom_retention_days`, and refusal of non-positive values at mint time
- plan naming, including that "(inactive)" is gated on `is_paid` so a free org is never labelled inactive

## Reviewer notes

- `alembic heads` returns a single head. Worth confirming on merge, since a second head aborts every upgrade.
- The free-tier naming fix lives in `services/plan.py`, not `services/usage.py`: `build_plan` moved there in #2683 while this branch was open, so the change was relocated during the rebase and three existing assertions in `tests/backend/services/test_plan.py` were updated from "Community" to "Free".
- The sweep has no `celery beat` process deployed against it yet, same as the existing job retention sweep. The schedule entry is in place for when there is one.

### DIFF
--- a/apps/backend/src/rhesis/backend/alembic/versions/b258958270d0_add_trace_retention_index.py
+++ b/apps/backend/src/rhesis/backend/alembic/versions/b258958270d0_add_trace_retention_index.py
@@ -1,0 +1,31 @@
+"""Add trace retention index (organization_id, created_at).
+
+The trace retention sweep (jobs/trace_retention.py) hard-deletes rows
+per org filtered by created_at. Without this composite index the sweep
+would need a sequential scan per org on every run.
+
+Revision ID: b258958270d0
+Revises: ff71b040aebf
+Create Date: 2026-09-01
+"""
+
+from typing import Union
+
+from alembic import op
+
+revision: str = "b258958270d0"
+down_revision: Union[str, None] = "ff71b040aebf"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.create_index(
+        "idx_trace_org_created",
+        "trace",
+        ["organization_id", "created_at"],
+    )
+
+
+def downgrade() -> None:
+    op.drop_index("idx_trace_org_created", table_name="trace")
--- a/apps/backend/src/rhesis/backend/alembic/versions/b258958270d0_add_trace_retention_index.py
+++ b/apps/backend/src/rhesis/backend/alembic/versions/b258958270d0_add_trace_retention_index.py
@@ -5,7 +5,7 @@ per org filtered by created_at. Without this composite index the sweep
 would need a sequential scan per org on every run.
 
 Revision ID: b258958270d0
-Revises: ff71b040aebf
+Revises: 42f4cf7014a0
 Create Date: 2026-09-01
 """
 
@@ -14,7 +14,7 @@ from typing import Union
 from alembic import op
 
 revision: str = "b258958270d0"
-down_revision: Union[str, None] = "ff71b040aebf"
+down_revision: Union[str, None] = "42f4cf7014a0"
 branch_labels = None
 depends_on = None
 
--- a/apps/backend/src/rhesis/backend/app/auth/quota_gates.py
+++ b/apps/backend/src/rhesis/backend/app/auth/quota_gates.py
@@ -35,7 +35,11 @@ import logging
 from fastapi import Depends, Response
 from sqlalchemy.orm import Session
 
-from rhesis.backend.app.dependencies import get_current_organization, get_tenant_db_session
+from rhesis.backend.app.dependencies import (
+    get_current_organization,
+    get_tenant_context,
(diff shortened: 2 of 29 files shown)


harry-rhesis:
Thanks — went through all four. Two applied, two I'd push back on, with reasoning.

**Applied: backstop doesn't fail open when the org row is missing.** Confirmed reachable and confirmed the consequence: `db.get` returning `None` fell through to `get_policy(None)`, which resolves the community tier, so a possibly-paid org would be backstopped at 10x the *free* limit and could 402 on ingest purely because a lookup failed. That contradicts the fail-open contract in the function's own docstring. Fixed in d1c78fc, with three tests covering the missing row, no org in context, and a lookup that raises.

**Applied: the `USAGE_QUOTAS_ENABLED=false` interaction.** Your reading is right and it is intended. Both halves were already documented but the module never connected them, so an operator flipping `TRACE_RETENTION_ENABLED` on a self-hosted instance would have to derive why nothing gets deleted. The docstring now says it directly.

**Not applied: `== str(org.id)` forcing a cast.** The premise doesn't hold for this codebase. `organization_id` is `GUID()`, and `GUID.process_bind_param` does `return str(value)` on postgres for *any* input, so `== org.id` and `== str(org.id)` produce an identical bind parameter — there's no cast, no index-usage difference, and no cross-dialect divergence to avoid. `quota/enforcement.py:168` already compares this column to a `str` on the same basis.

**Not applied: `CREATE INDEX CONCURRENTLY`.** The locking concern is real in principle, but none of the 52 index-creating migrations in this repo use it, so adopting it here alone would be a one-off divergence rather than a pattern. It also needs an autocommit block, and a failed concurrent build leaves an INVALID index requiring manual cleanup. Given the current table size I'd rather keep it consistent and revisit as a deliberate repo-wide change if and when `trace` gets big enough to need it — happy to be overruled if you'd rather take the divergence now.

Issue #2042, opened 2026-06-25
Add Trello MCP tool integration


Ayush7614:
## Summary

Add Trello as a supported MCP tool provider so users can import boards, cards, and project context from Trello into Rhesis for test generation and Architect grounding.

## Problem / Opportunity

Trello remains a common lightweight project-management tool, especially for smaller teams and Kanban workflows. Users tracking requirements on Trello boards have no way to connect them to Rhesis today.

## Proposal

Follow the established provider pattern (GitLab, Asana, Shortcut, Linear, Azure DevOps):

### Backend

- Add `trello` `ToolProviderType` to a new Alembic migration (chain from current head; add merge migration if needed)
- Route `trello` extract + test-connection through MCP in `services/tool/actions.py`
- Add provider-specific credential/metadata validation in `routers/tools.py`
  - `credentials`: `TRELLO_API_KEY`, `TRELLO_TOKEN`
  - `tool_metadata`: optional scope such as `{ "board_id": "..." }` (same pattern as Asana `workspace_gid` / Linear `team_id`)

### SDK

- Add `sdk/src/rhesis/sdk/agents/mcp/provider_templates/trello.json.j2` pointing to a Trello MCP server (e.g. `npx -y atlassian-trello-mcp` or `@delorenj/mcp-server-trello`)

### Frontend

- Add `trello` to `EXTRACT_PROVIDERS` in `apps/frontend/src/config/tool-providers.tsx`
- Add Trello icon, description, and credential fields (API key + token) in `ToolConnectionDrawer.tsx`

## Acceptance Criteria

- [ ] Trello `ToolProviderType` seeded in DB
- [ ] Provider template renders a valid MCP config for the chosen Trello MCP server
- [ ] Users can configure a connection (API key, token, optional board scope) via the Tools UI
- [ ] Search, extract, and test-connection operations work against Trello boards/cards
- [ ] Imported content surfaces correctly as a `Source` in Knowledge
- [ ] Backend + SDK tests cover credential validation and template rendering

Ayush7614:
cc: @EmanueleDeRossi1 @harry-rhesis

sujalgawas:
can you please assign this issue to me?

harry-rhesis:
@sujalgawas kindly implemented this one.

PR #2689, opened 2026-09-02
feat: Trello mcp integration

sujalgawas:
### Summary
issue #2042 

Integrated trello by following the proposal in the issue

### Criteria
Completed all the mentioned criteria in the issue.

- [x] Trello ToolProviderType seeded in DB
- [x] Provider template renders a valid MCP config for the chosen Trello MCP server
- [x] Users can configure a connection (API key, token, optional board scope) via the Tools UI
- [x] Search, extract, and test-connection operations work against Trello boards/cards
- [x] Imported content surfaces correctly as a Source in Knowledge
- [x] Backend + SDK tests cover credential validation and template rendering

### tests done
1. created manual test files test_tools_trello.py and test_trello_provider_template.py
2. ran all the tests mentioned in issue
3. manual testing


harry-rhesis:
Hi @sujalgawas, thanks for coming over and contributing to our repo, much appreciated! There a few failing checks. Would you mind fixing those before we can review? Can you also please rebase your branch? Cheers, Harry

sujalgawas:
@nicolai-rhesis can you check the pr? Please

harry-rhesis:
@sujalgawas thanks for the continued iterations on this PR, appreciate it. The checks are still failing. Could you double-check again?

sujalgawas:
@harry-rhesis I think it should be fixed now I missed Migration lint check. sorry for so much of back and forth. I am new to open sourcing

harry-rhesis:
Hey @sujalgawas, thanks for putting this together! Really appreciate you contributing to the project.

All the CI failures here come from one thing: the branch needs to be rebased onto the latest `main`. Nothing is wrong with your code itself. The Python lint, SDK tests, backend tests, frontend build, and security scans are all passing.

Here's what's happening: since you branched off, new migrations landed on `main`. Your migration (`b6723e65f233`) points to an older parent revision (`b258958270d0`), but `main` has moved on to `b7e1c9d4a2f3`. This creates two "heads" in the migration chain, and Alembic refuses to run when it sees that. That's why the backend container crashes on startup, which then takes down the Java SDK tests and all three E2E shards with it.

Here's how to fix it, step by step:

**1. Fetch the latest main and rebase your branch onto it:**

```bash
git fetch origin
git rebase origin/main
```

If you get merge conflicts, resolve them file by file, then `git add <file>` and `git rebase --continue` for each one.

**2. Update your migration's parent revision:**

Open `apps/backend/src/rhesis/backend/alembic/versions/b6723e65f233_add_trello_tool_provider_type.py` and change:

```python
down_revision = "b258958270d0"
```

to:

```python
down_revision = "b7e1c9d4a2f3"
```

This tells Alembic that your migration comes after the current head on `main`, so there's a single chain again.

**3. Force-push your rebased branch:**

```bash
git push --force-with-lease
```

That should clear all five failing checks. If you run into any trouble with the rebase, feel free to ask here and we'll help you through it.

Thanks again for the contribution!

harry-rhesis:
Thanks @sujalgawas for pushing the migration fix, that cleared the CI failures.

I went through the whole change in detail and traced the Trello path end to end against how the other MCP providers work. The structure you followed is right: the provider template, the transport routing in `actions.py`, the validators, and the migration all match the existing Asana and Linear patterns closely. The complexity refactor into `_validate_tool_metadata_on_update` and `_validate_and_merge_credentials_on_update` was also the correct response to tripping the ruff ceiling. Splitting the function rather than silencing the rule is exactly what we ask for.

There are three things that stop the integration from working today. I do not think it has been run end to end yet, and CI cannot catch any of them, so let me walk through each one.

## 1. The MCP server package does not exist

`sdk/src/rhesis/sdk/agents/mcp/provider_templates/trello.json.j2` points at `@delano/mcp-trello`. That package is not on the npm registry, it returns a 404.

It is worth knowing how this actually runs. In the backend image, `npx` is a shim that strips `-y` and calls `bunx --silent <package>` (see `apps/backend/Dockerfile:198`). So this becomes `bunx --silent @delano/mcp-trello`, the process exits immediately, and the MCP handshake never completes. Every Trello connection test and every import fails.

The package you most likely want is `@delorenj/mcp-server-trello`. It is actively maintained and reads exactly the two variables you are already passing, `TRELLO_API_KEY` and `TRELLO_TOKEN`. So your credential design is correct, it is just the package name.

The SDK test did not catch this because it asserts the rendered template equals the same package name, so it locks the value in rather than checking that it resolves.

## 2. Saving a Trello tool sends only the token, and always fails with a 400

There are two separate create paths in `ToolConnectionDrawer.tsx`. You patched the one inside `handleTestConnection`, which is why Test Connection works. The other is inside `handleSubmit`, around line 1281. It handles jira, confluence, gitlab and azure_devops, then falls through to:

```ts
credentials = {
  [getCredentialKey(provider.type_value)]: authToken.trim(),
};
```

`getCredentialKey('trello')` returns `TRELLO_TOKEN`, so the save request carries only the token and no API key. The backend rejects it with 400 "TRELLO integrations require 'TRELLO_API_KEY'". A user can test the connection successfully and then find they cannot save it.

That same block sets Asana workspace metadata but has no Trello branch, so the Workspace ID is dropped on create as well.

## 3. The Workspace ID field never reaches the prompt

This one is invisible from the code you changed, so it is an easy thing to miss. The three Jinja blocks you added are all guarded on `workspace_context`, and `workspace_context` is populated in `apps/backend/src/rhesis/backend/app/services/tool/mcp/operations.py:24`:

```python
"workspace_context": scope_context if provider == "asana" else None,
```

For Trello that is always `None`, so none of the three blocks ever render. The value the user types is validated, stored in `tool_metadata`, converted into a scope context by `mcp/config.py`, and then discarded one line before the prompt is built.

Rather than wiring it up, please remove the workspace field entirely from this PR. Two reasons. First, we do not want to add new options and configuration to the platform just because a new integration could support them, and Linear and Shortcut both ship with no scope metadata at all, so that is the shape a first Trello integration should match. Second, even if it were wired, it would only be a sentence of instruction inside a prompt, not real enforcement. If we want Trello workspace scoping later, `@delorenj/mcp-server-trello` enforces it properly through its own `TRELLO_ALLOWED_WORKSPACES` variable, which is a different and better implementation.

Removing it deletes a good chunk of code: the frontend TextField and its state, `_validate_trello_workspace_gid` and its four call sites, the branch in `mcp/config.py`, the three template blocks, the asana-or-trello conditions, and two tests. It also removes the dropped metadata bug in point 2.

One smaller note in the same spirit. The frontend asks the user to re-enter both credentials when changing either one, but `merge_trello_credentials_on_update` in this same PR exists specifically to avoid that, and it runs before validation on both the PATCH and test connection paths. So the merge can never be reached from the UI. Azure DevOps already handles this the simpler way: merge on the backend, let the user change one field. I would follow that and drop the two "please re-enter both" guards.

## Documentation

Please also add Trello to the docs. Three places:

- `docs/content/docs/tools/index.mdx`, add a `<details>` block following the existing ones. It needs to explain how to get both values: create or open a Power-Up at [trello.com/power-ups/admin](https://trello.com/power-ups/admin), generate an API key on its API key tab, then use the Token link next to the key to authorize and generate the token. Worth noting the token is shown only once.
- `docs/src/public/integrations/providers/trello.svg`, add the icon.
- `docs/src/components/FrameworkIntegrationsShowcase.jsx`, add the Trello entry pointing at `/docs/tools#trello`.

Have a look at `docs/AGENTS.md` for the writing rules before you start.

## Please test it end to end before pushing

None of the three issues above can be caught by CI as it stands, so running it yourself is the important step:

```bash
./rh dev status    # see this checkout's ports
./rh dev up
./rh dev backend   # in one terminal
./rh dev frontend  # in another
```

Then in the UI, connect a Trello tool with a real API key and token, click Test Connection, and then actually click Save. After that go to Knowledge and import a real Trello card or board URL. All three steps need to pass. Test Connection on its own is not enough, that is the one path that already works today.

It would also be good to add a frontend test covering the save path, since that gap is what let point 2 through.

Two small things while you are in there: `initial_data.json` lost its trailing newline, and the new frontend test carries a 15000ms timeout, the only custom timeout in that file.

Thanks again for taking this on, and sorry for the long list. The overall shape of the integration is right, which is the harder part to get correct. Happy to help if anything above is unclear.

sujalgawas:
> SDK Trello provider template test is asserting the wrong MCP package name.
> 
> SDK test mismatch (blocking)
> 
> > Fix: in `tests/sdk/agents/mcp/test_trello_provider_template.py`, update the assertion to expect `@delorenj/mcp-server-trello` (to match `sdk/.../trello.json.j2`).
> 
> Found 1 issue (1 critical).

I have changed it. I don't know why it's still flaging it

harry-rhesis:
Heads up: #2717 just merged and added two migrations, so `b7e1c9d4a2f3` is no longer the head of the chain. Your `b6723e65f233` was still pointing at it, which would have left `main` with two heads and broken `alembic upgrade head` for everyone.

I've pushed a one-line commit here re-pointing it at the new head, `b8d2f3e4a5c6`. Nothing else in the PR is touched. The `migrations` check should go green on the next run.

harry-rhesis:
Re-pointed again: `main`'s Alembic head moved from `b8d2f3e4a5c6` to `a0a48c28ad64` when #2728 merged release v0.15.2 back, so this migration was pointing at a parent that is no longer the head.

Worth flagging that the `migrations` check on this PR was showing green while that was true. It last ran before #2728 landed, so it was passing against the old base. Merging on that green would have left `main` with two heads.

`b6723e65f233` now chains onto `a0a48c28ad64`. Verified against the merge commit the way CI does it: single head, `OK: migration chain is valid`.

harry-rhesis:
@sujalgawas nothing you did broke this, and I want to explain the mechanism because it is genuinely confusing from where you are sitting.

## Why the migrations check keeps going red

Your migration declares a parent through `down_revision`. The `migrations` check compares that parent against whatever is currently the head of the chain on `main`. Every time any other PR containing a migration merges, the head moves, your `down_revision` now points at an ancestor instead of the head, and the check turns red. You changed nothing. It went stale underneath you.

That has now happened three times on this branch:

- `b7e1c9d4a2f3`, invalidated when #2717 merged
- `b8d2f3e4a5c6`, invalidated when #2728 merged the v0.15.2 release back
- `a0a48c28ad64`, invalidated when #2744 merged

I re-pointed it for you the first two times. Main's head is now `e1b7c3d95a04`, so it is stale again.

There is a second and nastier version of the same thing, which I flagged earlier on this PR: the check can sit there **green** while being stale, because it last ran before the newest migration landed. So a green migrations check is only trustworthy if it ran after the most recent merge to main. That is the same effect behind your "I changed it, I don't know why it's still flagging" on the SDK test. CI was reporting on an older commit.

None of this is a judgment on your code. It is what happens when a PR carrying a migration stays open while main keeps moving.

## Where the review actually stands

You have done most of it, and that deserves saying plainly:

- The MCP package name is fixed and now points at `@delorenj/mcp-server-trello`.
- The save path is fixed. `handleSubmit` has its Trello branch, so the API key is actually sent and the 400 is gone.
- The docs are in, all three files, and the token instructions read well.
- The test timeout is gone.

One substantive item is still open, the Workspace ID field. `apps/backend/src/rhesis/backend/app/services/tool/mcp/operations.py:24` still reads:

```python
"workspace_context": scope_context if provider == "asana" else None,
```

so for Trello it is still always `None`, the three Jinja blocks still never render, and the field is still inert. As before, the ask is to delete it rather than wire it up, so the integration matches the Linear and Shortcut shape.

## Proposed way forward

Sequencing matters more than effort here:

1. **Stop chasing the migrations check.** It is not telling you anything about your code, and it will go stale again the next time someone merges a migration. Leave it red while you finish up.
2. **Remove the workspace field**, and while you are in there drop the two "please re-enter both" guards in favor of the backend merge, as in my earlier note.
3. **Test it end to end locally**, specifically Save and then a real import from a Trello URL, since that is the part CI cannot check for you.
4. **Push, then ping me here.** I will rebase onto main and re-point `down_revision` to whatever the head is at that moment, and we merge straight away. Doing the re-point last, immediately before merging, shrinks the stale window from days to minutes.

The branch is also 24 commits behind main now, so the rebase in step 4 is worth doing regardless, to get the e2e tests running against current code.

Thanks for sticking with this. The integration itself is nearly there, and the churn you have been fighting is a property of the repo, not of your work.

sujalgawas:
@harry-rhesis I have tested the integration manually with importing a card, now it works. I have forced pushed this so that you can just rebase and update the migration and merge.  let me know if there is any issues.

harry-rhesis:
@sujalgawas thanks for testing the card import and confirming it works. That was the one thing CI could not check for us.

I took it from here as you suggested. Three changes, all mine, none touching your logic:

1. **Rebased onto current `main`.** Your 13 commits replayed cleanly and the changed-file list is identical, so nothing of yours was lost.
2. **Re-pointed the migration** onto `a3f6c81d05b2`, main's current head.
3. **Pinned the MCP server** to `@delorenj/mcp-server-trello@1.8.1`.

That third one is the only surprise. While this PR was open, main gained `test_provider_template_pins.py`, which requires every stdio template to pin an exact version. All the other providers were pinned in that change. Trello predated the rule, so it was the last unpinned one and `test-sdk` went red on it. I used 1.8.1, the current latest, not the `2.0.0-beta.0` on the `next` tag, and updated the test assertion to match.

19 checks passing, 4 running, none failing. The branch is up to date with `main` and I will merge once it goes green.

Two notes for your side:

- I force-pushed twice, so your local copy is behind. Run `git fetch origin && git log origin/trello_mcp_integration..trello_mcp_integration --oneline` first. If it prints nothing, `git reset --hard origin/trello_mcp_integration` is safe. If it prints commits, tell me and I will help replay them.
- Every force-push resets the fork approval gate, so the workflows need a maintainer to approve them before they run. If you ever see no checks at all rather than failing ones, that is why.

Thanks for working through the review items.

sujalgawas:
Its ready to merge, right? cause everything is green and i ran 

git fetch origin && git log origin/trello_mcp_integration..trello_mcp_integration --oneline

git reset --hard origin/trello_mcp_integration

without any issues

harry-rhesis:
Merged as `6267d1b34`. Thank you, @sujalgawas.

This went through a lot of rounds, and you stuck with all of them. Worth saying what you actually fixed along the way: a nonexistent npm package, a save path that silently dropped the API key, and a workspace field that turned out to be completely inert. None of those were catchable by CI, and the last one was buried in a file you had not even touched.

Two things stood out. You removed the workspace field rather than wiring it up, which was the harder call and the right one. And you did not just patch the save bug, you added tests for the create path and for partial credential updates, so neither can regress quietly.

You also got dealt a genuinely unfair hand with the Alembic churn. Your `down_revision` went stale four times through no fault of yours, and a version pinning rule landed on main partway through. That is the repo making life hard for a contributor, not you doing anything wrong, and it is something we should make less painful.

Trello is live now, documented alongside the other providers. Looking forward to your next contribution.

PR #2707, opened 2026-09-04
Surface trace tokens and cost, and fix the totals behind them

Arman-Beykmohammadi:
## Purpose

Rhesis already ingested LLM token counts and already priced them with LiteLLM, but almost none of it reached the user. Enrichment wrote a full per-span cost breakdown into `Trace.enriched_data["costs"]`, the backend already sent `total_tokens` and `total_cost_usd` on `TraceSummary` and `TraceDetailResponse` — and no frontend component read any of it. `formatCost()` and `formatTokenCount()` had existed in `trace-utils.ts` with no caller outside their own unit test. You opened a trace, saw `6 spans · 1.32s · development`, and had no idea what the run cost or how many tokens it burned.

Surfacing it first meant fixing what was underneath, because there were **four** different definitions of "this trace's token total" in the codebase, and three of them were wrong:

| Site | What it did |
|---|---|
| traces list | read `ai.llm.tokens.total` off the root span alone — usually an agent-run span with no token attribute, so it showed `0` |
| test-run trace list | the same root-span-only pattern, a near-verbatim copy |
| trace detail | summed **every** span, double-counting frameworks that report aggregated usage on the agent-run span *and* per-call usage on its children |
| project metrics | SQL `SUM` over all span rows in the project, with the same double-count |

Cost enrichment already avoided the double-count, because it prices only spans where `ai.operation.type == "llm.invoke"`, which excludes the aggregate-carrying agent-run span. That filtered set is the correct basis for tokens too, so making tokens and cost derive from one source is the core of this change.

Two further bugs surfaced while doing it, both fixed here:

- **Project-metrics cost was inflated by span count.** `mark_trace_processed` writes the identical trace-level `enriched_data` blob onto *every* span row of the trace, and the metrics query summed it across rows — so a six-span trace reported six times its real cost. That number goes on screen with the new rollup tiles, so it could not stay.
- **Unpriceable models lost their tokens entirely.** On a LiteLLM pricing failure the loop hit `continue` *before* appending the breakdown, so the span's token counts never reached `enriched_data`. A self-hosted or unrecognised model reported zero usage, and a trace mixing priced and unpriced spans silently undercounted.

## What Changed

**Backend — one source of truth**

- `services/telemetry/enrichment/core.py` rolls input/output/total tokens into `TokenCosts`, and records unpriceable spans at zero cost with their real tokens instead of dropping them. The per-span body moves into `_price_span` to stay under the `ruff.toml` complexity ceiling.
- New `services/telemetry/token_totals.py` is the single accessor all read paths now use. It prefers the enrichment breakdown and falls back to summing `ai.llm.tokens.total` over `llm.invoke` spans.
- The fallback matters because enrichment is dispatched asynchronously after ingest, so a trace opened immediately has `enriched_data = {}`. `query_traces` gets a matching correlated subquery so the list and the drawer agree in that window too, instead of the list showing `0` while the drawer shows the real figure.
- The two list endpoints now share `trace_summary_totals` rather than each carrying its own copy of the same arithmetic.
- `get_trace_metrics_aggregated` collapses to one row per trace before aggregating tokens and cost. `total_spans`, `error_count` and the duration percentiles stay span-level, which is why `base` is not deduplicated wholesale.
- `SpanNode` gains `cost_usd` and `model_name`, indexed out of the breakdown by `span_id`. The raw span attributes carry tokens but never a price, so the Span Details panel had no other way to get it.

**Frontend — four surfaces**

- Token and cost chips in the trace drawer header, hidden rather than showing `0` when a trace has no LLM spans.
- A Usage block in Span Details, shown only when a span actually has tokens or cost. `ai.llm.tokens.*` is dropped from the raw attributes table so it does not appear twice.
- Tokens and Cost columns in the traces list, falling back to an em dash.
- A four-tile project rollup, which finally gives `GET /telemetry/metrics` a consumer — it had been implemented and unused, its docstring still saying "for future dashboard use".

## Additional Context

**`total_tokens` is read as reported, never derived as `input + output`.** Google ADK deliberately folds cache-read tokens into `ai.llm.tokens.total` (`google_adk/mapping.py` uses `=` where every other integration uses `setdefault`), so for ADK spans the total legitimately exceeds `input + output`. Deriving it would silently discard cached tokens, and asserting the identity in a test would fail against real ADK data. The drawer tooltip therefore presents input and output as two independent figures rather than a breakdown of the total.

**Both new columns are `sortable: false`.** The grid paginates server-side but sorts client-side, and `list.ts` drops `sort_by`/`sort_order` before calling the API. A sortable Cost header would silently reorder the current page only — worse than no sorting, since sorting by cost is exactly what someone would try. Backend sort support for these fields is a separate change.

**The rollup tiles say when they are broader than the table.** `GET /telemetry/metrics` only narrows by project, environment and time, so an active endpoint, trace-source or search filter would make the tiles cover more traces than the rows below. They carry a caption saying so rather than silently disagreeing. Widening the endpoint to accept the remaining filters is its own change.

Deliberately out of scope:

- **Embedding tokens are no longer counted.** Embedder spans carry `ai.llm.tokens.*` but map to `ai.operation.type = embedding.create`, so the `llm.invoke` filter excludes them and the detail endpoint, which used to sum every span, no longer picks them up. This is deliberate: enrichment never priced embeddings either, so counting their tokens would put tokens and cost back out of step. Anyone who has looked at an embedding-heavy trace before will notice the number drop.

- **The SDK's own agent emits no token attributes.** `sdk/src/rhesis/sdk/agents/base.py` sets `ai.operation.type = llm.invoke` but never the token attributes, even though `packages/rhesis/src/rhesis/telemetry/attributes.py` has a builder that produces exactly the right set. Traces from Rhesis's own agent will show no tokens. It is an SDK change on the SDK's release cycle.
- **Normalizing non-Rhesis token conventions at ingest.** Spans arriving with only `gen_ai.usage.*`, OpenInference or OpenLLMetry keys are stored verbatim and read by nothing, because translation happens client-side per framework. A server-side normalizer in `create_trace_spans` would fix that for every producer at once, but it carries its own backfill question.
- Cost in EUR (the backend computes it; it stays out of the response schema), per-org currency preference, and cost budgets or alerting.
- The usage page and the `usage` table are untouched. `MODEL_TOKENS` there meters tokens *Rhesis* spends on an org's behalf — test generation, LLM judges. Traced tokens are the customer's own agent's tokens arriving over OTLP. Two separate ledgers, deliberately.

This PR is ~1650 lines (roughly 850 production, 810 tests), above the 400-line guidance in `.claude/skills/pull-request/SKILL.md`. Kept as one PR deliberately, since splitting the token unification from the surfaces would land a correctness fix nobody can see and then a UI change that has to be trusted. The four commits are scoped so each reads on its own.

## Testing

**Automated** — 204 backend tests and 847 frontend tests pass; `ruff`, `eslint`, `prettier` and `tsc` are clean.

New backend coverage:

- `tests/backend/services/telemetry/test_token_totals.py` — the pydantic-ai double-count case built explicitly from that span shape, the enriched and fallback paths agreeing for the same spans, and a trace with no LLM spans returning zeros rather than raising.
- `tests/backend/crud/test_trace_metrics_aggregated.py` — a multi-span trace reports its cost **once**, the regression test for the inflation, plus assertions that `total_spans`, `error_count` and the percentiles are unchanged by the restructuring.
- `tests/backend/routes/test_telemetry_query.py::TestTraceTokenAndCostVisibility` — the same properties over the real endpoints: no double count, list and detail agreeing, the ADK total surviving, `null` for a trace with no LLM spans, and per-span cost reaching `SpanNode`.
- `test_enrichment.py::test_calculate_costs_unknown_model` previously asserted that unpriceable spans were dropped. That contract is what this PR fixes, so the test now asserts the tokens are kept at zero cost, alongside a new mixed priced/unpriced case.

**Against the running stack** — verified end to end on a local backend, Postgres and Celery worker by posting pydantic-ai-shaped and ADK-shaped traces over HTTP:

- a pydantic-ai run reports **420 tokens, not 840**
- list and detail agree both before and after enrichment
- **the token total does not change when enrichment lands** for a trace that has stopped growing: the SQL fallback and the enriched path produce the same number. A multi-turn conversation still adding spans under one `trace_id` reports its pre-growth total for one enrichment cycle, and every surface reports the same stale value meanwhile rather than each showing a different one
- per-span costs sum exactly to the trace cost (`$0.0162`), and the agent-run span is correctly unpriced
- an ADK-shaped total of `950` survives instead of collapsing to `150`
- a trace with no LLM spans sends `null`, so the column renders an em dash

Also sanity-checked against a real 190-span, 14-turn `reg-advisor` conversation: 47 `llm.invoke` spans summing to exactly the reported 44,721 input + 1,014 output, and `$0.0127` matching LiteLLM's `gemini-3.1-flash-lite` rates to per-span rounding.

**To reproduce locally**: run an agent that has a Rhesis integration (`cd agents/reg-advisor && uv run python chat_terminal/chat.py`), then open Traces at `localhost:3000` scoped to that agent's project. Open the trace *immediately* to see tokens present while cost is still blank — that is the pre-enrichment fallback — then refresh and confirm cost appears while the token number stays put.

**Not verified**: the rendered chips in a browser. Every layer beneath the rendering has a real check, but nobody has yet looked at the UI itself.

---

## Review round two

Thanks @harry-rhesis and peqy. Everything raised is addressed, plus three UI changes Arman asked for.

**From the review**

- The rollup tiles no longer render on a test run's Traces tab, where they were counting the whole project across all time next to a dozen listed rows.
- Both SQL fallbacks now share one `span_total_tokens_expr`: `COALESCE(NULLIF(total, 0), COALESCE(input, 0) + COALESCE(output, 0), 0)`. Each side needs its own `COALESCE` because `NULL + 100` is `NULL` in SQL, which would otherwise drop a span reporting only one side. The `NULLIF` answers peqy's question about a zero total beside real usage, and the same rule now applies in both Python paths.
- `Trace.total_tokens` is deleted. Nothing read it, and it returned the per-span shape that caused the original bug.
- `TraceSummary.total_input_tokens` / `total_output_tokens` now feed a tooltip on the Tokens cell.
- Embedding exclusion and the staleness window are both documented above.

**New in this round**

- **The four tiles moved above the grid card** instead of sitting inside it.
- **The traces grid sorts on the server.** It paginated server-side but sorted client-side, so a sort only reordered the rows already on screen. Tokens sort on the same expression the response reports, cost sorts `NULLS LAST` so unpriced traces do not lead a "most expensive" list, and every sort ends with `start_time` then `id` so paging does not repeat rows. `conversation_input`, `endpoint_name` and `trace_metrics_status` are marked unsortable, since none of them has a SQL sort key and MUI applies `sortingMode` grid-wide.
- **Per-span usage rolls up the subtree.** Only `ai.llm.invoke` leaves carry token attributes, so clicking a pipeline, coordinator or agent span showed nothing at all. It now sums the span and everything under it, says how many LLM spans fed the number, and renders as text lines matching the Timing block.

**Verification:** 214 backend and 853 frontend tests pass; ruff, eslint, prettier and tsc are clean. Checked against a live backend: sorting by cost puts the most expensive trace first, sorting by tokens puts the largest first, and `sort_by=endpoint_name` returns 400. New coverage in `TestTraceListSorting`, `TestTokenTotalFallbackShapes` and the `subtreeUsage` unit tests.

The `trivy (backend)` failure was an install-step TLS error inside the action, not a finding. Both Trivy steps set `exit-code: 0`, so a vulnerability cannot fail that job. Re-run triggered.

### DIFF
--- a/apps/backend/src/rhesis/backend/app/constants.py
+++ b/apps/backend/src/rhesis/backend/app/constants.py
@@ -251,7 +251,14 @@ class AISpanAttributes:
 
     OPERATION_TYPE = "ai.operation.type"
     MODEL_NAME = "ai.model.name"
+    TOKENS_INPUT = "ai.llm.tokens.input"
+    TOKENS_OUTPUT = "ai.llm.tokens.output"
     TOKENS_TOTAL = "ai.llm.tokens.total"
+
+    # Only spans with this operation type carry countable token usage. Agent-run
+    # spans repeat their children's tokens as an aggregate, so counting them too
+    # double-counts the trace.
+    OPERATION_LLM_INVOKE = "llm.invoke"
 
 
 # Keys inside Trace.enriched_data (JSONB) populated by the enrichment service.
@@ -261,6 +268,12 @@ class EnrichedDataKeys:
     COSTS = "costs"
     TOTAL_COST_USD = "total_cost_usd"
     TOTAL_COST_EUR = "total_cost_eur"
+    TOTAL_TOKENS = "total_tokens"
+    TOTAL_INPUT_TOKENS = "total_input_tokens"
+    TOTAL_OUTPUT_TOKENS = "total_output_tokens"
+    BREAKDOWN = "breakdown"
+    SPAN_ID = "span_id"
+    MODEL_NAME = "model_name"
 
 
 # Test Execution Context Constants
--- a/apps/backend/src/rhesis/backend/app/constants.py
+++ b/apps/backend/src/rhesis/backend/app/constants.py
@@ -258,6 +258,10 @@ class AISpanAttributes:
     # Only spans with this operation type carry countable token usage. Agent-run
     # spans repeat their children's tokens as an aggregate, so counting them too
     # double-counts the trace.
+    #
+    # This also excludes embedding spans, which carry ai.llm.tokens.* but map to
+    # embedding.create. That is deliberate: enrichment never priced embeddings, so
+    # counting their tokens would put tokens and cost back out of step.
     OPERATION_LLM_INVOKE = "llm.invoke"
 
 
--- a/apps/backend/src/rhesis/backend/app/crud/telemetry.py
+++ b/apps/backend/src/rhesis/backend/app/crud/telemetry.py
@@ -910,17 +910,61 @@ def get_trace_metrics_aggregated(
 
     base = db.query(T).filter(*filters).subquery()
 
-    # JSONB extraction expressions for tokens and costs
-    tokens_expr = base.c.attributes[AISpanAttributes.TOKENS_TOTAL].as_float()
-    cost_expr = base.c.enriched_data[EnrichedDataKeys.COSTS][
-        EnrichedDataKeys.TOTAL_COST_USD
-    ].as_float()
+    # Tokens and costs aggregate per *trace*, not per span row.
+    #
+    # enriched_data is a trace-level blob that mark_trace_processed writes onto every
+    # span of the trace, so summing it across span rows multiplies the real figure by
+    # the span count. Collapse to one row per trace first: MAX is exact for the
+    # enriched columns precisely because every row carries the same value.
+    #
+    # raw_tokens is the pre-enrichment fallback, summed over llm.invoke spans only --
+    # the same filter enrichment applies, so an unenriched trace reports the number it
+    # will keep once enrichment lands. Cost has no fallback; it cannot be derived from
+    # span attributes.
+    per_trace = (
+        db.query(
+            base.c.trace_id.label("trace_id"),
+            func.max(
+                base.c.enriched_data[EnrichedDataKeys.COSTS][
+                    EnrichedDataKeys.TOTAL_TOKENS
+                ].as_float()
+            ).label("enriched_tokens"),
+            func.max(
+                base.c.enriched_data[EnrichedDataKeys.COSTS][
+                    EnrichedDataKeys.TOTAL_COST_USD
+                ].as_float()
+            ).label("cost_usd"),
+            func.coalesce(
+                func.sum(
+                    case(
+                        (
+                            base.c.attributes[AISpanAttributes.OPERATION_TYPE].as_string()
+                            == AISpanAttributes.OPERATION_LLM_INVOKE,
+                            base.c.attributes[AISpanAttributes.TOKENS_TOTAL].as_float(),
+                        ),
+                        else_=0,
+                    )
+                ),
+                0,
+            ).label("raw_tokens"),
+        )
+        .group_by(base.c.trace_id)
+        .subquery()
+    )
+
+    token_cost_agg = db.query(
+        func.coalesce(
+            func.sum(
+                func.coalesce(per_trace.c.enriched_tokens, per_trace.c.raw_tokens, 0),
+            ),
+            0,
+        ).label("total_tokens"),
+        func.coalesce(func.sum(per_trace.c.cost_usd), 0).label("total_cost_usd"),
+    ).one()
 
     agg = db.query(
         func.count(func.distinct(base.c.trace_id)).label("total_traces"),
         func.count(base.c.id).label("total_spans"),
-        func.coalesce(func.sum(tokens_expr), 0).label("total_tokens"),
-        func.coalesce(func.sum(cost_expr), 0).label("total_cost_usd"),
         func.count(case((base.c.status_code == "ERROR", 1))).label("error_count"),
         func.coalesce(func.avg(base.c.duration_ms), 0).label("avg_duration_ms"),
         func.coalesce(sqlfunc.percentile_cont(0.5).within_group(base.c.duration_ms), 0).label(
@@ -954,8 +998,8 @@ def get_trace_metrics_aggregated(
     return {
         "total_traces": agg.total_traces or 0,
         "total_spans": total_spans,
-        "total_tokens": int(agg.total_tokens or 0),
-        "total_cost_usd": round(float(agg.total_cost_usd or 0), 6),
+        "total_tokens": int(token_cost_agg.total_tokens or 0),
+        "total_cost_usd": round(float(token_cost_agg.total_cost_usd or 0), 6),
         "error_rate": round(error_count / total_spans, 4) if total_spans else 0,
         "avg_duration_ms": round(float(agg.avg_duration_ms or 0), 2),
         "p50_duration_ms": round(float(agg.p50_duration_ms or 0), 2),
--- a/apps/backend/src/rhesis/backend/app/crud/telemetry.py
+++ b/apps/backend/src/rhesis/backend/app/crud/telemetry.py
@@ -16,7 +16,11 @@ from sqlalchemy import and_, desc, func, or_, select
 from sqlalchemy.orm import Session
 
 from rhesis.backend.app import models
-from rhesis.backend.app.constants import TestExecutionContext
+from rhesis.backend.app.constants import (
+    AISpanAttributes,
+    EnrichedDataKeys,
+    TestExecutionContext,
+)
 from rhesis.backend.app.schemas.telemetry import (
     OTELSpanCreate,
     StatusCode,
@@ -40,6 +44,9 @@ class TraceRow(NamedTuple):
     total: int
     tags_count: int
     comments_count: int
+    # Summed over the trace's llm.invoke spans. Only used when the trace has not
+    # been enriched yet -- see services/telemetry/token_totals.py.
+    llm_tokens: int
 
 
 # ============================================================================
@@ -332,7 +339,30 @@ def query_traces(
         .scalar_subquery()
     )
 
-    # -- Column 3: total — total matching rows *before* LIMIT/OFFSET
+    # -- Column 3: llm_tokens — tokens over this trace's llm.invoke spans.
+    #    The pre-enrichment fallback: enrichment is dispatched asynchronously after
+    #    ingest, so a freshly ingested trace has no enriched_data and would
+    #    otherwise list as 0 while the detail endpoint reported the real figure.
+    #    The llm.invoke filter is the same one enrichment applies, so the two
+    #    numbers agree instead of double-counting aggregate-reporting frameworks.
+    llm_tokens_col = (
+        select(
+            func.coalesce(
+                func.sum(InnerTrace.attributes[AISpanAttributes.TOKENS_TOTAL].as_float()), 0
+            )
+        )
+        .where(
+            and_(
+                InnerTrace.trace_id == models.Trace.trace_id,
+                InnerTrace.organization_id == org_uuid,
+                InnerTrace.attributes[AISpanAttributes.OPERATION_TYPE].as_string()
+                == AISpanAttributes.OPERATION_LLM_INVOKE,
+            )
+        )
+        .scalar_subquery()
+    )
+
+    # -- Column 4: total — total matching rows *before* LIMIT/OFFSET
     #    Uses a window function so pagination total comes from the same query.
     total_col = func.count().over().label("total_count")
 
@@ -360,7 +390,14 @@ def query_traces(
     )
 
     query = (
-        db.query(models.Trace, span_count_col, total_col, tags_count_col, comments_count_col)
+        db.query(
+            models.Trace,
+            span_count_col,
+            total_col,
+            tags_count_col,
+            comments_count_col,
+            llm_tokens_col,
+        )
         .filter(models.Trace.organization_id == org_uuid)
         .options(
             # Scoped to what the loop below actually reads off this chain (existence
@@ -511,7 +548,14 @@ def query_traces(
 
     results = query.order_by(desc(models.Trace.start_time)).limit(limit).offset(offset).all()
     return [
-        TraceRow(trace=r[0], span_count=r[1], total=r[2], tags_count=r[3], comments_count=r[4])
+        TraceRow(
+            trace=r[0],
+            span_count=r[1],
+            total=r[2],
+            tags_count=r[3],
+            comments_count=r[4],
+            llm_tokens=int(r[5] or 0),
+        )
         for r in results
     ]
 
@@ -849,8 +893,6 @@ def get_trace_metrics_aggregated(
 
     from sqlalchemy import case, literal_column
     from sqlalchemy.sql import functions as sqlfunc
-
-    from rhesis.backend.app.constants import AISpanAttributes, EnrichedDataKeys
 
     T = models.Trace
 
--- a/apps/backend/src/rhesis/backend/app/crud/telemetry.py
+++ b/apps/backend/src/rhesis/backend/app/crud/telemetry.py
@@ -12,7 +12,7 @@ from enum import Enum
 from typing import Any, Dict, List, NamedTuple, Optional, Union
 from uuid import UUID
 
-from sqlalchemy import and_, desc, func, or_, select
+from sqlalchemy import and_, asc, desc, func, or_, select
 from sqlalchemy.orm import Session
 
 from rhesis.backend.app import models
@@ -49,6 +49,87 @@ def span_total_tokens_expr(attributes):
         + func.coalesce(attributes[AISpanAttributes.TOKENS_OUTPUT].as_float(), 0.0),
         0.0,
     )
+
+
+# Grid fields the traces list can sort by, mapped to how each is ordered in SQL.
+#
+# Deliberately a whitelist: the columns left out (conversation_input, endpoint_name,
+# trace_metrics_status) have no SQL sort key -- they come from a JSONB attribute, a
+# three-hop join and a relationship -- so the grid marks them unsortable rather than
+# sorting one page of rows and calling it ordered.
+TRACE_SORT_FIELDS = frozenset(
+    {
+        "start_time",
+        "duration_ms",
+        "trace_id",
+        "environment",
+        "root_operation",
+        "span_count",
+        "total_tokens",
+        "total_cost_usd",
+    }
+)
+
+
+def _trace_sort_clauses(
+    sort_by: Optional[str], sort_order: str, span_count_col, llm_tokens_col
+) -> list:
+    """ORDER BY clauses for the traces list.
+
+    Sorting is server-side so a page of results is genuinely the top N of the whole
+    filtered set, not the newest N reshuffled in the browser.
+
+    Always ends with start_time then id. Without a tiebreaker the many rows tied at
+    zero cost have no defined order between pages, and pagination silently repeats
+    and skips rows.
+    """
+    from fastapi import HTTPException
+
+    if sort_by and sort_by not in TRACE_SORT_FIELDS:
+        raise HTTPException(
+            status_code=400,
+            detail=(
+                f"Cannot sort traces by '{sort_by}'. "
+                f"Sortable fields: {', '.join(sorted(TRACE_SORT_FIELDS))}"
+            ),
+        )
+
+    columns = {
+        "start_time": models.Trace.start_time,
+        "duration_ms": models.Trace.duration_ms,
+        "trace_id": models.Trace.trace_id,
+        "environment": models.Trace.environment,
+        "root_operation": models.Trace.span_name,
+        "span_count": span_count_col,
+        # Mirrors what the response actually shows: the enriched total when the trace
+        # has been priced, the llm.invoke sum until then.
+        "total_tokens": func.coalesce(
+            models.Trace.enriched_data[EnrichedDataKeys.COSTS][
+                EnrichedDataKeys.TOTAL_TOKENS
+            ].as_float(),
+            llm_tokens_col,
+            0,
+        ),
+        # Cost has no pre-enrichment fallback, so unpriced traces sort last rather
+        # than leading a "most expensive first" list on a NULLS FIRST default.
+        "total_cost_usd": models.Trace.enriched_data[EnrichedDataKeys.COSTS][
+            EnrichedDataKeys.TOTAL_COST_USD
+        ].as_float(),
+    }
+
+    descending = sort_order.lower() != "asc"
+    clauses = []
+
+    if sort_by:
+        expression = columns[sort_by]
+        ordered = desc(expression) if descending else asc(expression)
+        clauses.append(ordered.nullslast())
+
+    if sort_by != "start_time":
+        clauses.append(desc(models.Trace.start_time))
+    clauses.append(desc(models.Trace.id))
+
+    return clauses
 
 
 class TraceRow(NamedTuple):
@@ -303,6 +384,8 @@ def query_traces(
     test_id: Optional[str] = None,
     conversation_id: Optional[str] = None,
     trace_metrics_status: Optional[str] = None,
+    sort_by: Optional[str] = None,
+    sort_order: str = "desc",
     limit: int = 100,
     offset: int = 0,
 ) -> List[TraceRow]:
@@ -561,7 +644,11 @@ def query_traces(
         )
         query = query.filter(models.Trace.trace_metrics_status_id.in_(matching_status_ids))
 
-    results = query.order_by(desc(models.Trace.start_time)).limit(limit).offset(offset).all()
+    query = query.order_by(
+        *_trace_sort_clauses(sort_by, sort_order, span_count_col, llm_tokens_col)
+    )
+
+    results = query.limit(limit).offset(offset).all()
     return [
         TraceRow(
             trace=r[0],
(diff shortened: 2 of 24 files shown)


apps/backend/src/rhesis/backend/app/crud/telemetry.py
```diff
@@ -332,7 +339,30 @@ def validate_uuid_param(value: Optional[str], param_name: str) -> Optional[UUID]
...
+    # -- Column 3: llm_tokens — tokens over this trace's llm.invoke spans.
+    #    The pre-enrichment fallback: enrichment is dispatched asynchronously after
+    #    ingest, so a freshly ingested trace has no enriched_data and would
+    #    otherwise list as 0 while the detail endpoint reported the real figure.
+    #    The llm.invoke filter is the same one enrichment applies, so the two
+    #    numbers agree instead of double-counting aggregate-reporting frameworks.
+    llm_tokens_col = (
+        select(
+            func.coalesce(
+                func.sum(InnerTrace.attributes[AISpanAttributes.TOKENS_TOTAL].as_float()), 0
```
harry-rhesis: This sums only `ai.llm.tokens.total`. When the key is absent, `->>` yields NULL, the cast yields NULL, and `SUM` skips it, so the span contributes 0. Both Python paths fall back to `input + output` in that case: `token_totals_from_spans` below, and `_span_token_counts` in `enrichment/core.py`. So a pre-enrichment `llm.invoke` span carrying input and output but no total lists as 0 while the drawer shows the real figure, which is the disagreement this column was added to prevent.

To be fair on how far this reaches: I checked every producer and none of the shipped integrations hit it. genai, pydantic-ai, langchain, ADK and haystack all set the total whenever either side is present. It is reachable through the public `create_llm_attributes(tokens_input=100)` with one side only (`tests/sdk/telemetry/test_attributes.py:85` asserts the total is omitted for exactly that input) and through hand set OTLP attributes. It also self corrects once enrichment runs, since `_span_token_counts` derives the total into the blob. So this is a gap in the fallback rather than a live bug, and it is not a regression: before this PR the list showed 0 for those traces anyway.

Suggested:

```python
func.coalesce(
    InnerTrace.attributes[AISpanAttributes.TOKENS_TOTAL].as_float(),
    InnerTrace.attributes[AISpanAttributes.TOKENS_INPUT].as_float()
    + InnerTrace.attributes[AISpanAttributes.TOKENS_OUTPUT].as_float(),
    0,
)
```

Worth noting that no test covers this shape. Every fixture in `test_telemetry_query.py` and `test_trace_metrics_aggregated.py` passes a full `(input, output, total)` triple.

harry-rhesis (commented):
Thank you for this @Arman-Beykmohammadi, it is a genuinely strong piece of work and the description is one of the clearest I have read in this repo. Collapsing four definitions of "this trace's tokens" into one, and deriving it from the same `llm.invoke` filter that cost enrichment already applies, is the right call: tokens and cost can now only agree, or both be wrong together, which is a much better place to be. The per-trace collapse in `get_trace_metrics_aggregated` is correct, and `MAX` is exact there for exactly the reason you give. Trusting the reported total rather than deriving it is right too, and the ADK justification convinced me.

I reviewed the token accrual against the code rather than the description, and cross checked every producer in `sdk/` and `packages/`. **I found nothing that would misreport a trace's tokens for any of the shipped integrations.** Notes below, ordered by what I would act on.

### Worth fixing before merge

1. **The project rollup renders on the test run Traces tab, with whole project and all time totals.** `TracesClient` is also mounted by `TestRunTracesTab`, and in that scope the filters reduce to `{timeRange: 'all', testRunId}`, so the tiles lose both the project narrowing and any time bound. A test run showing 12 traces gets four tiles counting the entire project. Inline on `TracesClient.tsx`.
2. **Embedding tokens are now excluded from every total.** Embedder spans carry `ai.llm.tokens.*` but are not `llm.invoke`, so the detail endpoint, which used to sum every span, no longer counts them. I think the new behaviour is correct, but it is a silent change and deserves a line in the out of scope list. Inline on `constants.py`.

### Worth fixing, low urgency

3. **The SQL fallback does not derive a missing total from `input + output`, while both Python paths do.** Narrower than it first looks: no shipped integration produces that shape, and it self corrects once enrichment runs. Still a hole in the exact window the fallback exists to cover, and no test touches it. Inline on both `crud/telemetry.py` sites.
4. **Two leftovers.** `Trace.total_tokens` now has no reader anywhere and still returns the raw per span attribute, which is the shape that caused the original bug. `TraceSummary.total_input_tokens` and `total_output_tokens` are sent on every list row and read by nothing. Inline on `token_totals.py` and `schemas/telemetry.py`.

### Notes, no action needed

5. A staleness window on traces that are still growing, which I confirmed is self healing. It does make one sentence in the description slightly too strong. Inline on `token_totals.py`.
6. A naming nit in the sum comprehensions. Inline on `enrichment/core.py`.

### On testing

I could not run either suite in my environment, so the 204 backend and 847 frontend results are your verification, not mine. Everything above comes from reading the code and tracing the producers. The manual end to end checks you describe are exactly the right ones, and the "not verified: the rendered chips in a browser" line is appreciated honesty.

apps/backend/src/rhesis/backend/app/crud/telemetry.py
```diff
@@ -868,17 +910,61 @@ def get_trace_metrics_aggregated(
...
+                    EnrichedDataKeys.TOTAL_COST_USD
+                ].as_float()
+            ).label("cost_usd"),
+            func.coalesce(
+                func.sum(
+                    case(
+                        (
+                            base.c.attributes[AISpanAttributes.OPERATION_TYPE].as_string()
+                            == AISpanAttributes.OPERATION_LLM_INVOKE,
+                            base.c.attributes[AISpanAttributes.TOKENS_TOTAL].as_float(),
```
harry-rhesis: Same gap as `llm_tokens_col` above: `raw_tokens` sums only the reported total, so an `llm.invoke` span with input and output but no total contributes nothing to the rollup tile while the drawer reports it.

The same `COALESCE(total, input + output, 0)` here keeps the tile, the list and the drawer aligned through the pre-enrichment window.

apps/frontend/src/app/(protected)/traces/components/TracesClient.tsx
```diff
@@ -200,6 +222,15 @@ export default function TracesClient({
         </Alert>
       )}
 
+      <TraceMetricsSummary
```
harry-rhesis: This renders unconditionally, and `TracesClient` is also mounted by `TestRunTracesTab.tsx:72` with `fixedTestRunId`. In that scope `sanitizeTraceDrawerFiltersForTestRunScope` (`trace-filter-params.ts:87`) returns `{timeRange: 'all', testRunId, traceMetricsStatus, testResultId, testId}`, which means:

* `drawerFilters.projectId` is unset, so `rollupProjectId` falls back to `scopedProjectId`, which is set on a test run page
* `timeRange: 'all'` means `buildTraceQueryParams` emits neither `start_time_after` nor `start_time_before`

So a test run listing 12 traces gets four tiles counting the whole project across all time. The caption tells the reader the tiles are broader than the table, which is a good instinct, but it does not really cover numbers that have nothing to do with the page they are on.

Simplest fix is to skip the rollup in test run scope:

```tsx
{!fixedTestRunId && (
  <TraceMetricsSummary
    projectId={rollupProjectId}
    ...
  />
)}
```

Everything else about the component looks right to me, including the `useEffect` choice and the reasoning for it.

apps/backend/src/rhesis/backend/app/constants.py
```diff
@@ -251,8 +251,15 @@ class AISpanAttributes:
...
     OPERATION_TYPE = "ai.operation.type"
     MODEL_NAME = "ai.model.name"
+    TOKENS_INPUT = "ai.llm.tokens.input"
+    TOKENS_OUTPUT = "ai.llm.tokens.output"
     TOKENS_TOTAL = "ai.llm.tokens.total"
 
+    # Only spans with this operation type carry countable token usage. Agent-run
+    # spans repeat their children's tokens as an aggregate, so counting them too
+    # double-counts the trace.
+    OPERATION_LLM_INVOKE = "llm.invoke"
```
harry-rhesis, line 265: Agreed that `llm.invoke` is the right basis, and picking the filter cost enrichment already uses is what makes the whole change hold together. One consequence is worth calling out in the description, because it is invisible otherwise.

Embedding spans carry `ai.llm.tokens.*` but are not `llm.invoke`, so their tokens are now excluded everywhere. Concretely:

* Haystack `Embedder` components map to `ai.embedding.generate` via `SPAN_KIND_RULES` in `haystack/mapping.py`, and therefore to `ai.operation.type = embedding.create`, while `haystack/tracer.py:810` sets `ai.llm.tokens.*` on that same span from `usage` or `billed_units`.
* The same holds for any GenAI convention producer: `genai.py:147` maps `OP_EMBEDDINGS` to `embedding.create` and then sets the token attributes unconditionally just below.

The detail endpoint used to include those tokens because it summed every span. It no longer does. I think the new behaviour is correct, since enrichment never priced embeddings either and tokens now stay consistent with cost, but someone will notice their embedding tokens disappear from a trace they have looked at before. A line in the deliberately out of scope list, next to the SDK agent gap, would cover it.

apps/backend/src/rhesis/backend/app/schemas/telemetry.py
```diff
@@ -103,6 +103,8 @@ class TraceSummary(BaseModel):
     root_operation: str
     status_code: str
     total_tokens: Optional[int] = None
+    total_input_tokens: Optional[int] = None
```
harry-rhesis, line 106: These two are populated on every list row and declared in `telemetry.ts:111-112`, but nothing reads them. The drawer tooltip uses `TraceDetailResponse`, not `TraceSummary`.

Your own docstring on `trace_summary_totals` explains why the split necessarily stays zero here before enrichment, which means a caller cannot really rely on them as they stand. Either drop them from `TraceSummary`, or feed them into a tooltip on the new Tokens column so the list gets the same detail the drawer has.

apps/backend/src/rhesis/backend/app/services/telemetry/token_totals.py
```diff
@@ -0,0 +1,132 @@
...
+from rhesis.backend.app.constants import AISpanAttributes, EnrichedDataKeys
+from rhesis.backend.app.models.trace import Trace
+
+# input, output, total
+TokenTotals = Tuple[int, int, int]
+
+ZERO_TOKENS: TokenTotals = (0, 0, 0)
+
+
+def token_totals_from_enriched(enriched_data: Optional[dict]) -> Optional[TokenTotals]:
```
harry-rhesis, line 19: Now that this module is the single accessor, `Trace.total_tokens` (`models/trace.py:183`) has no remaining reader anywhere in `apps/` or `tests/`. I grepped to confirm. The one hit that looks like a reader, `enrichment/core.py:146`, is `CostBreakdown.total_tokens`, not the model property.

It still returns the raw per span attribute, which is precisely the shape that produced the original bug, so leaving it in place is a live trap for whoever reaches for it next. Worth deleting in this PR while the context is fresh, or at minimum pointing its docstring here.

apps/backend/src/rhesis/backend/app/services/telemetry/token_totals.py
```diff
@@ -0,0 +1,132 @@
...
+        totals[0] += input_tokens
+        totals[1] += output_tokens
+        # Trust the reported total; Google ADK folds cache-read tokens into it, so
+        # deriving it from input + output would quietly discard those.
+        totals[2] += input_tokens + output_tokens if reported_total is None else int(reported_total)
+
+    return (totals[0], totals[1], totals[2])
+
+
+def trace_token_totals(spans: Sequence[Trace]) -> TokenTotals:
```
harry-rhesis, line 78: Note rather than a request, and I think the current behaviour is the right one.

This prefers the enrichment blob with no freshness check, so on a trace that is still growing (multi turn conversations keep adding spans under one `trace_id`) every surface reports the pre-growth total until re-enrichment lands.

I checked that it does land: `post_ingest_link.delay(...)` fires on every ingest POST (`routers/telemetry.py:210`), not once per trace, and `enrich_trace` re-runs whenever any span has `processed_at is None` (`processor.py:73-85`). So the window is one enrichment cycle, and every surface agrees on the stale value meanwhile, which is the property that actually matters.

It does mean the description line "the token total does not change when enrichment lands" holds for a trace that has stopped growing, but not for one mid conversation. Might be worth softening so nobody reads it as a stronger guarantee than it is.

apps/backend/src/rhesis/backend/app/services/telemetry/enrichment/core.py
```diff
@@ -136,6 +141,9 @@ def calculate_token_costs(spans: List[Trace]) -> Optional[TokenCosts]:
     return TokenCosts(
         total_cost_usd=round(total_cost_usd, 6),
         total_cost_eur=round(total_cost_eur, 6),
+        total_input_tokens=sum(span.input_tokens for span in cost_breakdown),
+        total_output_tokens=sum(span.output_tokens for span in cost_breakdown),
+        total_tokens=sum(span.total_tokens for span in cost_breakdown),
```
harry-rhesis: Small naming nit. `span` here iterates `CostBreakdown` objects rather than `Trace` objects, which reads oddly a few lines under a function whose other loops are over real spans. `entry` or `b` would save the double take.

harry-rhesis:
## `trivy (backend)` failure: infrastructure, not this change

Short version: this is a CI flake, there is nothing here for you to fix, and re-running the job should clear it. Flagging the detail anyway, since "failing security check" is worth being precise about.

**The job never scanned anything.** It failed in the action's own setup step, `Install Trivy`:

```
aquasecurity/trivy info checking GitHub for tag 'v0.69.2'
aquasecurity/trivy info found version: 0.69.2 for v0.69.2/Linux/64bit
##[error]Process completed with exit code 35.
```

That is `bash ./trivy/contrib/install.sh -b /home/runner/.local/bin/trivy-bin v0.69.2` failing while downloading the release tarball, after it had already resolved the version successfully. Exit code 35 is curl's `CURLE_SSL_CONNECT_ERROR`, a TLS handshake failure. Every step after it, `Run Trivy` included, is marked `skipped` in the log.

Three things make it conclusive rather than a guess:

1. **`trivy (frontend)` passed in the same run.** Run `33871477968`, started one second earlier at `12:11:02`, same action pin (`aquasecurity/trivy-action@57a97c7`), same `version: v0.69.2`. It installed and scanned fine. Same runner image, same minute, different outcome.
2. **A vulnerability finding cannot fail this job by design.** Both Trivy steps in `.github/workflows/security.yml` set `exit-code: 0`. Findings flow to the SARIF upload and the Security tab, and the table output is informational. The only way this job goes red is infrastructure.
3. **The diff introduces no dependency surface.** 21 files, all `.py`, `.ts` and `.tsx`, with no `pyproject.toml`, `uv.lock` or `package.json` among them. There would have been nothing new for Trivy to flag even if it had run.

So a re-run of `trivy (backend)` is all this needs. Happy to kick it off if you would rather not.

**One durable note, for a separate CI change and not this PR:** the install step has no retry, so any transient failure fetching the binary from the release CDN turns a security check red for reasons unrelated to the code under review. `aquasecurity/trivy-action` supports caching the binary, which would make this rarer. Worth an issue if it shows up again.

Unrelated, but worth recording alongside my review: `Backend tests` passed in CI at 9m10s on this head SHA, which confirms the suite result I could not run locally.

apps/backend/src/rhesis/backend/app/crud/telemetry.py
Arman-Beykmohammadi: Fixed in d7dbe4f75. Both SQL sites now go through one shared span_total_tokens_expr, so they cannot drift from each other again:

COALESCE(NULLIF(total, 0), COALESCE(input, 0) + COALESCE(output, 0), 0)

Two things came out of writing it. Each side of the sum needs its own COALESCE, because NULL + 100 is NULL in SQL, so a span that reported only input would still have contributed nothing. That is the create_llm_attributes(tokens_input=100) case you pointed at, so without it the fix would have missed the exact shape you found.

The NULLIF is peqy asking the same question from the other end: a reported total of 0 next to non-zero input or output. I treat it as missing too, and applied the same rule in token_totals_from_spans and _span_token_counts so SQL and Python cannot disagree about it.

You were right that no fixture covered this. TestTokenTotalFallbackShapes in test_telemetry_query.py now covers a missing total, only one side present, a zero total beside real usage, and the ADK case where the reported total is larger and has to win.

apps/backend/src/rhesis/backend/app/crud/telemetry.py
Arman-Beykmohammadi: Same fix here, in d7dbe4f75. raw_tokens calls the same span_total_tokens_expr the list column uses, so the tile, the list and the drawer stay on one definition through the pre-enrichment window.

apps/frontend/src/app/(protected)/traces/components/TracesClient.tsx
Arman-Beykmohammadi: Good catch, and thanks for tracing it all the way through sanitizeTraceDrawerFiltersForTestRunScope. Fixed in d739e8be5 with the gate you suggested.

I agree the caption was the wrong tool for this. It is fine for "these totals are wider than your filters", but a test run page showing whole-project numbers next to twelve rows is not a wider version of the same thing, it is a different question answered.

The fixedTestRunId term in hasUnsupportedRollupFilters is gone too, since it can no longer be reached.

apps/backend/src/rhesis/backend/app/constants.py
Arman-Beykmohammadi: Agreed, and thanks for digging out the concrete producers. I have added it to the out of scope list in the description, and put a note in constants.py next to OPERATION_LLM_INVOKE so the next person reading the filter sees why embeddings are not in it.

Your reasoning is the reason I am comfortable with the change: enrichment never priced embedding spans, so counting their tokens would put tokens and cost back out of step, which is the thing this PR exists to stop.

apps/backend/src/rhesis/backend/app/services/telemetry/token_totals.py
Arman-Beykmohammadi: Deleted in d7dbe4f75. I grepped apps/ and tests/ before removing it and agree the only hit that looks like a reader is CostBreakdown.total_tokens.

Leaving it would have been worse than dead code, since it returns exactly the per-span shape that caused the bug this PR is fixing, sitting on the model where it looks like the obvious thing to reach for.

apps/backend/src/rhesis/backend/app/schemas/telemetry.py
Arman-Beykmohammadi: Took the tooltip option, in 85351fcda. The Tokens cell now carries the same input and output split the drawer chip shows, through the existing tokenSplitLabel.

You are right that a caller cannot lean on them before enrichment. tokenSplitLabel returns undefined when both are zero, so in that window the cell just has no tooltip rather than one reading "0 input, 0 output".

apps/backend/src/rhesis/backend/app/services/telemetry/token_totals.py
Arman-Beykmohammadi: Fair, and thanks for confirming it self heals. I have softened that line in the description to say it holds once a trace has stopped growing, and to name the one enrichment cycle where a multi turn conversation reports the pre-growth total.

The property I actually care about is the one you identified, that every surface agrees on the same stale value rather than each showing a different one, so I have said that explicitly instead of implying something stronger.

apps/backend/src/rhesis/backend/app/services/telemetry/enrichment/core.py
Arman-Beykmohammadi: Renamed to entry in d7dbe4f75. You are right that it reads badly that close to loops over real spans.

Arman-Beykmohammadi:
Both addressed in d7dbe4f75.

On the SQL fallback: agreed, and fixed at both sites. They now share one span_total_tokens_expr so they cannot drift:

COALESCE(NULLIF(total, 0), COALESCE(input, 0) + COALESCE(output, 0), 0)

One wrinkle worth recording. Each side of the sum needs its own COALESCE, because NULL + 100 is NULL in SQL. Without that, a span reporting only input would still have contributed nothing, which is one of the shapes the fix was meant to catch.

On the question about a reported total of 0 while input and output are above zero: no shipped integration produces it. genai, pydantic-ai, langchain, ADK and haystack all set the total whenever either side is present. It is reachable through create_llm_attributes with a single side, and through hand set OTLP attributes.

I went with your suggestion anyway. The two readings contradict each other, and unlike the ADK case where a larger total is meaningful (cache-read tokens folded in), a zero total beside real usage has no honest interpretation. Treating it as missing can only ever move the number toward the truth. The NULLIF does it in SQL, and token_totals_from_spans and _span_token_counts do the same in Python so the paths agree.

Covered by TestTokenTotalFallbackShapes in test_telemetry_query.py: missing total, one side only, zero total beside real usage, and the ADK case where the larger reported total has to win.

PR #2713, opened 2026-09-11
Join MAF conversation turns into one trace

harry-rhesis:
## Purpose

A multi-turn Microsoft Agent Framework conversation arrived in Rhesis as one unrelated trace per turn. Every turn was correctly labelled with the same conversation id, but nothing put them together, so the Conversation tab showed a one-turn conversation per trace instead of the session.

MAF opens its own root span per turn, so OpenTelemetry mints a fresh trace id each time. The Google ADK integration already solves this, and the machinery it uses (`ConversationTraceRegistry`, `RetracedSpan`) lives in the shared `genai.py`. MAF simply never adopted it.

Fixing that surfaced two more problems that had to be fixed for the join to be an improvement rather than a regression. Both are described below.

## What Changed

**`fix(sdk): stop one maf exporter starving the others`**

`conversation_root_attributes` popped the per-trace conversation content as it stamped a root span. `enable()` wraps *every* exporter on the provider, so a process running its own OTLP collector beside Rhesis has two translating exporters exporting the same spans: whichever reached the trace root first took the content, and the other stamped a root span carrying no conversation at all. Which one won was arbitrary.

The content is now read rather than popped, and released a bounded number of reads later. Deferring keeps entries alive long enough for the other exporters to see the same trace; the bound keeps a long-running process from holding 10 KB of conversation text per trace until the 4096-entry cap evicts it. The queue is larger than an OTEL export batch, since a busy service can flush hundreds of run roots at once and the exporters do not drain their queues in lockstep.

This is a pre-existing bug, not one introduced here. It is first in the branch because the join cannot be verified end to end while it stands.

**`docs(sdk): correct what the maf conversation reader does`** and **`refactor(sdk): use the shared release queue in the maf registry`**

The docstring still pointed at `_ConversationContentRegistry.consume`, renamed in this branch, so the cross-reference resolved to nothing, and "always consumes" was wrong twice over. It now says which span releases and which does not, including the one path the release queue does not reach: a plain `agent.run()` under an enclosing Rhesis span leaves its entries to the store's eviction cap. That path is pre-existing, unchanged here, and bounded; see the note below.

The refactor then drops the private queue for `DeferredReleaseQueue` (added in #2718, now merged), so MAF and ADK share one implementation.

**`fix(sdk): release maf conversation content nothing will read`**

A run under a Rhesis `@endpoint`/`@observe` span has its turn owned by that span, so no MAF span stamps it, and nothing freed the per-trace content either. A plain `agent.run()` emits no `workflow.run`, so the one guard that released anything never fired and the entries sat until the store's 4096-entry cap evicted them, which is the leak the release queue exists to prevent.

Releasing from any nested run root would be wrong the other way round: inside a MAF-rooted trace the root is the reader and exports after its children. So the fix tracks, at span start, whether a trace's root is a MAF span, and releases from a nested run root only when it is not. One entry per trace, bounded.

That rule also corrects the previous one, which released on every nested `workflow.run`, including a sub-workflow under a `workflow.run` trace root that still had to read its content. Span start is the only moment the answer is available: a trace root starts before the nested spans that ask about it, and exports after them.

**Still not covered**, and called out in the commit: a trace whose only MAF span is a bare `chat` span, from calling a chat client with no agent around it. There is no run root to release from, so those entries still wait for the entry cap.

**`fix(sdk): let an empty maf read keep its release slot`**

A read of a trace with nothing recorded took a slot in the release queue, so a run of content-free traces (single-turn runs, or content capture switched off) pushed the entries of traces that *do* have content out of the queue before the other exporters had read them. The check covers every store: leaving one out makes its entries invisible to the queue, so they are never freed and the store grows until the entry cap evicts it, which is the leak the queue exists to prevent.

**`fix(sdk): join maf conversation turns into one trace`**

Claim each turn's trace at span start, where the conversation ContextVars are still readable, and rewrite later turns onto the first turn's id at export. The first turn keeps the id it was born with, so anything that already observed or published it still resolves.

Two cases stand the rewrite down, because the trace id is already owned and published elsewhere: an enclosing `@endpoint`/`@observe` span, and a platform-driven turn. Both join the turns themselves, and moving spans off that id would strand the link and the reply on a trace with no spans.

The same commit also stamps the turn root on a root `invoke_agent` span. Only `workflow.run` carried it, and a plain `agent.run()` emits no workflow span at all. Without this the join would have put a plain-agent chat's turns on one trace with **no** conversation attributes anywhere on it, which is worse than leaving them apart. A nested agent span is deliberately left alone: inside a workflow that is every agent in the run, and the first one would release the content the root needs.

Docs: a "Multi-Turn Conversations" section on the MAF page, covering the binding, the ordering requirement, and the two stand-down cases.

## Additional Context

- Independent of #2711. It touches no file that PR touches, so it applies to `main` on its own.
- Uses the shared `DeferredReleaseQueue` from #2718, which has merged, so this branch sits on `main` directly and stands alone.
- #2715 fixes the same pop-on-read flaw in the Google ADK integration. Both now consume the shared queue rather than keeping a copy each.
- peqy's third point, that a nested `invoke_agent` under an `@endpoint`/`@observe` span never releases, is fixed in the last commit rather than deferred. Worth noting `AncestryRegistry` could not have closed it: `record()` only indexes spans that have a parent, so a root `workflow.run` is absent from it and "is my parent a MAF span" would wrongly answer no and release early. Keying on "does a MAF span root this trace" avoids that and costs one entry per trace instead of one per span.
- peqy also asked whether `release()` should free entries immediately. It should not: a nested `workflow.run` inside a root one would destroy the content before the root reads it, since children export first. That is the bug the first commit fixes.
- The join's limit is documented in a test: without a bound conversation id, each turn keeps its own trace. The target trace has to be known when the run root is created, and MAF assigns no span attributes before then, so only the Rhesis ContextVar is readable at that point.

## Testing

`3036 passed` in `tests/sdk` (excluding `tests/sdk/integration`, which needs Docker). Ruff check and format clean. Each commit was verified to pass standing alone, so there is no broken intermediate to bisect through.

The MAF suite grew from 74 to 118 tests: 9 end-to-end tests driving real MAF agents through the join, plus exporter-level tests for the agent-root stamping and the registry's read/release contract.

Every behavioural change was mutation-tested, and each mutation is caught by the test written for it:

| Mutation | Caught by |
|---|---|
| Remove the claim | `test_turns_of_one_conversation_share_one_trace` |
| Remove the retrace at export | `test_turns_of_one_conversation_share_one_trace` |
| Drop the ContextVar guard | `test_a_platform_owned_conversation_is_left_alone` |
| Revert the agent-root stamp | `test_exporter_stamps_turn_root_on_root_agent_span` |
| Let a nested agent span consume | `test_a_nested_agent_span_leaves_the_content_for_the_root` |
| Make the registry read pop again | `test_two_wrapped_exporters_both_stamp_the_conversation`, and the end-to-end join test |
| Never release a served entry | `test_conversation_content_registry_releases_what_falls_behind` |
| Let an empty read take a slot | `test_conversation_content_registry_ignores_a_trace_with_nothing_recorded` |
| Never release from a nested run root (the old leak) | `test_a_nested_run_root_releases_when_no_maf_span_roots_the_trace` |
| Always release from a nested run root | `test_a_nested_run_root_leaves_the_content_when_a_maf_span_roots_the_trace` |
| Restore the old "release on any nested workflow.run" rule | both of the above |
| Never mark the trace root at span start | `test_a_parentless_maf_span_marks_its_trace` |
| Leave the marker set unbounded | `test_maf_rooted_traces_is_bounded` |
| Drop any one store from the guard | `test_conversation_content_registry_releases_whichever_store_holds_the_content`, one case per store |

One note on the test harness, since it cost me time and may cost a reviewer some. The MAF `session_provider` fixture rides whatever provider is already global, which already carries a `BatchSpanProcessor`, so the suite genuinely runs with two wrapped exporters. That is what made the popping bug observable. Separately, a test that never drains leaves its spans buffered and they arrive in the next test's flush, which reads as a turn that failed to join. The `drained_exporter` fixture handles that and says why.

I could not build the docs site to check the new page renders (`docs/src/node_modules` is absent). I verified `CodeBlock` is imported on the page and the block matches the existing pattern; a reviewer running the site locally is the real check.

One note on how that last batch was verified. "Never mark the trace root" initially survived every test, because the release tests marked traces by hand and so never exercised `on_start`. Driving the processor's `on_start` directly closed it. Worth knowing as a reviewer that an early release is not immediate data loss either way: `release()` queues, so the root keeps its content unless several hundred traces pass in between.

### DIFF
--- a/docs/content/docs/tracing/agent-framework.mdx
+++ b/docs/content/docs/tracing/agent-framework.mdx
@@ -153,6 +153,23 @@ async def chat(input: str, conversation_id: str = None) -> dict:
   Do not call `agent_framework.observability.configure_otel_providers()` (or `setup_observability()`) after creating the `RhesisClient`. Doing so replaces the Rhesis tracer provider and spans will no longer reach Rhesis. `auto_instrument("agent_framework")` already enables MAF instrumentation against the Rhesis provider for you.
 </Callout>
 
+## Multi-Turn Conversations
+
+MAF opens its own root span per turn, so OpenTelemetry mints a fresh trace id every time you run an agent or a workflow. Bind a conversation id and the integration puts all of that conversation's turns on one trace, with the run root of each turn carrying the turn's message and reply:
+
+<CodeBlock filename="session.py" language="python">
+{`from rhesis.telemetry.context import set_conversation_id
+
+set_conversation_id("user-42")
+
+for message in ["Hello", "Tell me more"]:
+    result = await agent.run(message)`}
+</CodeBlock>
+
+The first turn keeps the trace id it was born with and later turns join it, so a link to the first turn still resolves. Bind the id before the run starts: the target trace has to be known when the run root is created, and MAF assigns no span attributes before then.
+
+Two cases stand this down, because there the trace id is already owned and published elsewhere: a run inside an `@endpoint` or `@observe` span, and a platform-driven turn. Both join the turns themselves, so the integration leaves their spans where they are.
+
 ## Worked Example: Travel Agent
 
 The repository ships a runnable multi-agent MAF demo - a `HandoffBuilder` workflow with a coordinator and three specialists - built specifically to produce agent, LLM, tool, and handoff traces in Rhesis. See [`agents/travel-agent`](https://github.com/rhesis-ai/rhesis/tree/main/agents/travel-agent) and its [architecture notes](https://github.com/rhesis-ai/rhesis/blob/main/agents/travel-agent/docs/architecture.md).
--- a/sdk/src/rhesis/sdk/telemetry/integrations/agent_framework/mapping.py
+++ b/sdk/src/rhesis/sdk/telemetry/integrations/agent_framework/mapping.py
@@ -87,6 +87,10 @@ WORKFLOW_RUN_SPAN_NAME = "workflow.run"
 # Graph View knows how to connect.
 HANDOFF_TOOL_PREFIX = "handoff_to_"
 
+# MAF names an agent span after the operation that opened it, so the agent's own
+# name follows the prefix (``invoke_agent researcher``).
+_AGENT_SPAN_PREFIXES: tuple[str, ...] = (f"{OP_INVOKE_AGENT} ", f"{OP_CREATE_AGENT} ")
+
 # Operation -> Rhesis span name. The validator in
 # :mod:`rhesis.telemetry.attributes` accepts ``ai.<domain>(.<action>)?`` and
 # anything starting with ``function.``, so we are careful to land here.
@@ -339,3 +343,12 @@ def translate_handoff_attributes(
 def is_workflow_run_span(original_name: str | None) -> bool:
     """Return True for MAF's top-level ``workflow.run`` span (the run root)."""
     return original_name == WORKFLOW_RUN_SPAN_NAME
+
+
+def is_agent_invocation_span(original_name: str | None) -> bool:
+    """Return True for MAF's per-run agent span (``invoke_agent <name>``).
+
+    A plain ``agent.run()`` emits no ``workflow.run`` span, so this is the trace
+    root there and the only span that can carry the turn's conversation.
+    """
+    return bool(original_name) and original_name.startswith(_AGENT_SPAN_PREFIXES)
--- a/sdk/src/rhesis/sdk/telemetry/integrations/agent_framework/translator.py
+++ b/sdk/src/rhesis/sdk/telemetry/integrations/agent_framework/translator.py
@@ -35,6 +35,7 @@ from rhesis.sdk.telemetry.integrations.agent_framework import mapping
 from rhesis.sdk.telemetry.integrations.genai import (
     AncestryRegistry,
     ConversationTraceRegistry,
+    DeferredReleaseQueue,
     RetracedSpan,
 )
 from rhesis.sdk.telemetry.integrations.genai import (
@@ -380,10 +381,15 @@ class _ConversationContentRegistry:
         self._output_by_trace: dict[int, str] = {}
         self._session_by_trace: dict[int, str] = {}
         self._max_entries = max_entries
-        # Traces whose content has been handed to an exporter, oldest first.
-        self._served: dict[int, None] = {}
-        self._max_served = max_served
         self._lock = threading.Lock()
+        # Every store goes in: one left out never gets freed. Guarded by the
+        # lock above, which the queue expects the caller to hold.
+        self._release = DeferredReleaseQueue(
+            self._session_by_trace,
+            self._input_by_trace,
+            self._output_by_trace,
+            max_served=max_served,
+        )
 
     def _evict_if_needed(self, store: dict[int, str]) -> None:
         if len(store) < self._max_entries:
@@ -462,7 +468,7 @@ class _ConversationContentRegistry:
             session = self._session_by_trace.get(trace_id)
             conv_input = self._input_by_trace.get(trace_id)
             conv_output = self._output_by_trace.get(trace_id)
-            self._queue_release(trace_id)
+            self._release.queue_release(trace_id)
         return session, conv_input, conv_output
 
     def release(self, trace_id: int | None) -> None:
@@ -474,44 +480,7 @@ class _ConversationContentRegistry:
         if trace_id is None:
             return
         with self._lock:
-            self._queue_release(trace_id)
-
-    def _queue_release(self, trace_id: int) -> None:
-        """Drop the entries of whatever has fallen far enough behind.
-
-        Deferred rather than immediate so every exporter reading the same batch
-        sees the same content, and bounded so a long-running process does not
-        hold 10 KB of conversation text per trace until the entry cap evicts it.
-        The queue is deliberately larger than an OTEL export batch, since a busy
-        service can flush hundreds of run roots at once and the exporters do not
-        drain their queues in lockstep.
-
-        Caller holds the lock.
-        """
-        if not self._has_entries(trace_id):
-            # Nothing to release, so it must not take a slot: a run of traces
-            # with no conversation content on them would otherwise push out the
-            # entries of the ones that have it.
-            return
-        self._served.pop(trace_id, None)
-        self._served[trace_id] = None
-        while len(self._served) > self._max_served:
-            stale = next(iter(self._served))
-            del self._served[stale]
-            self._session_by_trace.pop(stale, None)
-            self._input_by_trace.pop(stale, None)
-            self._output_by_trace.pop(stale, None)
-
-    def _has_entries(self, trace_id: int) -> bool:
-        """Whether any of the three stores holds something for this trace."""
-        return any(
-            trace_id in store
-            for store in (
-                self._session_by_trace,
-                self._input_by_trace,
-                self._output_by_trace,
-            )
-        )
+            self._release.queue_release(trace_id)
 
 
 # Shared singleton: the dedup processor records chat I/O at span end, the
--- a/sdk/src/rhesis/sdk/telemetry/integrations/agent_framework/translator.py
+++ b/sdk/src/rhesis/sdk/telemetry/integrations/agent_framework/translator.py
@@ -358,17 +358,25 @@ class _ConversationContentRegistry:
     It is the gate for turn-root stamping: only when the agent has set a real
     conversation id for the run does the root ``workflow.run`` span become a
     Rhesis conversation turn root. Pure single-turn runs (e.g. the ``run_traces``
-    smoke test) set no id and stay plain traces. All stores are bounded; the
-    oldest entries are evicted once the cap is hit. Input/output strings are
+    smoke test) set no id and stay plain traces.
+
+    Entries are released a bounded number of reads after the first exporter
+    reads them, rather than on that read, so that several wrapped exporters all
+    stamp the same content; see :meth:`read_for_root`. Every store is bounded
+    independently of that, the oldest entries being evicted once the cap is hit.
+    Input/output strings are
     truncated to :data:`~rhesis.telemetry.constants.ConversationContext.MAX_IO_LENGTH`
     at record time.
     """
 
-    def __init__(self, max_entries: int = 4096) -> None:
+    def __init__(self, max_entries: int = 4096, max_served: int = 512) -> None:
         self._input_by_trace: dict[int, str] = {}
         self._output_by_trace: dict[int, str] = {}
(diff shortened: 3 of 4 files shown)


Arman-Beykmohammadi:
Reviewed this one alongside #2714 and #2715, since the three overlap.

**The problem is real and the fix is the right shape.** A multi-turn MAF conversation arriving as one trace per turn is a genuine gap, claim-at-start plus retrace-at-export is the correct approach given MAF opens its own root per turn, and reusing what ADK already does rather than inventing a third mechanism is the right instinct. Standing the rewrite down for `root_trace_id` and `conversation_trace_id` is exactly right, and the reasoning for why those two vars are the signal is worth keeping in the comment. Stamping the turn root on a root `invoke_agent` so a plain `agent.run()` does not join into a trace with no conversation on it anywhere is a good catch, and the fact that you found it while checking the join rather than after shipping it shows.

I ran the suite on the branch: 118 passed, matching what you have in the description. The branch is 55 commits behind main but nothing on main has touched `sdk/src/rhesis/sdk/telemetry/` or `packages/rhesis/src/rhesis/telemetry/` since the merge base, so it applies cleanly.

**I would not merge it as-is.** Nothing here is unsafe, and if you disagree with me on all of the below I would still rather have this in than out. But there are a few things I would settle first, and one of them I think is worth pushing back on properly.

## 1. `_MafRootedTraces` does not protect anything observable

This is my main concern. The justification in the docstring and in the commit message is that a nested run root releasing would take content the trace root still needs. That was true before `DeferredReleaseQueue`, and it is what the first commit on this branch fixes. After that commit it is no longer true: `queue_release` does not free anything, and re-reading a trace pops and reinserts it at the front of the queue, so the root's own read refreshes the entry no matter what a nested span did earlier.

I tested this rather than argue it. I made `_MafRootedTraces.contains` return `False` unconditionally, which is the same as always releasing from a nested run root, and ran the tests:

- `test_the_root_still_reads_after_a_nested_span_exported` passes. The root still gets `conversation.id` and the input, unchanged.
- The only failures are the two parametrisations of `test_a_nested_run_root_leaves_the_content_when_a_maf_span_roots_the_trace`, and they fail on `assert not self._queued_for_release(trace_id)`, which is an assertion about queue internals rather than about anything the backend receives.

So the behavioural test the guard exists for does not notice the guard, and the tests that do notice it are testing the mechanism rather than an outcome. The window it closes needs 512 distinct content-bearing traces read between a nested export and its root's export, and in practice those two are in the same batch.

You noted this yourself under Additional Context, that an early release is not immediate data loss because `release()` queues. I think that observation is right and should be followed to its conclusion: drop the class, the module global, the `on_start` branch and the tests that go with it, and always queue on a run-root export.

That is also what #2715 does in ADK for the same case, where a nested `AgentTool` `invocation` queues unconditionally. Right now the two integrations make opposite choices about one problem in the same batch of PRs, while both descriptions say the registries behave identically apart from the stores they own. ADK's version is the simpler one and it is sufficient, so I would rather MAF matched it than the other way round.

To be clear about what I am not saying: the leak you were closing is real, and the `@endpoint`/`@observe` case where nothing reads does need releasing. Always queueing handles that case too.

## 2. Two conversation stores that do not know about each other

This is the cross-PR one, and it is the concern I care most about after the point above.

After these three land there are two mechanisms for putting a conversation's turns on one trace, with two separate stores:

- `_anchors` in `packages/rhesis/src/rhesis/telemetry/conversation.py`, anchor plus synthetic parent, trace id fixed at span creation. Used by `conversation_turn`, `@endpoint`/`@observe`, LangGraph, and now Haystack.
- `ConversationTraceRegistry` in `genai.py`, claim plus retrace at export. Used by ADK, and now MAF.

Having two mechanisms is fine and I think unavoidable: one needs to own span creation, the other is for frameworks whose root span we do not create. Having two stores is the part I would like to change. `get_conversation_anchor` and `anchor_conversation` were exposed in #2711 with a docstring that says they exist so an integration can join the same anchor "rather than keeping a second store that would disagree with it for the same conversation id", and this PR adds a second consumer to the store that does exactly that.

It is reachable, not theoretical. Open turn one with `conversation_turn("c-1")`, then run turn two as a bare `agent.run()` under `set_conversation_id("c-1")` outside any turn. The shared store has `c-1` anchored to the first trace; MAF's registry has never heard of it, anchors `c-1` to the second, and you get two traces for one conversation. That is the same bug #2714 fixes for Haystack. The stand-down cases here cover nesting, which is tested and correct, but not interleaving.

I think `ConversationTraceRegistry.claim` and `target` could read and write `_anchors` through those two accessors, with hex to int conversion being most of the work, and then all five integrations would agree on one key. If that is more than you want in this PR, I am happy with an issue and a line in the docs saying the join is per-integration and does not coordinate with `conversation_turn`.

Worth saying that the output shape is consistent either way. Both mechanisms end up with several `is_turn_root` spans on one trace id, and the exporter keys `turn_root_conversations` by trace id, so the backend sees the same thing. It is the keying that diverges, not the result.

## 3. `create_agent` counts as a run root

`_AGENT_SPAN_PREFIXES` in `mapping.py` includes both `invoke_agent ` and `create_agent `, while the docstring on `is_agent_invocation_span` says `invoke_agent <name>`. I confirmed `is_agent_invocation_span("create_agent assistant")` returns `True`.

This is latent rather than live: the MAF version we pin defines `AGENT_CREATE_OPERATION` but never opens a span with it, so nothing emits one today. But if that changes, a root agent-creation span gets stamped as a conversation turn root carrying whatever input and output that trace collected, which would show up as a phantom turn. I suspect the pair came across from the name mapping, where both operations legitimately map to `ai.agent.invoke`. Either drop `create_agent` or keep it and say in the docstring why creating an agent counts as running one.

## 4. The bare chat span gap is worse than the description says

The description covers this as a memory issue: a trace whose only MAF span is a bare `chat` has no run root to release from, so its entries wait for the entry cap. That part is right.

The part I would add is that `claim` fires on every MAF span, including `chat`, so a chat-client-only app does get its turns joined onto one trace. And `conversation_root_attributes` returns `None` for a `chat` span, so that joined trace carries no `is_turn_root` and no `conversation.id` anywhere on it. That is the same outcome you describe as "worse than leaving them apart" when justifying the root `invoke_agent` stamp, so I think the two cases should be settled the same way: either stamp a root `chat` span too, or do not claim from one.

## 5. Claiming from every MAF span rather than the root

The comment justifies the breadth by saying the claim has to be recorded before any span of the trace exports. That is true, but the root starts first, so `if span.parent is None` satisfies it just as well, and it would sit next to `_maf_rooted_traces.mark` above it, which already only fires there.

Two things come out of claiming per span. `claim` writes a self-entry, `targets[T1] = T1`, for the anchor trace on the first later span, which is harmless for `target()` but takes a slot and triggers an eviction pass. More interesting, if an app changes the conversation id mid-run, a nested span anchors the new conversation onto the current trace, quietly merging two conversations. Narrow, but claiming only from the root removes it for free.

## On the testing

Worth saying separately: the mutation table is genuinely useful and made this much faster to review, and the note about the `session_provider` fixture riding a provider that already carries a `BatchSpanProcessor` saved me working out for myself why the popping bug was observable here and not in ADK. Please keep doing both.

The one thing I would take from point 1 is that a mutation surviving is information too. The two tests that caught my `contains` mutation both assert on queue internals, and the one that asserts on the exported span did not notice. That is the signal that the guard is not doing work, and a mutation table that only records catches will not show it.

harry-rhesis:
Thanks — this was a genuinely useful review, and the branch is better for it. Taking your points in order.

**First, the branch state.** Rebased onto current main, so the merge commit is gone and no force-push is needed from you. Worth flagging why my own rebase failed the first time, in case the same thing bites you: the shared clone had been made shallow, which truncated local `origin/main` to 7 commits with a bogus root, so there was no merge base and the rebase tried to replay 4847 commits. `git fetch --unshallow` fixed it.

## 1. `_MafRootedTraces` — you were right, it is gone

I ran your mutation before deciding, and got what you got: `test_the_root_still_reads_after_a_nested_span_exported` passes without the guard, and the only failures assert on queue internals. Your reasoning holds — `queue_release` frees nothing on the way in, and a read moves a trace back to the newest, so the root's own read refreshes the entry whatever a nested span did earlier.

Removed the class, the module global, the `on_start` branch and the four tests. MAF now queues unconditionally from a nested run root, matching ADK. The leak I was closing stays closed, because unconditional queueing covers that case too.

Your meta-point is the part I'd have missed on my own, and I've put it in the commit message: a surviving mutation is information. `never-mark-the-root` survived every test, and I responded by writing a test to catch the mark rather than asking whether the guard did any work. A table that only records catches hides exactly that.

## 2. Two stores — real, and now fixed in #2784

Reachable, not theoretical, and you are right that #2711 exposed those accessors to prevent it. I did not want to change the shared `ConversationTraceRegistry` from inside this PR, because it would alter ADK's behaviour from a PR that does not mention ADK, so it is #2784 instead: the registry reads and writes `_anchors` through the two accessors, and all five integrations now agree on one anchor per conversation id.

One detail from doing it that is worth recording: the anchor has to be written *before* it is read back, so the store's first-writer-wins is the single decision about who anchors. A get-then-set lets two concurrent turns both believe they are the anchor.

I have also added a callout to the MAF docs page saying the join is per-integration and does not coordinate with `conversation_turn`, since you were right that the page implied the guarantee was uniform.

## 3. `create_agent` — dropped

Confirmed `is_agent_invocation_span("create_agent assistant")` returned `True`, and that the pinned MAF defines `AGENT_CREATE_OPERATION` but never opens a span with it. Your guess about where the pair came from looks right: both operations legitimately map to `ai.agent.invoke` in the name table, and the prefix tuple picked up both. Building an agent is not running one, so it is out rather than documented.

## 4 and 5 — settled together

Claiming now happens only from the span that roots the trace, and only when that span is a run root. That resolves both:

- A bare `chat` span no longer claims. You are right that joining those turns onto a trace carrying no `conversation.id` anywhere is the same outcome I used to justify stamping a root `invoke_agent`, so it should be settled the same way. I chose not to claim rather than to stamp, because a root chat span is a chat-client call rather than a served turn.
- The root starts first, so restricting to it loses nothing, and it takes the mid-run conversation-id change and the `targets[T1] = T1` self-entry with it.

Three new tests, one per behaviour, and each is caught by the corresponding mutation.

The one remaining gap is still the bare chat span's memory: no run root means nothing releases, so those entries wait for the entry cap. It is recorded in the commit message and the PR body.

**Current state:** 8 commits, 3187 passing, all signed off, 0 behind main.

PR #2714, opened 2026-09-11
Join Haystack turns through the shared conversation anchor

harry-rhesis:
## Purpose

`haystack/conversation.py` kept its own copy of the synthetic-parent trick and remembered a conversation's first-turn trace id in a private instance field. So a Haystack app that also opened a `conversation_turn` for the same conversation id, for a turn Haystack does not serve, got **two traces for one conversation**: each mechanism consulted only its own record and neither knew the other had already anchored that id.

PR #2711 exposed `get_conversation_anchor` and `anchor_conversation` for exactly this, and the LangGraph integration already uses them. This makes Haystack the last integration with a private copy.

## What Changed

- `_conversation_parent_context` is gone, replaced by the shared `build_conversation_parent_context`. Alongside deleting the duplicate, this picks up a check the private copy lacked: a rejection of trace id zero, which OTEL will not accept.
- A named conversation now resolves its anchor through the shared store, so a turn opened here lands on the same trace as one `conversation_turn` or `@endpoint` opened for the same id, in either order.
- The instance field stays for turns opened *without* a conversation id, where there is no key to share. That path is unchanged and now has a test.

**One deliberate behaviour change.** Calling `start_conversation("c-1")` twice used to start a fresh trace. It now rejoins `c-1`'s trace, because the shared store is first-writer-wins: whichever turn ran first keeps the id anything else in the process already observed and published. That is the rule `conversation_turn` follows, so I took matching it as the point rather than something to work around. The docstring and `haystack.mdx` both said the old thing and are corrected.

## Additional Context

- Uses the two accessors #2711 added to `packages/rhesis/src/rhesis/telemetry/conversation.py`. That PR has since merged, so this branch is rebased onto `main` and stands on its own.
- Companion PR #2713 fixes the equivalent conversation-joining gap in the MAF integration. It is independent of both this and #2711.

## Testing

`204 passed` in the Haystack suite, `813 passed` across `tests/sdk/telemetry`. Ruff check and format clean.

The two new cross-mechanism tests were mutation-tested: making `_parent_context` ignore the shared store and consult only the instance field makes both fail, while the negative control (`test_a_different_conversation_is_not_joined`) still passes. So they are testing the join, not just exercising it.

Also added a `reset_conversation_anchors` fixture. The anchor store is process-wide, so without it two tests reusing a conversation id would land on one trace and `test_later_turns_hang_off_the_synthetic_parent` would fail depending on collection order.

To reproduce the bug this fixes, before this change: open a `conversation_turn("c-1")`, then run a `RhesisTracing` turn under `start_conversation("c-1")`, and count distinct trace ids. Two before, one after.

### DIFF
--- a/docs/content/docs/tracing/haystack.mdx
+++ b/docs/content/docs/tracing/haystack.mdx
@@ -146,7 +146,7 @@ tracing.flush()`}
 
 Assign `turn.output` yourself. Only your application knows which part of a pipeline result is the reply the user saw — it may be a tool result or a value held in agent state rather than the last assistant message.
 
-Every turn after the first joins the first turn's trace, so the conversation reads as one trace. Calling `start_conversation` again begins a new one. `RhesisTracing` never raises: if Rhesis is not configured it logs a warning and every method becomes a no-op, so tracing cannot take your application down. Check `tracing.enabled` to report it.
+Every turn after the first joins the first turn's trace, so the conversation reads as one trace. Calling `start_conversation` with a new id begins a new one; passing an id this process already recorded a turn for rejoins that conversation, whether the earlier turn came from here or from [`conversation_turn`](/docs/tracing/conversation-tracing#standalone-apps-owning-the-turn). `RhesisTracing` never raises: if Rhesis is not configured it logs a warning and every method becomes a no-op, so tracing cannot take your application down. Check `tracing.enabled` to report it.
 
 ## Trace Links
 
--- a/sdk/src/rhesis/sdk/telemetry/integrations/haystack/conversation.py
+++ b/sdk/src/rhesis/sdk/telemetry/integrations/haystack/conversation.py
@@ -17,11 +17,15 @@ from collections.abc import Iterator
 from contextlib import contextmanager
 from typing import TYPE_CHECKING, Any, Optional
 
-from opentelemetry import trace
-from opentelemetry.trace import NonRecordingSpan, Span, SpanContext, TraceFlags
+from opentelemetry.trace import Span
 
 from rhesis.telemetry.constants import ConversationContext
 from rhesis.telemetry.context import get_root_trace_id, set_root_trace_id
+from rhesis.telemetry.conversation import (
+    anchor_conversation,
+    build_conversation_parent_context,
+    get_conversation_anchor,
+)
 
 if TYPE_CHECKING:
     from rhesis.sdk.telemetry.integrations.haystack.tracer import RhesisTracer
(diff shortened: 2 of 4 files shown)


Arman-Beykmohammadi:
Reviewed this one alongside #2713 and #2715, since the three overlap.

**Mergeable once the zero trace id guard goes in.** Everything else here I like. Deleting the private copy of the synthetic parent trick is a real simplification, the shared store is the right place for this to live, and making `start_conversation("c-1")` twice rejoin rather than split is the correct call: first-writer-wins is the rule `conversation_turn` already follows, so matching it beats working around it. The two cross-mechanism tests are the right tests, and the `reset_conversation_anchors` fixture is worth having on its own.

I ran the suite on the branch: 204 passed. It still applies cleanly to main.

**The one thing I would fix before merging.**

peqy's comment on line 274 is still open. Its timestamp is about five minutes after your last commit, so I think it just missed the cut rather than being considered and dismissed. I reproduced it, and I think it is worth doing:

```
is_valid: False  trace_id: 00000000000000000000000000000000
shared anchor  -> 00000000000000000000000000000000
parent context -> None
after a good later write -> 00000000000000000000000000000000   (first-writer-wins)
```

With a disabled tracer config, `Tracer.start_span` returns a `NonRecordingSpan` whose context is all zeros. `turn()` formats that and hands it straight to `anchor_conversation` with no validity check. `build_conversation_parent_context` then rejects zero for every later turn, so the conversation scatters for the life of the process, and because `_anchors.set` uses `setdefault` a healthy later turn cannot correct it.

What makes this worth fixing now rather than later is that moving to the shared store changed the blast radius. Before this PR a poisoned anchor was confined to one `RhesisTracing` instance. After it, the bad value sits in the process-wide store, so it takes `conversation_turn` and `@endpoint` down for that conversation id too. `conversation_turn` guards against exactly this at `conversation.py:239`, with a comment explaining why. Haystack should do the same, before both the `anchor_conversation` call and the `_unnamed_anchor` assignment. One line.

**Two smaller notes, neither blocking.**

`turn()` sets `root_trace_id` but not `conversation_id`, while `conversation_turn` binds both. Nothing breaks today because MAF and ADK stand down on either var, but an integration running inside a Haystack turn cannot tell which conversation it is in, so its own content registry records no session id. Worth binding both for symmetry.

The PR description says this makes Haystack the last integration with a private copy. That is true within this mechanism, but #2713 adds a second consumer to `ConversationTraceRegistry` in `genai.py`, which is a separate store keyed the same way and never consulted from here. So a `conversation_turn("c-1")` followed by a bare MAF `agent.run()` under `set_conversation_id("c-1")` still produces two traces for one conversation, which is the same bug this PR fixes for Haystack. Not something to solve in this PR, and I have written it up properly on #2713.

harry-rhesis:
Thanks — the zero trace id is a real bug and it is fixed. You were also right about why it was worth fixing now rather than later: this PR is what widened the blast radius, and I had not thought about that when I moved the anchor onto the shared store.

**The guard is in.** `turn()` now checks `span_context.is_valid` before it anchors anything, the same check `conversation_turn` makes and for the same reason. Restructured slightly so the guard reads once and covers both the named and the unnamed path, rather than repeating the condition.

Worth recording that the unnamed path turned out to need no separate guard: assigning `None` to `_unnamed_anchor` leaves it `None`, so that mutation is behaviourally equivalent and no test can catch it. That is why it is one `if trace_id:` around both branches instead of two guarded assignments — the structure makes the invalid case a single skip.

Three tests: a named conversation is left unanchored, an unnamed turn is left unanchored, and — the one that matters for your first-writer-wins point — a healthy later turn can still anchor the conversation afterwards.

**Conversation id is bound too.** You are right that this was asymmetric with `conversation_turn`, and your reasoning about the consequence is the concrete version of it: an integration nested in a Haystack turn could not tell which conversation it was in, so its own content registry recorded no session id. Bound for the length of the turn and restored after, with two tests.

**The description claim.** Corrected. You are right that "the last integration with a private copy" was only true within this mechanism, and #2713 was adding a second consumer to the other store at the same time. That is now #2784, which makes `ConversationTraceRegistry` read and write the shared anchors, so the `conversation_turn("c-1")` followed by a bare `agent.run()` case you describe is fixed rather than just written down.

**Current state:** 2 commits, 209 passing, all signed off, 0 behind main.

PR #2715, opened 2026-09-11
Stop one Google ADK exporter starving the others

harry-rhesis:
## Purpose

The same flaw #2713 fixes in the MAF integration, in Google ADK. I flagged it there as "probably present, deserves its own change"; it is present.

`conversation_root_attributes` popped the per-trace conversation content as it stamped an ADK run root. `enable()` wraps *every* exporter on the provider, so a process running its own OTLP collector beside Rhesis has two translating exporters exporting the same spans: whichever reached the trace root first took the content, and the other stamped a root span carrying no conversation at all. Which one won was arbitrary.

It costs more here than in MAF. ADK's run root is the *only* place that content can go, because the v1 `invocation` span has no attributes of its own and the user's query and the final answer live on nested `call_llm` spans. An exporter that loses the race loses the conversation for that trace entirely, rather than just some of its fields.

## What Changed

The content is read rather than popped, and released a bounded number of reads later, matching the MAF registry. Deferring keeps entries alive long enough for the other exporters to see the same trace; the bound keeps a long-running process from holding 10 KB of conversation text per trace until the 4096-entry cap evicts it. The queue is larger than an OTEL export batch, since a busy service can flush hundreds of run roots at once and the exporters do not drain their queues in lockstep.

**A read of a trace with nothing recorded takes no slot in the release queue.** I found this by writing a test that was wrong, and reading why it failed rather than adjusting the assertion. #2713 now carries the same guard, so the two registries behave identically apart from the stores each one owns. Without the guard, a run of content-free traces (single-turn runs, or content capture switched off) pushes the entries of traces that *do* have content out of the queue early. ADK has four stores rather than MAF's three, and all four are checked, the ADK session id included, so the eviction path cannot quietly miss one and grow on its own.

## Additional Context

- Independent of #2713 and #2714. It branches off `main` and touches no file either of them touches. Merge order does not matter.
- Uses the shared `DeferredReleaseQueue` from #2718, which has merged, so this branch sits on `main` directly and stands alone.
- The last commit swaps this registry's private copy of the queue for the shared class, so ADK and MAF no longer carry two implementations that can drift.
- This is the last of the three integrations that kept conversation state this way. Haystack (#2714) now shares the anchor store; MAF (#2713) and ADK read without popping.

## Testing

`3019 passed` in `tests/sdk` (excluding `tests/sdk/integration`, which needs Docker). Ruff check and format clean. The ADK suite grew from 128 to 137 tests.

Every behavioural change is mutation-tested, and each mutation is caught by the test written for it:

| Mutation | Caught by |
|---|---|
| Make the read pop again | `test_every_wrapped_exporter_stamps_the_conversation`, plus 3 registry tests |
| Never release a served entry | `test_a_read_queues_the_release`, `test_the_adk_session_id_is_released_too` |
| Let an empty read take a slot | `test_a_trace_with_nothing_recorded_does_not_take_a_slot` |
| Stop releasing the ADK session store | `test_the_adk_session_id_is_released_too` |
| Drop any one store from the guard | `test_content_in_any_one_store_counts_as_content`, one case per store |
| Omit any one store when constructing the shared queue | same test, re-run after the refactor |

`test_every_wrapped_exporter_stamps_the_conversation` runs a real ADK agent, captures ADK's own spans through a processor attached *after* `enable()` so it stays unwrapped, then replays that batch through two freshly built translating exporters. With the wrapped exporter from the fixture that makes three readers of the same trace, all of which must stamp. Note that OTEL cannot remove a span processor, so the raw one stays attached for the session; nothing else reads its exporter.

The existing ADK suite did not catch this because its `session_provider` attaches a single processor, so the race never occurred. The MAF suite happens to run with two and that is where the bug first showed itself.

## Review feedback applied

peqy's one comment was valid and is fixed in `test(sdk): stop the adk replay exporter accumulating spans`. The raw span processor the two-exporter test attaches cannot be removed (OTEL has no API for it), so its exporter kept collecting every span in the session into an unbounded list nothing reads. `raw.shutdown()` makes its export a no-op, which stops the growth quietly since `SimpleSpanProcessor` ignores the returned `FAILURE`.

One nuance worth recording, since the comment did not mention it: this isolates the memory, not the test. The leftover processor is still found and wrapped by every later `enable()`, so subsequent tests keep running with two translating exporters over the same spans. That is harmless, and is exactly the configuration this PR makes safe.

### DIFF
--- a/sdk/src/rhesis/sdk/telemetry/integrations/google_adk/translator.py
+++ b/sdk/src/rhesis/sdk/telemetry/integrations/google_adk/translator.py
@@ -55,6 +55,7 @@ from rhesis.sdk.telemetry.integrations.genai import (
     TRUTHY_ENV_VALUES,
     AncestryRegistry,
     ConversationTraceRegistry,
+    DeferredReleaseQueue,
     RetracedSpan,
     TranslatedSpan,
     content_capture_enabled,
@@ -324,10 +325,16 @@ class _ConversationContentRegistry:
         self._adk_session_by_trace: dict[int, tuple[int, str]] = {}
         self._rhesis_conversation_by_trace: dict[int, str] = {}
         self._max_entries = max_entries
-        # Traces whose content has been handed to an exporter, oldest first.
-        self._served: dict[int, None] = {}
-        self._max_served = max_served
         self._lock = threading.Lock()
+        # All four stores go in: one left out never gets freed. Guarded by the
+        # lock above, which the queue expects the caller to hold.
+        self._release = DeferredReleaseQueue(
+            self._rhesis_conversation_by_trace,
+            self._adk_session_by_trace,
+            self._input_by_trace,
+            self._output_by_trace,
+            max_served=max_served,
+        )
 
     def _evict_if_needed(self, store: dict) -> None:
         if len(store) < self._max_entries:
@@ -418,51 +425,12 @@ class _ConversationContentRegistry:
             adk_session = self._adk_session_by_trace.get(trace_id)
             conv_input = self._input_by_trace.get(trace_id)
             conv_output = self._output_by_trace.get(trace_id)
-            self._queue_release(trace_id)
+            self._release.queue_release(trace_id)
         conversation_id = rhesis_id or (adk_session[1] if adk_session else None)
         return (
             conversation_id,
             conv_input[1] if conv_input else None,
             conv_output[1] if conv_output else None,
-        )
-
-    def _queue_release(self, trace_id: int) -> None:
-        """Drop the entries of whatever has fallen far enough behind.
-
-        Deferred rather than immediate so every exporter reading the same batch
-        sees the same content, and bounded so a long-running process does not
-        hold 10 KB of conversation text per trace until the entry cap evicts it.
-        The queue is deliberately larger than an OTEL export batch, since a busy
-        service can flush hundreds of run roots at once and the exporters do not
-        drain their queues in lockstep.
-
-        Caller holds the lock.
-        """
-        if not self._has_entries(trace_id):
-            # Nothing to release, so it must not take a slot: a run of traces
-            # with no conversation content on them would otherwise push out the
-            # entries of the ones that have it.
-            return
-        self._served.pop(trace_id, None)
-        self._served[trace_id] = None
-        while len(self._served) > self._max_served:
-            stale = next(iter(self._served))
-            del self._served[stale]
-            self._rhesis_conversation_by_trace.pop(stale, None)
-            self._adk_session_by_trace.pop(stale, None)
-            self._input_by_trace.pop(stale, None)
-            self._output_by_trace.pop(stale, None)
-
-    def _has_entries(self, trace_id: int) -> bool:
-        """Whether any of the four stores holds something for this trace."""
-        return any(
-            trace_id in store
-            for store in (
-                self._rhesis_conversation_by_trace,
-                self._adk_session_by_trace,
-                self._input_by_trace,
-                self._output_by_trace,
-            )
         )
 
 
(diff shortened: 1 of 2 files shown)


Arman-Beykmohammadi:
Reviewed this one alongside #2713 and #2714, since the three overlap.

**This is mergeable as-is.** It is the smallest and the clearest of the three. The pop-to-read change is the right fix, the reasoning about why an empty read must not take a slot is sound, and passing all four stores into the queue is load-bearing rather than defensive: leaving one out is exactly the leak the queue exists to stop. Good call finding the empty-read case by reading why a wrong test failed instead of adjusting the assertion. peqy's point about the replay exporter accumulating spans is applied.

I checked out the branch and ran the suite: 137 passed. The branch is 55 commits behind main, but nothing on main has touched `sdk/src/rhesis/sdk/telemetry/` or `packages/rhesis/src/rhesis/telemetry/` since the merge base, so it still applies cleanly and merge order does not matter.

Two things I want to record, neither of which blocks the merge.

**1. ADK stamps by one key and joins by another.**

`read_for_root` resolves the conversation id as `rhesis_id or adk_session`, so an app that only uses ADK sessions still gets `conversation.id` stamped on its run root. But `claim()` at `on_start` only ever sees `get_conversation_id()`, because the ADK session id lives in span attributes and is not readable until `on_end`. So a session-only ADK app gets its turns grouped in the UI while each turn keeps its own trace.

That is the exact split #2713 treats as a bug worth fixing in MAF. Here it is a structural limit rather than an oversight, and I do not think it is fixable without a different hook. What I would like is for it to be written down somewhere, because the MAF docs page #2713 adds reads as though the guarantee is uniform across integrations, and it is not.

**2. Nested run roots.**

`conversation_root_attributes` calls `read_for_root` before the `parent is not None` check, so a nested `AgentTool` `invocation` queues a release for a trace whose real root has not read yet. That is fine, and I think it is the right choice: `queue_release` does not free anything, and a re-read moves the trace back to the front of the queue, so the root's own read refreshes it.

I am flagging it only because #2713 makes the opposite choice for the same case and builds a registry to avoid it. One of the two should change so the registries actually behave identically, which is what both PR descriptions claim. My preference is that MAF adopts what you did here. Details in my comment on #2713.

harry-rhesis:
Thanks. Nothing to change here, and I want to be explicit about why for each point rather than just agreeing.

**1. Stamping by one key and joining by another.** Your analysis is right, and it is already written down — `docs/content/docs/tracing/google-adk.mdx` has a callout under Conversation Grouping headed "Turn labelling and turn joining need different things", which says a session-only app is labelled and grouped in the Conversation tab but not joined into one trace, and why: the target trace has to be known when ADK creates the run span, and ADK assigns no attributes that early. It is pre-existing on main rather than something this PR adds, so I have not duplicated it.

The real gap your point identified was on the other side: the MAF page #2713 adds read as though the guarantee was uniform. I have added a callout there saying the join is per-integration and pointing readers at each integration's own page, rather than restating the differences in two places.

**2. Nested run roots.** Agreed, and MAF now matches this rather than the other way round. I ran your `contains` mutation on #2713 and got your result: the behavioural test passes without the guard, and only the queue-internals assertions fail. So `_MafRootedTraces` is gone and MAF queues unconditionally from a nested run root, exactly as ADK does here. Details in my reply on #2713.

**On the cross-PR store concern** you raised on #2713: that is now #2784, which makes `ConversationTraceRegistry` anchor through the shared store. It changes this integration's behaviour too — an ADK turn will join a trace `conversation_turn` anchored — so it is worth a look from you even though it is not this PR.

One thing #2784 turned up that is relevant here: the existing `TestConversationTraceRegistry` tests in `test_google_adk.py` passed only because they happened to reuse conversation ids consistently. Once the anchors are process-wide their result depends on collection order, so that class now has an isolation fixture.

**Current state:** 4 commits, 137 passing, all signed off, 0 behind main. Unchanged from what you reviewed apart from the rebase.

PR #2719, opened 2026-09-11
feat: configurable endpoint timeout_seconds

harry-rhesis:
## Summary

- Adds a configurable `timeout_seconds` setting to endpoints, stored in `endpoint_metadata` JSON (no DB migration)
- Exposed as a top-level API field with 1-3600s validation, respected by both REST (httpx per-request override) and SDK (RPC/WebSocket timeout) invokers
- Frontend: timeout field on both create and edit flows, with system-default fallback display
- SDK: `timeout_seconds` field on the `Endpoint` entity, included in push/pull round-trips

Fixes the issue where the hardcoded 30s timeout caused confusing failures during endpoint setup for slower endpoints.

## Test plan

- [x] 12 schema validation tests (bounds, edge cases, metadata extraction)
- [x] 4 CRUD merge helper tests (empty metadata, key preservation, overwrite, None no-op)
- [x] 2 REST invoker tests (custom timeout, default fallback)
- [x] 3 SDK invoker tests (RPC, WebSocket, default fallback)
- [x] 8 SDK entity tests (defaults, serialization, push/pull)
- [x] Ruff and TypeScript clean

### DIFF
--- a/apps/backend/src/rhesis/backend/app/crud/endpoint.py
+++ b/apps/backend/src/rhesis/backend/app/crud/endpoint.py
@@ -66,11 +66,29 @@ def get_endpoints(
     )
 
 
+def _merge_timeout_into_metadata(
+    db_endpoint: models.Endpoint,
+    timeout_seconds: Optional[int],
+) -> None:
+    """Merge timeout_seconds into endpoint_metadata without overwriting other keys."""
+    if timeout_seconds is None:
+        return
+    meta = dict(db_endpoint.endpoint_metadata or {})
+    meta["timeout_seconds"] = timeout_seconds
+    db_endpoint.endpoint_metadata = meta
+
+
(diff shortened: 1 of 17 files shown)


harry-rhesis:
Both fixed in 8a2e555:

1. **update_endpoint clear case** -- now uses `model_fields_set` to distinguish "not sent" vs "sent as null". Sending `timeout_seconds: null` pops the key from `endpoint_metadata`, returning to system default. Renamed helper to `_set_metadata_timeout` to reflect set/clear semantics.

2. **Response validator ORM mutation** -- the `_extract_timeout_from_metadata` validator now returns early for non-dict inputs. ORM objects expose `timeout_seconds` via the read-only `@property`, which Pydantic's `from_attributes` picks up during field validation.

PR #2739, opened 2026-09-16
fix(frontend): persist playground test parameters in localStorage

harry-rhesis:
## Summary

- Persist test parameters per endpoint in `localStorage` so they survive page reloads and navigation
- Auto-restore saved parameters and re-open the panel when revisiting an endpoint
- Updated label from generic description to actionable: "Sent with every message. Use `{{ test_parameters.<key> }}` in the endpoint's request mapping."
- Clear from storage when the editor is emptied

## Context

The test parameters panel worked end-to-end (frontend sends them via WebSocket, backend extracts them and passes them through the template renderer), but the UX had three issues:
1. Parameters were lost on reload or endpoint switch (no persistence)
2. No indication whether parameters apply per-message or per-session
3. Unclear that the endpoint's request mapping must reference them via Jinja syntax

## Test plan

- [ ] Open playground, set test parameters JSON, send a message, reload the page. Parameters should be restored and panel open.
- [ ] Switch to a different endpoint. Parameters from the first endpoint should not carry over.
- [ ] Switch back. Original endpoint's parameters should be restored.
- [ ] Clear the JSON editor. Parameters should be removed from localStorage.
- [ ] Verify the updated label text reads clearly.


harry-rhesis:
Addressed peqy's review feedback: added an else branch in `handleTestParamsChange` that clears `testParamsParsed` to `null` and removes the localStorage entry when the parsed JSON is valid but not a plain object (e.g. arrays, strings, numbers). This prevents stale parameters from being sent with messages while the editor shows non-object content.

Also applied Prettier formatting to 32 pre-existing files on the release branch.

PR #2751, opened 2026-09-17
Add organization branding settings

### LINKED ISSUE #2578: Per-organization UI branding configuration
## Summary

Let each organization configure its own branding through the UI settings page, stored in the database, instead of relying solely on deployment-wide environment variables.

## Problem / Opportunity

PR #2576 introduced white-label theming via `BRAND_*` env vars (`BRAND_PRIMARY_COLOR`, `BRAND_SECONDARY_COLOR`, `BRAND_FAVICON_URL`, `BRAND_PRODUCT_NAME`, `BRAND_FONT_FAMILY`, `BRAND_FONT_BASE_URL`). This works well for single-tenant deployments (one brand per instance), but multi-tenant deployments and managed hosting need per-org branding so each customer sees their own identity without requiring separate infrastructure.

## Users & Use Cases

- Platform operators hosting multiple organizations on a single deployment
- Organization admins who want to apply their own brand without touching infrastructure config

## Proposal

Build on the branding infrastructure from #2576 rather than replacing it. The env vars become the deployment-wide default; org-level settings override them.

### Backend

- Add branding fields to the organization model (primary/secondary color, favicon, product name, font family, font source). A JSON column or dedicated columns on the organization table.
- Expose branding through the existing org settings API (`GET/PATCH /organizations/{id}`).
- Favicon and font file uploads go through object storage (`STORAGE_SERVICE_URI`), not env vars or external URLs.

### Frontend

- Add a branding section to the organization settings page with fields for: primary color (hex picker), secondary color (hex picker), favicon (file upload), product name (text), font family (text), and font files (file upload for 300/400/700 weights).
- The existing branding pipeline in `src/config/branding.ts` (`getServerBranding()`) resolves values in order: org record > env var > Rhesis default. The `Branding` and `BrandFont` interfaces stay the same.
- `getDesignTokens()` in `src/styles/theme.ts` already accepts `BrandColors` with optional `primaryColor`, `secondaryColor`, and `fontFamily`. No changes needed to the theme layer.
- Font weight remapping (`w500->400`, `w600->700`, `w800->700`) already handles brand fonts that only ship 300/400/700, so org-uploaded fonts get the same treatment.
- The `/brand-fonts/*` Next.js rewrite proxy (for CORS) would need to route per-org to the storage backend instead of a single `BRAND_FONT_BASE_URL`.

### Non-Goals / Out of Scope

- Per-user or per-project branding
- Live preview editor for branding changes
- Changing chart palettes, sign-in artwork, or documentation/blog links per org (these stay Rhesis-branded as noted in the env vars docs)

## Acceptance Criteria

- [ ] Organization settings page has a branding section with fields for primary color, secondary color, favicon, product name, font family, and font files
- [ ] Saved branding persists in the database and applies on next page load for all org members
- [ ] Env var values serve as deployment-wide defaults when org-level branding is not configured
- [ ] Clearing org branding reverts to env var / Rhesis defaults
- [ ] Font weight remapping works the same for org-uploaded fonts as for env var fonts
- [ ] Uploaded favicons and font files are served from object storage, not external URLs

harry-rhesis:
## Summary

- **Backend**: new `organization_settings` JSONB column on `organization` with a branding section. PATCH endpoints for colours, product name, favicon upload and brand font (Google Fonts or uploaded files). Three-tier resolution: org setting > deployment `BRAND_*` env var > Rhesis default.
- **Frontend**: Branding tab under Organization Settings. Colour picker changes preview in the MUI theme instantly via `BrandPreviewContext` (no page reload), with debounced API saves. Google Font dropdown renders each option in its actual typeface and applies the font on selection. Favicon upload with dimension validation and preview.
- **Deployment**: `BRAND_*` env vars in helm configmap, `values.yaml` defaults and docker-compose. Docs page for operators and end users.

Closes #2578

## Test plan

- [ ] Backend: `pytest tests/backend/models/test_organization_settings.py tests/backend/routes/test_organization_settings.py tests/backend/schemas/test_organization_settings.py tests/backend/services/test_organization_branding.py tests/backend/utils/test_uploads.py`
- [ ] Frontend: `npx jest --testPathPatterns 'BrandingColorsForm|BrandMark|branding\.test|useDocumentTitle|brand-fonts'`
- [ ] Manual: open Organization Settings > Branding, pick a primary colour from the swatch and confirm the app bar changes in real time
- [ ] Manual: select a Google Font and confirm the page typography changes on selection (no button click)
- [ ] Manual: upload a favicon and confirm the browser tab icon updates
- [ ] Manual: click Reset on each section and confirm values revert to deployment defaults
- [ ] Verify the Alembic migration applies cleanly: `alembic upgrade head`


harry-rhesis:
Thanks for the review. Assessed each point:

**1. Blocking storage I/O on the event loop** — Valid. `store_bytes()` calls the sync `put_object_bytes()` (`fs.open().write()`), and `delete_object`/`delete_file` call `fs.rm()` synchronously despite being declared `async`. These block the event loop when called from the async branding endpoints. This is a pre-existing pattern in `StorageService` (see `save_file`, `delete_file`), not introduced by this PR. Will address across all StorageService callers in a follow-up.

**2. OffLoopSession may hop threads** — Not an issue. `_read_branding` and `_apply_branding` each wrap the entire DB interaction in a single `anyio.to_thread.run_sync` call and the session is never used concurrently from two threads. Sequential `run_sync` calls may land on different threads, but SQLAlchemy sessions are not thread-affine outside a transaction, and `_apply_branding` commits before returning.

**3. ETag when hash is missing** — Not applicable. `sha256` is computed at upload time by `store_bytes()`, so every branding asset row always has a hash. There are no older rows to worry about since branding assets are a new feature in this PR.

harry-rhesis:
Good points. Two things happening:

**1. Blocking storage I/O**: PR #2753 landed the `StorageService` fix (offloads `save_file`, `get_file`, `delete_file` to the threadpool, adds `put_object_bytes_async`). Will rebase this branch onto it and switch the branding callers to the async variants.

**2. OffLoopSession thread-hop**: The pattern here is strictly sequential, not concurrent, and `_apply_branding` commits before returning, so there is no transaction carried across thread boundaries. SQLAlchemy's thread-safety warning is about concurrent access. That said, consolidating read+apply into one `run_sync` call per endpoint is cleaner and removes the question entirely. Will do that in this rebase.

harry-rhesis:
Both addressed in cb3b17ba5:

1. **Blocking storage I/O**: `upload_favicon` and `upload_font` now use `store_bytes_async`, which wraps the sync `put_object_bytes` in `anyio.to_thread.run_sync`. (`delete_object`/`delete_file` are already fixed in #2753.)

2. **Session thread-hop**: Replaced the separate `_read_branding` + `_apply_branding` pair with `_read_and_apply_branding`, which reads the previous state and applies the update in a single `run_sync` call. The Session never leaves the thread it was used in. All four branding endpoints updated.

PR #2756, opened 2026-09-17
Keep the usage card current while a run executes

Arman-Beykmohammadi:
## Purpose

The Usage card on a test run's summary read its token and cost totals once, in a bare `useEffect`, and then held whatever it saw. Open a run while it is still executing and the card sat at the first numbers it got while every other figure on the page moved.

## What Changed

It is a react-query query now, polling on the same three seconds `useLiveTestRun` uses, with its own key in `constants/query-keys.ts`.

When it stops is the part worth explaining. Terminal status is the wrong signal, even though it is the obvious one: enrichment prices a run asynchronously *after* it ends, so stopping when the run goes terminal stops polling exactly when cost is about to arrive.

So it stops when the figures settle instead. Cost present means priced; no tokens at all means there is nothing to price. Failing either, it gives up two minutes after the run finished, so a run whose models have no published price does not poll for as long as the tab stays open.

That window is measured from the run's `completed_at`, not from when the card mounted. Opening a run that finished last week therefore costs one request rather than two minutes of them, while opening one that finished five seconds ago still waits for its price.

The decision is a plain exported function that takes the current totals, whether the run is going, when it finished, and the time. Every branch is tested directly, which is a great deal more useful than trying to test react-query's scheduler.

`KpiRow` passes `isRunning` down rather than the hook re-deriving it; `RunSummary` already had it.

## Additional Context

Cost now shows a dash instead of `$0.00` until it is known. Tokens come off the spans and are there as soon as the calls are traced, while cost stays a real zero until enrichment runs, so `$0.00` was asserting the run had been free. Of the readings available for that zero, that is the only one that is certainly wrong.

**Deliberately out of scope.** The card keeps its current shape: tokens as the headline, cost beside them, "across N traces" underneath. Recutting it around cost, with the model in the footnote, is a separate change. Nothing else on the summary starts polling.

## Testing

`npm run format`, `npx tsc --noEmit`, `npm run lint` and jest all clean. Eleven new tests, plus two on `KpiRow`.

The poll logic was checked against both ways of getting it wrong. Replacing it with the obvious "stop when the run is terminal" turns the keeps-polling test red. Removing the two-minute cap turns three tests red, including the one that opens a year-old unpriced run and expects no polling at all.


Arman-Beykmohammadi:
Both points addressed, and the first one turned out to go deeper than the hook.

**On the zero-cost question: yes, it can mean priced.** `enriched_cost_expr(..., default=0.0)` bottoms out at `0.0` for three different situations — a scope enrichment has not reached, one whose models LiteLLM cannot price, and one that genuinely cost nothing. Nothing in the response could tell them apart, so the hook was guessing, and both symptoms you predicted were real: it polled through the full window for a free run and then showed a dash where `$0.00` was the right answer.

Rather than work around that in the frontend, the metrics response now carries the two signals it was missing:

- `enriched_traces` — how many traces enrichment has been through. Polling stops on this instead of on cost, so it stops when there is demonstrably nothing left to learn rather than when a number happens to appear. The two-minute window from `completed_at` stays, demoted to a backstop for enrichment that dies partway and leaves traces unprocessed forever.
- `priced_traces` — how many carry a cost figure at all. This is what separates "cost nothing" from "nothing could be priced". The card now shows `$0.00` for a run whose traces were priced and add up to zero, and keeps the dash only where nothing was priced.

`priced_traces` needed a cost expression that keeps its NULL rather than coalescing to zero. `enriched_cost_expr` already takes `default=None` for the traces list, where an unpriced trace has to sink under `nullslast` instead of claiming to have spent nothing, so `per_trace_usage_subquery` now carries that variant alongside the coalesced one and `COUNT` skips the NULLs. Both counts collapse per trace, not per span row.

Seven new backend tests cover it, including the case that made the old logic wrong: an enriched trace with an empty breakdown counts as enriched but not priced, while a trace priced at `0.0` counts as both. Reverting the frontend to the old cost-based guess turns four tests red.

**On the query key**: `projectId ?? ''` is the pattern in five other places here (`PermissionsContext` keys on the same value the same way), and `enabled` makes the collision unreachable, so on its own I would have left it. But the same `?? ''` also appeared in the request body, which is the worse half — the key and the request each fell back separately. Both are gone: the key uses a `'no-project'` sentinel, and `projectId` is narrowed inside `queryFn`, so the cache key and the request now name the same project by construction.

Issue #2198, opened 2026-07-19
Introduce first-class annotation entity and migrate reviews off JSONB


harry-rhesis:
## Summary

Introduce a first-class `annotation` entity that can attach to any entity (tests, test results, traces, and beyond), replacing today's model where reviews live as JSONB blobs inside `test_result.test_reviews` and `trace.trace_reviews`. Migrate the existing review code and data onto it, then retrofit the Annotations hub built in #2194 to read from the new entity instead of flattening JSONB.

## Background / Context

#2194 ("Add Annotations review hub") ships a read-only hub over the current JSONB review model. It is agreed to land as-is against that model (see the review and discussion on that PR), with this issue tracking the follow-up to the foundation we actually want.

The JSONB model forces a workaround at every layer, and the hub PR surfaces all of them:

- `list_annotations` hand-writes `organization_id` / `project_id` predicates because `db.execute(text(...))` bypasses the ORM auto-filter.
- `populate_review_permitted_actions` exists only because JSONB sub-documents never flow through a `WithPermittedActions` schema.
- The hub route needs an `AUTHZ_EXEMPT_ROUTES` exemption plus an in-handler dual-gate (with a bare `"telemetry:read"` string, no `Permission.Telemetry` enum) because it cannot use the standard router-derived RBAC backstop.
- Search is `ILIKE` over `jsonb_array_elements` LATERAL scans with `UNION ALL` and `COUNT(*) OVER()`, no index path.
- Mutations go through `flag_modified` plus dict-key poking instead of a normal UPDATE.

A first-class entity inherits tenant filtering, affordances, and RBAC automatically, and collapses the hub into a standard CRUD list.

## Goals

- Reviews and annotations are rows in one `annotation` table, not JSONB embedded in parents.
- Annotations can attach to any entity via our established polymorphic pattern.
- The two behaviours the current reviews carry are preserved (override semantics and sub-entity targets, see below).
- The #2194 hub, review drawers, and all review-derived UI read from the new entity, and the JSONB workarounds are removed.
- No data loss: existing reviews are migrated.

## Deliverables

- New `annotation` model, schema, CRUD, and router.
- Data migration from `test_reviews` / `trace_reviews` JSONB into `annotation` rows.
- Backend read/write paths repointed off the JSONB shape.
- Frontend review consumers repointed to the new fetched-and-cached resource.
- #2194 hub retrofitted (raw SQL, dual-gate exemption, and affordance projection removed).
- Docs updated.

[...]

PR #2757, opened 2026-09-17
Move metric tuning and explorer labels onto the annotation entity

harry-rhesis:
Fourth in the stack: #2724 (backend entity) → #2734 (frontend cutover) → #2737 (SDK/MCP/skill) → this. All three are merged, so this targets `main` directly. Part of #2198.

Two JSONB stores of human judgement move onto the annotation table. Both were the kind of storage that forced a workaround at every layer, and in both cases the table removes a rule that only existed to work around the column.

## Metric tuning

A judgement of what a metric said about one of its own tuning cases was a capped array in `test.test_metadata["reviews"]`. It is now an annotation on the case, targeting the metric by id, with the decision carried as an `Accepted`/`Rejected` status so the hub shows it in the same column it shows `Pass`/`Fail`. The verdict it was made against and the score type that verdict was read under go in `attributes` — only tuning re-reads a stored verdict, to ask whether the metric has moved since.

**Two rules disappear with the array.** Rows are append-only and the judgement in force is the newest one that still stands, so re-judging a case no longer overwrites: a mis-click costs nothing instead of one of ten slots. And the evict-accepts-never-comments rule existed solely to stop the ten-slot cap destroying the comments the feature exists to produce. A table has no slots to spend, so it has nothing left to protect against.

Routes rename with the concept: `.../cases/{id}/annotate` and `.../tuning/annotations/accept-rest`.

Judgements are written through `crud.annotation` rather than `services.annotation`, because that service exists for the status override an annotation implies on its parent and a `Test` parent has none — which is also what lets accept-rest leave the commit to the request session instead of committing once per case. Listing forty cases, folding the agreement ratio and reading the rejections back for Improve are each one annotation query.

## Explorer labels

A pass/fail label on an explorer test had two possible authors sharing one metadata key, so whichever wrote last destroyed the other. A metric's verdict still lives in `test_metadata`; a person's is now an annotation. Storing them apart is the point: *the metric said fail and a human said pass* is the disagreement the explorer exists to surface, and one key can only hold whichever was written last.

Reads derive an effective label, a person's beating a metric's, so the tree, the import/export copy and everything downstream agree on what a test is labelled without knowing where the label came from.

Labels are **per person and edited in place** — labelling twice moves your row, clearing removes yours and leaves everyone else's alone. That is deliberately unlike the rest of annotations, where a second judgement is a second row: a tree cell shows one label, so a person editing theirs has to be editing, not appending.

Two things this would otherwise have broken quietly:

- Evaluating a metric with `overwrite=False` skips already-labelled tests. A human label is no longer in the metadata, so without reading annotations it would have re-scored exactly the tests someone had already judged.
- A copied test carries the effective label as plain metadata under `imported`/`exported`, never as an annotation, so a copy never claims someone judged a test they have never seen.

## Migration

Seeds the two statuses for existing organizations, then moves both stores. The tuning array is left in place for PR-5 to clear; the explorer pair is stripped, but only once its annotation has landed. Neither source stored an id, so idempotency is by natural key.

Three guards, each of which is a data-loss bug without it:

- **The strip requires an annotation carrying the same verdict**, not merely one existing. During a rolling deploy the new code can already have written that person a different verdict, which makes the insert skip the row as a duplicate; stripping it then would delete the old verdict and silently flip the test.
- **The downgrade's restore only fills an absent metadata label.** It writes with `||`, which overwrites keys, so without the guard a test the upgrade deliberately left alone would come back with the metric's verdict destroyed and relabelled as a person's.
- **The downgrade deletes every tuning judgement in the org**, not only those whose target still matches the case's metric. One left behind would strand a status the cleanup then deletes, failing on the foreign key with the downgrade half applied.

## Three pre-existing bugs found on the way

These were latent before this PR and are fixed here because this PR's features are what would have hit them:

- **`crud/test.delete_test` bypassed the soft-delete cascade registry**, soft-deleting by hand. A deleted test kept its files, and (since #2724) its annotations — rows the hub would list pointing at a parent nobody can open. `bulk_delete_tests` already cascaded, so only the single-delete path was affected.
- **`get_annotations_for_entities` ordered only by `updated_at`.** Callers read the first row as *the* latest, and rows written in one transaction share a timestamp, so two reads of the same data could disagree. The id now breaks the tie, as `models/mixins` already did.
- **`AnnotationStatus` described two different runtime shapes.** A full annotation names its status id `id`; one embedded in a parent payload names it `status_id`. Reading `.status.id` off a summary entry type-checked and returned `undefined`. Split into its own type — `tsc` stayed clean, confirming nothing depended on it.

## From the review pass

A correctness read found `case_outcome` telling a reviewer their judgement had been *invalidated by a material change* whenever the case carried any metric-target annotation — including one whose status was neither Accepted nor Rejected, which is what the SDK and MCP tools produce if they annotate a tuning case directly. So a case nobody had judged could claim a judgement was taken away. The decision filter moved into `judgements_for_metric`.

A conciseness read cut ~50 lines, the useful part being that the migration's skip-count was *broken* rather than merely verbose: it reported zero skips in precisely the case it existed to catch. Deleted, along with the docstring that promised it. Also dropped an `automated_label` field that had been added only to make a docstring claim true, and collapsed two duplicate batch loaders into one.

## Deliberately not here

- **Docs.** `docs/content/docs/annotations/index.mdx` still says "reviews" and links to `/docs/tasks/test-reviews`; the merge and redirect are PR-5, as planned.
- **The explorer annotations panel and indicator column.** Additive UI; the existing label control works unchanged against the new backend. A node now carries `annotations_count` and `last_annotation` so it can be built without another backend change.

## Test plan

- [x] Backend: 9084 passed, 0 failed (full suite). 716 across everything this PR touches, including the security drift suites.
- [x] New coverage: the rolling-deploy strip cases and the downgrade clobber case were each verified to **fail** against the pre-fix migration, not just pass after it.
- [x] Frontend: 2622 passed, `tsc` and `lint` clean.
- [x] Migration applies, reverses and re-applies against a real database, seeding exactly two statuses without duplicating and leaving existing annotations and metadata untouched. **Caveat:** the dev role is a superuser, so that round trip did not exercise row-level security — the per-org GUC path is covered by the test suite only.
- [ ] Manual walkthrough before merge, per the plan. The two checks a green suite cannot reach: judging a tuning case twice (it should append and show the newest, not overwrite), and evaluating a metric over an explorer topic with overwrite off (a hand-labelled test should be skipped; a metric-labelled test should keep its own label).


harry-rhesis:
Thanks, both checked against the code.

**On `updated_at`:** agreed with your own follow-up, no change needed. `Base.updated_at` is `Column(..., server_default=func.now(), onupdate=func.now())` (`models/base.py:27`), so the ORM bumps it on any UPDATE. Your parenthetical about ordering the `mine` lookup by `updated_at` isn't needed either: that query filters `deleted_at IS NULL` and clearing a label soft-deletes, so there is at most one live row per `(test, user)` and the ordering never decides anything.

**On the label author: you're right, and it is unrecoverable.** `test.user_id` is the creator, not the labeller. `update_test_node` on main writes `meta.label` without ever touching `user_id`, and its lookup (`get_test_in_test_set`) scopes to the organization rather than the owner, so any member could label a colleague's test and the row still named the creator.

I went looking for somewhere else that kept the actor, and there is nothing. The activity feed derives who-did-what from each row's own `user_id` (`RecentActivitiesService.determine_operation` works off timestamps), and `activity_log` is a job narrative table with no user column at all. `labeler = 'user'` recorded *that* a person set the label, never which one.

So best-effort is the only option, and I've kept the creator rather than `NULL`. In the explorer, adding a test and labelling it are normally the same person in the same session, so the creator is right in the common case, and `NULL` would give that up to avoid the rarer wrong one. What a wrong guess costs is bounded: a wrong name against the label, and the real labeller's next edit opens a second row instead of moving this one. Their new label still wins on read, since the read path orders by `updated_at DESC, id DESC`. A test with no creator is already skipped entirely and keeps its metadata label.

Documented in c42f76a73, as you suggested: a paragraph in the migration docstring on why the author is a guess and what it costs, plus a line at the `t.user_id` column so it's visible at the point of the choice. Behaviour unchanged, 67 alembic tests still pass including the explorer label cases that run that SQL.

PR #2762, opened 2026-09-17
Fix 22 Dependabot security advisories across Python lockfiles

harry-rhesis:
## Summary

- Bump soupsieve >=2.8.4 to >=2.9.0 across 8 projects (fixes 16 ReDoS alerts)
- Bump pydantic-ai >=1.56.0 to >=1.106.0 in examples/telemetry (fixes 6 SSRF + UploadedFile alerts)
- Remove dead nltk constraint pin from backend and SDK (transitive dep of garak, no upstream patch)
- Add crewai/pydantic-ai conflict pair in examples/telemetry so the two extras resolve independently against google-genai

3 nltk alerts remain with no fix available upstream.

## Test plan

- [ ] CI passes (lockfile regeneration, no source changes)
- [ ] Verify Dependabot alerts drop from 25 to 3 after merge


harry-rhesis:
Good question. Those markers are uv's internal lockfile representation for conflict forks. uv resolves each fork independently and writes the combined `extra == '...' and extra == '...'` markers to tag which fork a locked package belongs to. They aren't evaluated as PEP 508 at install time — uv's own installer interprets them directly. So it's expected output from the conflict declaration, not dead markers.

PR #2766, opened 2026-09-17
Lead the usage card with what the run cost

Arman-Beykmohammadi:
## Purpose

The Usage card on a test run's summary led with tokens, put cost beside them as a second figure, and captioned the whole thing "across 12 traces". Tokens are the workings. Cost is the answer people open the card for, and the trace count was never the question anyone was asking.

## What Changed

Cost is the headline now, with the tokens and the model that produced them underneath:

```
Usage
$0.03
112,557 tokens · gemini/gemini-2.5-flash
```

The model name wraps rather than clipping. The card is narrow and model names are long, so the grid's one-line-with-ellipsis treatment turned it into `gemini/ge...`, which answers nothing. `ModelLabel` takes a `truncate` prop for this; grid cells keep clipping, since a column has a fixed width to respect.

Where nothing has been priced there is no cost to lead with, so tokens keep the headline. Two different silences are worth telling apart there, and the run already knows which one it is:

| State | Card says |
| --- | --- |
| Enrichment still working through the run's traces | "Working out what this cost" |
| Enrichment finished, nothing it could price | "No priced models", linking to the docs |
| Priced, and the total is zero | `$0.00` — the run really was free |

The card also declares the six fields `GET /telemetry/metrics` has been sending since the input/output split landed: the input and output halves of both tokens and cost, plus `models_used` and `providers_used`. They were arriving on the wire and being dropped by the TypeScript interface. The trace tiles and the run comparison both need them next.

## Additional Context

The three-way branch moved into its own `UsageCard` component rather than growing a conditional inside `KpiRow`'s JSX. The Traces page needs the same rule, and one copy is what stops the two screens disagreeing about the same run.

`ModelLabel` needed a `component` prop as well. It renders a `Typography`, which maps `body2` to a `<p>`, and the card's subtitle is also a `Typography` — a `<p>` inside a `<p>` is invalid markup and React warns about it. It now takes `'p' | 'span'`, defaulting to `p`, so the existing grid call sites are untouched.

Nothing new was invented for the unpriced case: `isCostKnown` and `enriched_traces` both already exist from the polling work, and this reads them.

**Deliberately out of scope.** No backend change. The trace summary tiles still use their own component and their own layout; recutting those to match is the next change. Currency conversion is later still — every figure here is USD.

## Testing

`npm run format`, `npx tsc --noEmit`, `npm run lint` and jest all clean. 64 tests pass across the card, `ModelLabel`, the polling hook and both grids' usage columns.

Six new tests cover the reshaped card: cost as the headline, the model in the subtitle, the `+N` for several models, no stray separator when a run reported no model at all, the two unpriced messages, and a long name rendering in full.

Each new branch was checked by breaking it. Keying the fallback off the cost value instead of `priced_traces` turns the free-run test red. Dropping the empty-model guard turns the separator test red. Forcing truncation back on turns the long-name test red. Fixing `stillPricing` to false turns the still-working test red.


Arman-Beykmohammadi:
Both addressed.

**The `total_traces === 0` question was a real gap, and worse than it looked.** A run against an endpoint with no instrumentation produces no traces at all, which is not exotic — it is what you get whenever tracing is not wired up. The card read that as "enrichment finished and could price nothing", so it offered `0 tokens` above an explanation of *why models go unpriced*, which answers a question nobody asked.

Of your two suggestions I took hiding it. That run has no usage rather than usage of zero, and the card is already held back entirely before the first response arrives, so holding it back here is the same rule rather than a new one. It is a `hasTracedUsage` predicate exported beside the card, because the trace tiles need the same judgement next and one copy is what stops the two screens disagreeing.

**On the tooltip, the stated reason does not hold, but the change is still worth making.** MUI 7.3.5 has an explicit guard for this — `// There is no point in displaying an empty tooltip.` followed by `if (!title && title !== 0)` — so an empty title never opens and never warns. The warning you are thinking of is the one about disabled children, which needs a wrapper element.

What is true is that the Tooltip still wraps the child and attaches hover and focus listeners in order to decide it has nothing to show. So the single-model case now returns the label directly. Nothing changes on screen.

That distinction cost me a test. I first wrote one asserting no tooltip appears on hover for a single model — and it passed with the change reverted, because MUI's guard makes both versions behave identically. A test that cannot fail is worse than no test, so it is gone, and the one that stayed is the multi-model hover: reverting the restructure turns it red, which is the actual regression risk in splitting the return into two paths.

The no-traces fix is verified the same way — forcing `hasTracedUsage` to true turns its test red.

PR #2767, opened 2026-09-17
Recut the trace summary tiles around what they can answer

Arman-Beykmohammadi:
## Purpose

Four tiles sit above the traces table. They counted traces, spans, tokens and cost, each with an icon and no footnote, while the response they were built from carried the span breakdown, the error rate, the token split, the models and the providers. All of it fetched on every page load, none of it shown.

This spends those four slots on more.

## What Changed

| Tile | Value | Footnote |
| --- | --- | --- |
| Traces | 226 | 5 span types *(hover lists them)* |
| Spans | 3,089 | 40 errors · 99% ok |
| Usage | $0.19 | 571,062 tokens · 412,000 input · 159,062 output |
| Models | 2 models | openai/gpt-4o +1 *(hover lists them)* |

They are `KpiCard` now, the same component the test run summary uses, so one run is not described in two different shapes depending on which page you opened it from.

Usage follows the same rule as the summary card, for the same reason: cost leads where it is a figure somebody computed, tokens lead where it is not, and the tile says whether that silence is enrichment still working or nothing it could price. A scope priced at zero still shows `$0.00`, because then it really was free.

The tiles are held back entirely for a project that has traced nothing. The table below already says so in its own empty state, and four zeros say it worse.

First commit is a move rather than a change: `isCostKnown` lived in a test-runs hook and `hasTracedUsage` inside the summary card, so this page could only reach them across a feature boundary. Both are questions about a telemetry response rather than about a test run, and they now sit in `trace-utils` beside `formatCost`, which both pages already import from. `isPricingInProgress` joins them, named rather than left as the inline comparison it was — the second page needing it would otherwise have written the comparison out again and been free to get it subtly wrong.

## Additional Context

Two things differ from what I had planned, both deliberately.

The Usage tile does not say "cost" under the figure. That label was removed from the summary card during review of the previous change, and keeping it here would have made the two screens disagree about the same run.

The Models tile names the models rather than counting the providers. A provider count says less than the model that actually produced the spend, and `ModelLabel` carries the provider as its prefix anyway, so nothing is lost by dropping the separate count.

The error figure is deliberately worded against spans, not traces: `error_rate` is computed over span rows on the backend, so "40 errors" on the Spans tile is the honest placement for it.

**Deliberately out of scope.** `SummaryCard` is untouched — metric tuning still uses it, and this changes the call site rather than the component. No backend change; every field read here was already being sent. Currency conversion is a later change, so every figure is USD.

## Testing

`npm run format`, `npx tsc --noEmit`, `npm run lint` and jest all clean — 134 tests across the tiles, the shared predicates, the summary card, `ModelLabel` and the polling hook.

The component had no tests at all. It has 14 now: each tile's value and footnote, the two hover lists, both unpriced messages, a priced zero, the held-back empty scope, the unsupported-filters caption, and the run-scoped variant used by a test run's Traces tab.

Every new branch was checked by breaking it — the empty-scope guard, the priced check, the still-pricing check, the span-types footnote and the error footnote each turn exactly the expected test red.

One of those tests needed fixing before it was worth anything. The empty-scope test passed with the guard removed: `waitFor(getMetrics called)` resolves when the fetch *starts*, so nothing had re-rendered yet and the container was empty either way. It now flushes the state update inside `act` first, and it does fail when the guard goes.


Arman-Beykmohammadi:
Both addressed, though the second one turned out differently than described.

**The type-only import is now `import type`.** It is also the documented convention in the frontend rules; that line predates the rule and I widened it in this PR without bringing it in line.

**On the error count, the failure mode you described cannot happen — but a different one can, and does.**

I brute-forced every span count up to 60,000 against the backend's `round(error_count / total_spans, 4)`:

- Cases where `error_rate > 0` but the reconstruction yields `0`, so the tile wrongly reads "No errors": **0**. That one is unreachable. Both figures derive from the same `total_spans`, which pins them together — for the product to fall below 0.5 the span count has to be small, and a small span count forces the rate back up. So the tile could never have claimed a clean scope for one that had errors.
- Cases where the reconstructed count is simply **wrong**: **55,836**. The error is always upward. 3 errors in 11,667 spans rounds to `0.0003`, which reads back as 4.

So keying the branch off `error_rate === 0` would have fixed a bug that did not exist while leaving the real one in place: a tile confidently reporting four failures where there were three.

Fixed at the source instead. The aggregate already counts the failed spans and then keeps only the ratio, so `error_spans` is now sent alongside `error_rate` and the tile reads it. No arithmetic, no rounding, and the count is exact at any scale.

Two new backend tests cover the count and the zero case; on the frontend, a test with 11,667 spans and 3 errors asserts "3 errors" and the absence of "4 errors" — it fails if the reconstruction comes back. Also picked up a missing plural rule while there: one failed span now reads "1 error", not "1 errors".

Worth flagging separately: the branch had a merge commit from the "Update branch" button. The repo rule is rebase rather than merge, so I have rebased it onto current main and force-pushed; history is linear again and the PR diff shows only this work.

Arman-Beykmohammadi:
Superseded by #2769. This PR's head was left pointing at a merge commit that a later rebase removed from the branch; GitHub would not re-sync it and then refused to reopen it. The branch and the work are unchanged.

PR #2769, opened 2026-09-18
Recut the trace summary tiles around what they can answer

Arman-Beykmohammadi:
## Purpose

Four tiles sit above the traces table. They counted traces, spans, tokens and cost, each with an icon and no footnote, while the response they were built from carried the span breakdown, the error rate, the token split, the models and the providers. All of it fetched on every page load, none of it shown.

This spends those four slots on more.

## What Changed

| Tile | Value | Footnote |
| --- | --- | --- |
| Traces | 226 | 5 span types *(hover lists them)* |
| Spans | 3,089 | 40 errors · 99% ok |
| Usage | $0.19 | 571,062 tokens · 412,000 input · 159,062 output |
| Models | 2 models | openai/gpt-4o +1 *(hover lists them)* |

They are `KpiCard` now, the same component the test run summary uses, so one run is not described in two different shapes depending on which page you opened it from.

Usage follows the same rule as the summary card, for the same reason: cost leads where it is a figure somebody computed, tokens lead where it is not, and the tile says whether that silence is enrichment still working or nothing it could price. A scope priced at zero still shows `$0.00`, because then it really was free.

The tiles are held back entirely for a project that has traced nothing. The table below already says so in its own empty state, and four zeros say it worse.

First commit is a move rather than a change: `isCostKnown` lived in a test-runs hook and `hasTracedUsage` inside the summary card, so this page could only reach them across a feature boundary. Both are questions about a telemetry response rather than about a test run, and they now sit in `trace-utils` beside `formatCost`, which both pages already import from. `isPricingInProgress` joins them, named rather than left as the inline comparison it was — the second page needing it would otherwise have written the comparison out again and been free to get it subtly wrong.

## Additional Context

Two things differ from what I had planned, both deliberately.

The Usage tile does not say "cost" under the figure. That label was removed from the summary card during review of the previous change, and keeping it here would have made the two screens disagree about the same run.

The Models tile names the models rather than counting the providers. A provider count says less than the model that actually produced the spend, and `ModelLabel` carries the provider as its prefix anyway, so nothing is lost by dropping the separate count.

The error figure is deliberately worded against spans, not traces: `error_rate` is computed over span rows on the backend, so "40 errors" on the Spans tile is the honest placement for it.

**Deliberately out of scope.** `SummaryCard` is untouched — metric tuning still uses it, and this changes the call site rather than the component. No backend change; every field read here was already being sent. Currency conversion is a later change, so every figure is USD.

## Testing

`npm run format`, `npx tsc --noEmit`, `npm run lint` and jest all clean — 134 tests across the tiles, the shared predicates, the summary card, `ModelLabel` and the polling hook.

The component had no tests at all. It has 14 now: each tile's value and footnote, the two hover lists, both unpriced messages, a priced zero, the held-back empty scope, the unsupported-filters caption, and the run-scoped variant used by a test run's Traces tab.

Every new branch was checked by breaking it — the empty-scope guard, the priced check, the still-pricing check, the span-types footnote and the error footnote each turn exactly the expected test red.

One of those tests needed fixing before it was worth anything. The empty-scope test passed with the guard removed: `waitFor(getMetrics called)` resolves when the fetch *starts*, so nothing had re-rendered yet and the container was empty either way. It now flushes the state update inside `act` first, and it does fail when the guard goes.

---

## Since the first review

Two follow-ups from review, plus a fix you reported on screen.

**The separator rendered as literal `·`.** It was written as an escape inside a JSX text child, and JSX does not process backslash escapes there — the same escape three lines above, inside a template literal, was fine, which is what made it easy to miss. The usage subtitle is one composed string now rather than three flex children: both halves are plain text, so the separator belongs in the sentence where it wraps with the words either side and survives being copied. As an element its spacing lived in CSS and the text read `tokens·412,000 input`.

**The failed-span count is now sent, not reconstructed.** The tile derived it as `error_rate × total_spans`, and `error_rate` is rounded to four places before it leaves the API. Brute-forcing every span count up to 60,000 finds **55,836** combinations where that reconstruction lands on the wrong integer — always too high. 3 errors in 11,667 spans rounds to `0.0003`, which reads back as 4. The aggregate already counted the failed spans and kept only the ratio, so `error_spans` now travels alongside `error_rate`.

Worth recording what is *not* wrong, since it looks like it should be: a rate above zero can never rebuild a count of zero, so the tile could not have claimed "no errors" for a scope that had some. Both figures derive from the same span count, which pins them together. Also picked up a missing plural rule while there — one failed span now reads "1 error".

**Type-only import** in `trace-utils`, which is the documented frontend convention.

## Why this replaces #2767

That PR's head was left pointing at a merge commit which a later rebase removed from the branch. GitHub would not re-sync it and then refused to reopen it, so the work is unchanged and the pull request is new. Branch, commits and review history of the code are the same; only the PR number changed.


Arman-Beykmohammadi:
Both fixed, and the first one was more common than it looked.

**"N errors · 100% ok" was reachable at 199 spans.** You attributed it to `error_rate` being rounded, but the bigger culprit was `Math.round` on the share itself: anything at or above 99.5% rounds to 100 regardless of how the rate was computed. Brute-forcing every span count up to 60,000 with one to five errors gives **297,030** combinations where the tile contradicted itself, the smallest being 199 spans with a single error.

It now comes from the counts rather than the rate, capped at 99 while any span failed. I took the cap rather than the floor you suggested first, because flooring costs accuracy everywhere to fix the edge: a genuine 98.7% would read as 98 instead of 99. Capping only touches the case that was wrong.

**Span types are sorted busiest first.** Descending count rather than alphabetical, since the question a reader has is which kind of work dominates, with the name as a tie-break so the order is total. You are right that the order was not merely unspecified but genuinely unstable: the endpoint builds that dict from a grouped query with no `ORDER BY`, so two identical loads could produce different orders.

Both verified by reverting them: restoring the rounded share turns the contradiction test red, and removing the sort turns the hover-order test red.

PR #2770, opened 2026-09-18
Compare what two runs spent, not just their pass rates

Arman-Beykmohammadi:
## Purpose

Comparing two test runs showed pass rate and nothing else. The obvious follow-up — did the better run also cost more, and was it even the same model — had to be answered by opening both runs in turn and holding the numbers in your head.

## What Changed

Both cards now read:

```
Pass rate: 85% (+2.2%)
Cost: $0.05 (+$0.02)
Tokens: 180,000
Model: gemini/gemini-2.5-flash
```

The cost delta sits beside the pass-rate delta and is coloured the opposite way round: a run that got cheaper is a run that improved.

No delta is shown when either side is unpriced, and an unpriced run shows an em-dash rather than `$0.00`. A baseline of "no idea" would otherwise manufacture a saving that never happened.

Both sides get their numbers from the same endpoint through `useTestRunUsage`, so the two cards cannot describe one figure differently, and neither can disagree with the Usage card on the run's own summary page.

## Additional Context

I had planned to take the baseline's usage from the list response, which already attaches it per run. Two things made that the wrong route. The list is the run *picker*, carrying every selectable run, so threading usage through the props ships figures for runs nobody is looking at. And its `TestRunUsage` shape has no priced-trace count, so the baseline could not have told an unpriced run from a genuinely free one while the current run could — two different rules for the same question, on two cards sitting side by side. Reading the metrics endpoint for both is less code and one rule.

**The pass-rate delta was wrong, and is fixed here.** `currentPassRate` was rounded to an integer while `baselinePassRate` was not, and the delta subtracted one from the other. A run going from 82.4% to 84.6% reported **+2.6** where the truth is **+2.2**, and the error grows with the rounding on either side. The rate is kept exact now and rounded only where it is printed, so the displayed integers are unchanged and the delta is right.

`ModelLabel` gained an `sx` passthrough. It renders a `Typography`, so the `body2` variant gave it its own weight and colour; inline in a `Cost: $0.05` style line that made the model the one value on the card that looked unstyled. A passthrough rather than a second component, so one place still decides how `provider/model` and its `+N` are written.

**Deliberately out of scope.** No backend change and no new endpoint. `compare/page.tsx` is untouched. The comparison table below the cards is unchanged. Currency conversion comes later, so every figure here is USD.

## Testing

`npm run format`, `npx tsc --noEmit`, `npm run lint` and jest all clean — 146 tests across the new component, `ModelLabel`, the shared usage predicates, the summary card, the trace tiles and the polling hook.

Eleven new tests cover the lines and the delta: the three figures, nothing rendered before the numbers arrive or for a run that traced nothing, the em-dash for an unpriced run against a real `$0.00` for a free one, the missing model line, and the delta's sign, colour, suppression when equal, and suppression when either side is unpriced.

Each rule was checked by breaking it — swapping `isCostKnown` for a truthiness check on the cost, dropping the unpriced guard on the comparison, removing the traced-nothing guard, and swapping the two delta colours each turn exactly one test red.

One of those breaks initially appeared not to fail, which was my own error rather than a gap: Prettier had joined the line I was patching, so the edit silently did not apply. Re-applied against the real text, it fails as it should.


Arman-Beykmohammadi:
Fixed, and you were right about the mechanism this time — I checked it rather than assuming.

`0.1 * 3 - 0.3` is `5.551115123125783e-17`, and `formatCost` prints anything below a thousandth to six places, so that renders as `(+$0.000000)`: a delta claiming a difference too small to write down.

The delta is now rounded to six places before it is compared or printed. Six because that is the precision the backend already rounds costs to and the most `formatCost` will ever show, so nothing displayable is lost in the guard.

Two tests pin both ends of that choice, which matters more than the guard itself. One fails if the rounding goes away — `0.1 * 3` against `0.3` must show no delta at all. The other fails if the rounding goes further: a real difference of `$0.000002` must still appear, so rounding to two places would swallow it. I checked both by reverting to no rounding and by over-rounding to two places, and each turns exactly its own test red.

One thing worth recording, since it changes how load-bearing this is: it is **not reachable today**. Both costs arrive as single doubles straight from the API, so two equal costs are the same double and subtract to exactly zero — the noise needs an arithmetic chain, and there is none yet. It stops being unreachable the moment currency conversion lands and both sides get multiplied by a rate, which is the next change in this series. So the guard goes in now rather than being discovered later as a mystery `$0.000000`.

PR #2772, opened 2026-09-18
Show costs in USD, EUR, GBP or CHF

Arman-Beykmohammadi:
## Purpose

Costs were dollars everywhere. Enrichment has always also stored a EUR figure per trace, but nothing read it — and it could not safely be read, because it is frozen at whatever rate was cached when that trace was enriched, with no record of when. Reading it for EUR while converting live for anything else would put two currencies on one screen on two different bases.

So USD stays the record, and every other currency is a conversion of it at display time.

## What Changed

**Four currencies: USD, EUR, GBP and CHF.** Chosen in two places. An organization sets the default under Settings → Display, which everyone sees. An individual overrides it under their own Settings → Preferences, or chooses to follow the organization. Precedence is user, then organization, then USD.

**Rates are live, not constants.** The existing service asked Frankfurter for EUR alone, in four places. It now fetches every currency in one call and caches the snapshot for a day, and `GET /services/exchange-rates` exposes it. Frankfurter republishes European Central Bank reference rates, which the ECB sets once per working day — daily rather than real-time, which is the right granularity for figures measured in fractions of a cent.

**Thirteen call sites** switch to the chosen currency: both grids, both trace panels, the trace tiles and the test run summary card. Hovering a cost in either grid shows the same figure in the other currencies, and the day the rates are from.

## Additional Context

Several decisions worth naming, all of them about not asserting more than is known.

**A currency with no rate falls back to the stored one** rather than converting at a guess. The `USD_TO_EUR_RATE` env var only ever described EUR, so it still only supplies EUR; on an instance with no outbound network the other currencies simply are not offered, and costs show as real dollars rather than invented pounds.

**The rates carry their own date, and null when they have none.** A figure from the configured fallback has no business day behind it, so the hover says the age is unknown instead of implying the rate is current.

**Decimals come from the stored figure, not the converted one.** A hover listing one cost four ways would otherwise print it to four different precisions.

**A stored currency the product no longer offers is ignored**, rather than handed to `Intl.NumberFormat`, which would throw and take the page with it.

`usageColumns` is the one call site that could not read a context: it builds column definitions outside a component, so it takes the formatter as an argument and `TestRunsGrid` passes it down. Both grids list it among their memo dependencies so the columns rebuild when the currency changes.

Neither setting costs a request. The user's arrives server-seeded with the layout and the organization's rides along on the organization the context already holds; only the rates are fetched.

**Deliberately out of scope.** The EUR figure in the enrichment blob is left alone — it is still written, just not read, and removing it is a migration for another day. The run comparison cards are not converted here because they live in an open PR that has not merged yet; they follow in a one-line change once it does. No new currency needs a migration: both settings are JSONB and the list is a constant.

## Testing

Backend: `uvx ruff format --check`, `uvx ruff check` and 474 tests pass. Frontend: `npm run format`, `npx tsc --noEmit`, `npm run lint` and **2,782 tests** pass.

New coverage is 23 tests on the rate service, 5 on the endpoint, 9 on the two settings round-trips, 17 on the formatter and 13 on the context.

Everything load-bearing was checked by breaking it. Reversing the precedence, dropping the unknown-currency guard, taking decimals from the converted figure, removing the missing-rate fallback, dropping the base-currency identity, collapsing the multi-currency request back to EUR, and breaking each link of the stale-cache chain each turn exactly the expected test red.

One test needed rewriting rather than adding: `formatCost`'s existing suite now runs through `Intl`, and it passes untouched. The single visible change is that a cost in the thousands finally gets a thousands separator.


Arman-Beykmohammadi:
Both fixed.

**The unparseable `USD_TO_EUR_RATE` was a real hole, and worse than it reads.** The parse sat at the bottom of the fallback chain, so the one condition under which it runs is exactly the one the chain exists for — no rates, nothing cached. A misconfigured value would have raised out of the rates endpoint, whose entire purpose is to always return something. The docstring promised "never leave the caller with nothing" and held everywhere except there.

It is parsed in one place now instead of three, and guards two cases rather than one: an unparseable value logs and takes the built-in default, and so does a zero or negative one. A rate of `0` parses perfectly well and would make every cost in the product read as nothing.

Six new tests cover it, including empty string, `nonsense`, `0`, `-1` and `1,2` — that last one being the realistic misconfiguration, someone writing a European decimal comma. All six fail against the old bare `float()`.

**The failing Backend tests check was not from this branch.** `TestUpdateTestResultObjectAuth::test_creator_can_update` and its delete counterpart have been red on main since test results started serializing `created_at` and `updated_at`: the fixture builds a `spec_set` mock listing every attribute it allows, so response validation failed and the request 500'd before reaching the auth assertion the test exists to make. I confirmed it by running them on `main` with this branch not checked out.

Fixed here rather than left, since it was the only red check on this PR, but it is unrelated to currency and worth knowing it was already broken.

Arman-Beykmohammadi:
Fixed, and you were right to raise it twice — my previous change only covered half of it.

The fallback path did reset the clock, but the **stale-snapshot** path did not: it returned `self._snapshot` and left `_last_fetch` untouched, so the cache stayed stale and the next request tried the network again. During a provider outage that puts a fresh call, and up to its five-second timeout, in front of every single request. I confirmed it through the async path the endpoint actually uses: three requests made three attempts, and now make one.

Both failure paths reset the clock to the same short retry window the env fallback uses. The snapshot keeps its own `as_of`, so rates served past their day still report the day they are genuinely from rather than claiming to be current — that was the part worth being careful about in your suggestion, and there is a test pinning it.

While there I replaced the `is_fallback` boolean with the window itself. All three cases differ only in how long the answer is good for — fresh rates a day, stale rates and the env fallback five minutes — so carrying the window with the snapshot says that directly, rather than a flag that a second reader has to map onto a duration.

Three new tests: two calls during an outage make one attempt, a retry happens minutes later rather than a day later, and stale rates keep their real date. Reverting to `return self._snapshot` turns the first two red.

Worth noting this was not only a latency concern. The same missing reset is what produced the bug reported separately on this PR — GBP and CHF rendering with dollar signs — because an instance sitting on the EUR-only fallback had no rate for them and fell back to showing the stored figure under their label. That is fixed at both ends now: the backoff here, and the picker no longer offering a currency it has no rate for.

Arman-Beykmohammadi:
The red Backend tests check was two failures, and they split.

**One was mine.** `test_none_becomes_the_default_shape` asserts the exact default organization settings blob, and this branch adds a `display` section to it. I had searched `tests/backend/routes/` for settings tests and missed `tests/backend/models/`. Fixed, with two more alongside: an organization that has chosen no currency reads as unset, and a settings row written before the section existed still reads rather than raising.

**One was not.** `TestRedisBackedCachePoolSizing::test_cache_passes_its_pool_size_to_both_clients` fails identically on `main` with this branch not checked out. It is order-dependent — it passes when the whole file runs and fails when run on its own — so it is a test-isolation problem in the cache suite rather than anything to do with currency. Left alone here; it needs its own change.

PR #2775, opened 2026-09-19
Name the broken model setting, at boot and in the 500

### LINKED ISSUE #2671: Warn at boot when a default model cannot be built, and say so in the 500
## Summary

A deployment whose `DEFAULT_*_MODEL` cannot be constructed gives the operator no usable signal: nothing at boot, and a generic 500 on the first execute. Two changes to fix that.

Rewritten down from the original report. Both questions it raised are now settled (see Background); what is left is only the diagnostics.

## Background

Originally filed as: `POST /test_sets/{id}/execute/{endpoint_id}` returns a 500 on a backend with no `RHESIS_API_KEY`. The execute endpoint pre-flight-validates the evaluation and execution models by constructing them, both default to `rhesis/rhesis`, and `RhesisNativeModel.__init__` raises `ValueError("RHESIS_API_KEY is not set")` (`sdk/src/rhesis/sdk/models/providers/native.py:75`). That `ValueError` is the parent of `ModelConfigurationError`, which is the only type `validate_execution_model` catches, so it falls through to the generic handler.

Settled since:

- **Should execution work without the key? No.** #2681 decided the 500 is correct. The model that fails is the *deployment* default, which an API caller cannot fix, so a 400 saying "check your model settings" would misdirect. `tests/backend/routes/test_execution_validation.py:91` and `tests/backend/app/test_usage_tracking.py:518` defend that on purpose.
- **The `[Test] Java SDK` failure is fixed.** #2681 set `RHESIS_API_KEY` and `RHESIS_BASE_URL` on the `java` compose profile. The workflow has been green since.

Still true: the cause appears only in the backend logs. The response body is `{"detail": "An unexpected error occurred."}`, the generic 500 copy from `app/error_handlers.py:20`, and the backend starts clean whether or not its own default model can be built.

## Deliverables

1. **Warn at startup when a `DEFAULT_*_MODEL` cannot be constructed.** Listed as a follow-up on #2681 and never done. Try building each configured default at boot, log a warning naming the model and the reason. A warning, not a failed boot: a deployment that never runs a model should still start.

2. **Name the missing configuration in the 500 body.** Keep the status code, keep the plain-`ValueError`-propagates behaviour that #2681 chose. Only the response detail on this path changes, from the generic copy to one that says the deployment's default evaluation or execution model cannot be built.

## Acceptance criteria

- [ ] A backend started with an unconstructible `DEFAULT_GENERATION_MODEL`, `DEFAULT_EVALUATION_MODEL`, `DEFAULT_EXECUTION_MODEL` or `DEFAULT_EMBEDDING_MODEL` logs a warning at boot naming the model and the reason, and still starts.
- [ ] `POST /test_sets/{id}/execute/{endpoint_id}` on that backend returns a 500 whose body names the missing configuration rather than "An unexpected error occurred."
- [ ] `test_validate_execution_model_generic_error` and `test_raises_when_construction_fails` still pass, or are updated with the reason their intent changed.

harry-rhesis:
## Purpose

A deployment that cannot build its own `DEFAULT_*_MODEL` gives the operator nothing to work with. Nothing at boot, and on the first execute a 500 reading `{"detail": "An unexpected error occurred."}`. The cause appears only in the backend logs. That is what #2671 was actually reporting, once the two questions it raised were settled: #2681 decided the 500 is correct (the model that fails is the deployment default, which an API caller cannot fix) and fixed the `[Test] Java SDK` failure by configuring the key on the `java` compose profile. What was left is the diagnostics.

## What Changed

- **`warn_on_unbuildable_default_models()`** (`user_model_utils.py`) tries to build each of the four `DEFAULT_*_MODEL` settings and logs a warning naming the setting, the model string and the reason. A warning rather than a failed boot, because a deployment that never resolves a model still has to start. It catches every exception type, not just `ValueError`, since huggingface raises `ImportError` without torch and a boot diagnostic that takes the app down would be worse than the problem it reports. Once per process: deployment settings cannot change within one, and the test suite starts a fresh app lifespan per test.
- Called from the **API lifespan** and from **`worker_ready`**. A metric evaluation resolves a model on the worker, so where the API and worker environments differ that signal is the only place the worker's own gap shows up.
- **`_deployment_model_error()`** (`execution_validation.py`) replaces the generic 500 body on this path with one that names the setting. Status code unchanged. It goes through `internal_error`'s `public_detail`, so the exception text still stays out of the response and lands in the log with its traceback.
- `validate_execution_model` loops over its two purposes so the message names the one that failed rather than both.

Only a deployment default can reach the new branch. Every failure to build an organization's own configured model is raised as `ModelConfigurationError` by `_build_configured_model` and still answered with the existing 400. It catches `(ValueError, ImportError)` rather than bare `Exception`, so `QuotaExceededError` still reaches its own handler and becomes a 402.

## Additional Context

- Closes #2671. Follows #2681, which made the status-code decision this builds on and listed the startup warning as a follow-up.
- `test_validate_execution_model_generic_error` asserted that the plain `ValueError` propagates, which is exactly what this changes, so it is replaced by tests for the status, the type, the named setting and the exception text staying out of the body. `test_raises_when_construction_fails` covers `resolve_default_hosted_model` and is untouched.
- Not changed: `handle_execution_error`'s `ValueError` branch still answers 400 with the raw message. That is a different path, inside route bodies rather than the pre-flight dependency, and out of scope here.

## Testing

Verified for real, not only against mocks.

With `RHESIS_API_KEY` unset, the boot check logs:

```
WARNING DEFAULT_GENERATION_MODEL=rhesis/rhesis cannot be built on this deployment: RHESIS_API_KEY is not set. Anything that resolves to it will fail when it is used.
WARNING DEFAULT_EVALUATION_MODEL=rhesis/rhesis ...
WARNING DEFAULT_EXECUTION_MODEL=rhesis/rhesis ...
WARNING DEFAULT_EMBEDDING_MODEL=rhesis/rhesis-embedding ...
```

Silent with the key set.

The response, through the actual `http_exception_handler` rather than an assertion on the exception:

```
500 {'detail': "This deployment's default evaluation model could not be built. Check the backend's DEFAULT_EVALUATION_MODEL setting and the credentials it needs."}
```

New file `tests/backend/app/test_default_model_startup_check.py`, 6 tests: warns per setting, survives an `ImportError`, silent when everything builds, runs once per process, uses the right model type per setting (the embedding default is a different SDK type), and the worker runs the same check. Three new tests in `test_execution_validation.py` for the 500, the purpose it names, and `QuotaExceededError` not being swallowed.

```bash
cd apps/backend
uv run pytest ../../tests/backend/app ../../tests/backend/routes/test_execution_validation.py ../../tests/backend/security ../../tests/backend/services ../../tests/backend/utils ../../tests/backend/jobs
```

5226 passed, 15 skipped, 1 xfailed. All skips pre-existing. Ruff clean on every file touched.


harry-rhesis:
Closed by a branch rename (`larvacean` -> `fix/default-model-diagnostics`). Same commits, continued in #2777.

PR #2784, opened 2026-09-21
Anchor exporter-side conversation joins in the shared store

harry-rhesis:
## Purpose

Follow-up from Arman's review of #2713, where this was point 2 and the concern he cared most about after the one I acted on in that PR.

After #2713 and #2715 there were two mechanisms for putting a conversation's turns on one trace, which is fine and probably unavoidable: one has to own span creation, the other exists for frameworks whose root span we do not create. What was not fine is that they kept **two separate stores** keyed by the same conversation id:

- `_anchors` in `packages/rhesis/src/rhesis/telemetry/conversation.py`, used by `conversation_turn`, `@endpoint`/`@observe`, LangGraph and Haystack.
- `ConversationTraceRegistry` in `genai.py`, used by Google ADK and MAF.

So they disagreed. Open turn one with `conversation_turn("c-1")`, then run turn two as a bare `agent.run()` under `set_conversation_id("c-1")`: the shared store has `c-1` anchored to the first trace, the registry has never heard of it, anchors `c-1` to the second, and the conversation arrives as two traces under one id. That is the same bug #2714 fixes for Haystack.

`get_conversation_anchor` and `anchor_conversation` were exposed in #2711 to prevent precisely this. Their docstring says they exist so an integration can join the same anchor "rather than keeping a second store that would disagree with it for the same conversation id" — and the registry was that second store.

## What Changed

`ConversationTraceRegistry` now reads and writes the shared store instead of keeping its own anchors. It still owns the rewrite map, which is what the exporter needs and is specific to retracing.

The ordering matters and is deliberate: the anchor is **written before it is read back**, so the shared store's first-writer-wins is the single decision about who anchors a conversation. Two threads opening the same conversation at once both write; whichever the store keeps is the anchor, and the other reads it back and records a rewrite onto it like any later turn. A get-then-set would have let both believe they were the anchor.

The local `max_conversations` cap goes with the local store, since the shared store has its own. The rewrite map keeps its own bound.

A poisoned anchor is now also rejected here: an all-zero trace id is not a trace anything can be written to, so a claim against one records no rewrite rather than retracing spans onto it. #2714 stops Haystack writing such a value in the first place; this is the other half.

## Additional Context

- Independent of #2713, #2714 and #2715. It touches `genai.py`, which none of them modify, and applies to `main` on its own. It changes ADK's live behaviour immediately and MAF's once #2713 merges.
- `ConversationTraceRegistry.clear()` no longer clears anchors, because they are not this class's to clear. Tests that need a clean slate clear `rhesis.telemetry.conversation._anchors`, which is what the LangGraph tests already do.
- This does not merge the two *mechanisms*, only the two stores. Anchor-plus-synthetic-parent and claim-plus-retrace still exist separately, for the reason above. As Arman noted, the output shape was already consistent either way: both end with several `is_turn_root` spans on one trace id, and the exporter keys by trace id. It was the keying that diverged.

## Testing

`3169 passed` in `tests/sdk` (excluding `tests/sdk/integration`, which needs Docker). Ruff clean on the three files changed.

Six new tests in `test_genai.py` covering the interaction with the shared store, including both directions of the interleaving bug, and two in `test_google_adk.py` (one replacing a test of the cap that moved).

Mutation-tested, each caught by the test written for it:

| Mutation | Caught by |
|---|---|
| Never write the shared anchor | `test_the_first_claim_anchors_in_the_shared_store`, + 6 more |
| Read a private anchor instead of the shared one (the old behaviour) | `test_a_turn_joins_an_anchor_another_mechanism_wrote`, + 6 more |
| Accept an all-zero anchor | `test_a_poisoned_anchor_is_not_used` |
| Claim without a conversation id | `test_no_conversation_id_anchors_nothing` |
| Leave the rewrite map unbounded | `test_the_rewrite_map_is_bounded` |

The first two each take seven tests with them, including the existing end-to-end ADK join tests, which is the sign the store is actually load-bearing rather than shadowed by a local copy.

One thing worth a reviewer's eye: the existing `TestConversationTraceRegistry` tests passed before I added an isolation fixture, but only because they happened to reuse conversation ids consistently. With a process-wide store their result depended on collection order. The fixture makes that explicit rather than lucky.


harry-rhesis:
Valid, and fixed in `3d1d1a7`.

You are right that it was reachable and right about the consequence. `claim()` rejected `None` but not `0`, so a non-recording span from a disabled or shut-down provider would have had its all-zero id formatted and written as the anchor. First-writer-wins then makes that permanent for the process, with no healthy later turn able to correct it.

What makes this a fair catch rather than a marginal one is that it is the exact mirror of something this branch already guards. I added `test_a_poisoned_anchor_is_not_used` for the read side, then failed to apply the same reasoning to the write side in the same function. Rejecting a value when reading it back while happily writing it is not a coherent position.

`if not trace_id or ...` now covers the invalid id and `None` together, with a comment saying why the two are the same case. Two regression tests: the invalid claim writes no anchor, and a later valid turn can still anchor the conversation afterwards. The mutation that restores `trace_id is None` fails both.

Worth noting this is the same bug class as the one you found on #2714, where Haystack published a zero trace id into this store. That PR stops it being written from the Haystack side; this stops it being written from the exporter side and stops it being used. Both halves are now closed.

PR #2813, opened 2026-09-23
Fix run summary hiding tests from requirements with no metrics

harry-rhesis:
## Purpose

A run summary reported "1 failed" with no red cell anywhere in the grid. The failing test belonged to a requirement with no metrics. The endpoint answered fine, but with nothing to score it the test was stored as Error, which counts as a failure, and the grid had no row to show it in. Preflight did warn about the requirement, but said its tests "will be skipped", which isn't what happened. A second bug made the same grid misleading: rows for a metric name shared by two requirements showed up under both.

A test that no metric was meant to judge is a setup gap, not a failure of the system under test. So it now runs, keeps its response for review, and gets no pass/fail verdict (Inconclusive).

## What Changed

- **Backend:** a test is stored as Inconclusive when the dispatch-time `metric_plan` confirms no metric applied to it. Planned metrics that never came back, endpoint failures and unusable multi-turn contracts all stay Error.
- **Backend:** results keep a `no_metrics_applied` marker, so a reviewer's annotation can give the test a verdict, and deleting that annotation returns it to Inconclusive rather than Error.
- **Backend:** the requirement-metric coverage preflight check now checks scope per requirement (a multi-turn test under single-turn-only metrics is flagged too), and its wording says the tests run unscored.
- **Backend:** `get_test_outcomes_for_run` reports `inconclusive` instead of folding it into `pending`, and each verdict-matrix requirement carries its own `test_status`.
- **Frontend:** metric rows match on `requirement_id` as well as the metric key.
- **Frontend:** a requirement with no metrics gets a "No metrics configured" row built from its `test_status`, and the Pass rate strip counts those tests too.
- **Frontend:** the verdict-blocks subtitle is computed from the rows and columns that apply. It used the first row, which could be scoped out of every test ("blocks: 2×2" instead of "2×4 and 2×1").
- **Frontend:** the Failures card names errored tests ("10 of 10 verdicts · 1 error"), with a tooltip that says it counts errors.
- **Frontend:** the Tests tab's Failed filter matches Fail and Error only, the same set the Failures card counts. Inconclusive results get their own filter, shown only when the run has some.

## Additional Context

- Existing results are not backfilled. A test that already errored this way stays Error, but now shows as a red cell under its requirement.
- Not changed: the requirement header's Total column counts every column in the run, and the in-place preview path (`services/test_execution.py`) still returns Error for a test with no metrics.

## Testing

- Backend: `uv run pytest ../../tests/backend/test_outcomes.py ../../tests/backend/test_outcome_writers.py ../../tests/backend/services/test_preflight.py ../../tests/backend/services/test_preflight_requirement_coverage.py ../../tests/backend/services/test_verdict_matrix.py ../../tests/backend/routes/test_verdict_matrix_api.py ../../tests/backend/services/test_annotation_override_test_result.py ../../tests/backend/jobs/` (819 passed). The full suite was not run.
- Frontend: `npx jest summary test-result-status TestRunFilterBar` (300 passed), plus `tsc`, ESLint, Prettier and the hardcoded-styles check.
- Manually: in a test set, give one requirement no metrics and run it. That test should show as No verdict under a "No metrics configured" row and not count as a failure, and preflight should warn about the requirement first.


harry-rhesis:
Thanks, both good catches.

1) Fixed. The preflight detail text now says "won't get a pass/fail verdict and will appear unscored in the run summary" instead of "won't count toward the pass rate". The old wording contradicted the strip, which does include those tests as unscored cells. Also reworded the summary line to "have tests no metric can score", which reads better for the mixed-scope case (requirement has metrics, but not for every test scope).

2) The all-X case can happen: a requirement with single-turn-only metrics and a mix of single-turn + multi-turn tests leaves the multi-turn columns X across every metric row. The preflight catches this (it flags any requirement where at least one test scope is uncovered), so the user is warned before the run starts. But the grid doesn't surface those tests in a placeholder row the way it does for fully metric-less requirements. That's a reasonable improvement but separate from this PR. The metric plan doesn't prevent the all-X case; it just doesn't produce verdicts for those cells.

